Issue #33, opened 2019-03-24
Combine driver manifest files into one


leakingtapan:
**Is your feature request related to a problem? Please describe.**
So that use can refer to a single manifest file using URL to deploy the driver.

PR #35, opened 2019-03-29
Combine manifest files

leakingtapan:
Fixes: #33


leakingtapan:
coverall failure due to repo transfer

Issue #42, opened 2019-05-01
unable to add .spec.csi data to volumeclaimtemplates prevents use with StatefulSets


RevREB:
/kind bug

leakingtapan:
AFAIK `volumeclaimTemplates` is used for creating multiple claims for provisioning new storage backends. I assume you are looking for reuse the same EFS filesystem for multiple pods in stateful set?

I tested it and it works. See the following spec as an example:
```
apiVersion: v1
kind: PersistentVolume
metadata:
  name: efs-pv
spec:
  capacity:
    storage: 5Gi
  volumeMode: Filesystem
  accessModes:
    - ReadWriteMany
  persistentVolumeReclaimPolicy: Retain
  storageClassName: efs-sc
  csi:
    driver: efs.csi.aws.com
    volumeHandle: fs-4af69aab
---
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: efs-claim
spec:
  accessModes:
    - ReadWriteMany
  storageClassName: efs-sc
  resources:
    requests:
      storage: 5Gi
---
apiVersion: apps/v1
kind: StatefulSet
metadata:
  name: efs-app-sts
spec:
  selector:
    matchLabels:
      app: test-efs # has to match .spec.template.metadata.labels
  serviceName: efs-app
  replicas: 3 # by default is 1
  template:
    metadata:
      labels:
        app: test-efs # has to match .spec.selector.matchLabels
    spec:
      terminationGracePeriodSeconds: 10
      containers:
      - name: linux
        image: amazonlinux:2 
        command: ["/bin/sh"]
        args: ["-c", "while true; do echo $(date -u) >> /efs-data/out.txt; sleep 5; done"]
        volumeMounts:
        - name: efs-storage
          mountPath: /efs-data
      volumes:
      - name: efs-storage
        persistentVolumeClaim:
          claimName: efs-claim
```

PR #43, opened 2019-05-06
Add example stateful sets

leakingtapan:
add example for stateful sets


leakingtapan:
Ref #42

PR #48, opened 2019-07-17
Update to CSI v1.1.0

wongma7:
**Is this a bug fix or adding new feature?** new feature.

**What is this PR about? / Why do we need it?** fixes https://github.com/kubernetes-sigs/aws-efs-csi-driver/issues/27 . I didn't implement NodeGetVolumeStats.

Also, small unrelated change: I moved volumeCaps from driver.go to node.go where it's used

**What testing is done?**  `make test` passes

### DIFF
--- a/pkg/driver/node.go
+++ b/pkg/driver/node.go
@@ -21,14 +21,18 @@ import (
 	"fmt"
 	"os"
 
-	csi "github.com/container-storage-interface/spec/lib/go/csi/v0"
+	"github.com/container-storage-interface/spec/lib/go/csi"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 	"k8s.io/klog"
 )
 
 var (
-	nodeCaps = []csi.NodeServiceCapability_RPC_Type{}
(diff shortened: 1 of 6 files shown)


pkg/driver/node.go
```diff
@@ -134,17 +146,10 @@ func (d *Driver) NodeGetInfo(ctx context.Context, req *csi.NodeGetInfoRequest) (
...
-	return &csi.NodeGetIdResponse{
-		NodeId: d.nodeID,
-	}, nil
-}
-
 func (d *Driver) isValidVolumeCapabilities(volCaps []*csi.VolumeCapability) bool {
 	hasSupport := func(cap *csi.VolumeCapability) bool {
 		for _, c := range volumeCaps {
-			if c.GetMode() == cap.AccessMode.GetMode() {
+			if c == cap.AccessMode.GetMode() {
```
leakingtapan: could you rename the `volumeCaps` to `volumeCapAccessModes`? This took me sometime to realize that the type of `volumeCaps` is change after its moved

wongma7: yes, fixed

PR #49, opened 2019-07-18
Refactor node tests to be table-driven

wongma7:
**Is this a bug fix or adding new feature?** testing

**What is this PR about? / Why do we need it?** makes tests slightly easier to read/add since they're less verbose and you can compare test cases based on their expected inputs and outputs (somewhat subjective: https://github.com/golang/go/wiki/TableDrivenTests)

depends on https://github.com/kubernetes-sigs/aws-efs-csi-driver/pull/48 

**What testing is done?** `make test` still passes with all the same cases

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -2,7 +2,7 @@ module github.com/kubernetes-sigs/aws-efs-csi-driver
 
 require (
 	github.com/aws/aws-sdk-go v1.16.5
-	github.com/container-storage-interface/spec v0.3.0
+	github.com/container-storage-interface/spec v1.1.0
 	github.com/golang/mock v1.2.0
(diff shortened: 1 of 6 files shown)


leakingtapan:
I'm not against table driven test. But I really don't like jamming `testFunc`s of different test cases into a giant one and add a bunches of if-else branches to handle different situations differently. This is error prune and hard to reason/read.

wongma7:
Sure, that's why I say it is subjective. From my perspective, I don't like checking every testFunc to see how it differs, I would rather there be 1 testFunc and compare the ~5 flags.

wongma7:
/hold
discuss later, don't want to block other PRs

Issue #50, opened 2019-07-18
Update to latest node-driver-registrar


wongma7:
Use https://github.com/kubernetes-csi/node-driver-registrar as https://github.com/kubernetes-csi/driver-registrar has been deprecated

Issue #36, opened 2019-03-31
Use skip attach


leakingtapan:
**Is your feature request related to a problem? Please describe.**
https://kubernetes-csi.github.io/docs/skip-attach.html

PR #51, opened 2019-07-18
Clean up deploy manifest for 1.14

wongma7:
bug fix+feature
- apps/v1beta2 and apps/v1beta1 are deprecated and will be removed in 1.16 https://github.com/kubernetes/kubernetes/pull/70672
- fix https://github.com/kubernetes-sigs/aws-efs-csi-driver/issues/50, update to latest node-driver-registrar. . driver-registrar is deprecated, use node-driver-registrar instead.
- fix https://github.com/kubernetes-sigs/aws-efs-csi-driver/issues/36, create a CSIDriver (now in-tree and beta in 1.14) to skip attach
- registration-dir is /var/lib/kubelet/plugins_registry in 1.13+
- remove controller statefulset (can add it back later for CreateVolume)
- remove all unnecessary RBAC rules as the new node-driver-registrar doesn't need any nor does our node driver (can add them back later for CreateVolume)
- depends on https://github.com/kubernetes-sigs/aws-efs-csi-driver/pull/48

**What testing is done?** 
The static_provisioning example works correctly on a 1.14/1.15 cluster.

### DIFF
--- a/deploy/kubernetes/manifest.yaml
+++ b/deploy/kubernetes/manifest.yaml
@@ -1,4 +1,3 @@
-
 ---
 
 apiVersion: v1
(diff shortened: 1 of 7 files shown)


deploy/kubernetes/manifest.yaml
```diff
@@ -118,107 +67,20 @@ spec:
...
-          env:
-            - name: ADDRESS
-              value: /var/lib/csi/sockets/pluginproxy/csi.sock
-          volumeMounts:
-            - name: socket-dir
-              mountPath: /var/lib/csi/sockets/pluginproxy/
-      volumes:
-        - name: socket-dir
-          emptyDir: {}
+  attachRequired: false
```
leakingtapan: Since you added skip-attach to the driver, could you add this to `What is this PR about? / Why do we need it?` of the PR description as well?

deploy/kubernetes/manifest.yaml
```diff
@@ -87,18 +43,11 @@ spec:
...
-          image: quay.io/k8scsi/driver-registrar:v0.4.2
-          imagePullPolicy: Always
+          image: quay.io/k8scsi/csi-node-driver-registrar:v1.1.0
           args:
             - --csi-address=$(ADDRESS)
-            - --mode=node-register
-            - --driver-requires-attachment=true
-            - --pod-info-mount-version="v1"
             - --kubelet-registration-path=$(DRIVER_REG_SOCK_PATH)
-            - --v=5
```
leakingtapan: What's the default log level for csi-node-driver-registrar?

wongma7: not sure, let me add this back as it's useful.

deploy/kubernetes/manifest.yaml
wongma7: yes, updated description to mention skip attach explicitly

deploy/kubernetes/manifest.yaml
wongma7: fixed

leakingtapan:
Because of skip attach relies on CSIDriver object, we need to highlight that this is backward incompatible change for 0.1 release (which is compatible with 1.13)

Issue #32, opened 2019-03-23
Add e2e tests


leakingtapan:
**Is your feature request related to a problem? Please describe.**
We should add e2e test for supported features:
 - [ ] static provisioning
 - [ ] encryption in transit

wongma7:
/assign
I'll write tests.
Integrating them with CI is a different story.

wongma7:
Running upstream e2e tests works, but we should skip Disruptive for now.
"static provisioning", we get for free.
"encryption in transit", maybe, depending on how much we want to test. Is checking that mount option "tls" is present sufficient?
```
Summarizing 2 Failures:

[Fail] EFS CSI [Driver: efs.csi.aws.com] [Testpattern: Pre-provisioned PV (default fs)] subPath [It] should unmount if pod is gracefully deleted while kubelet is down [Disruptive][Slow][LinuxOnly]
/Users/mattwon/go/src/github.com/kubernetes-sigs/aws-efs-csi-driver/test/e2e/vendor/k8s.io/kubernetes/test/e2e/storage/utils/utils.go:212

[Fail] EFS CSI [Driver: efs.csi.aws.com] [Testpattern: Pre-provisioned PV (default fs)] subPath [It] should unmount if pod is force deleted while kubelet is down [Disruptive][Slow][LinuxOnly]
/Users/mattwon/go/src/github.com/kubernetes-sigs/aws-efs-csi-driver/test/e2e/vendor/k8s.io/kubernetes/test/e2e/storage/utils/utils.go:212

Ran 24 of 142 Specs in 1192.911 seconds
FAIL! -- 22 Passed | 2 Failed | 0 Pending | 118 Skipped
--- FAIL: TestEFSCSI (1192.91s)
```

leakingtapan:
> Integrating them with CI is a different story.
SGTM. Let's create a different issue to track this if you are going to focus on e2e test case for this issue.

wongma7:
'should concurrently access the single volume from pods on different node' will cover RWX mode from multiple pods.

Will look into other ways to verify traffic is actually encrypted. May be tricky, might need to start a privileged pod on the node to for example check if there is an stunnel process for the mount

PR #55, opened 2019-07-19
Run upstream e2e tests 

wongma7:

**What is this PR about? / Why do we need it?** ref: https://github.com/kubernetes-sigs/aws-efs-csi-driver/issues/32. We should run all the existing upstream tests for storage drivers. They cover static provisioning, RWX mode from multiple pods, and more.

hack/run-e2e-test is based on the one in EBS CSI repo: https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/hack/run-e2e-test. It creates a kops cluster, creates an EFS file system, runs tests, and cleans up.

test/e2e/e2e_test.go is the tests themselves. It simply imports the upstream test suites and runs them.

depends on https://github.com/kubernetes-sigs/aws-efs-csi-driver/pull/48 and https://github.com/kubernetes-sigs/aws-efs-csi-driver/pull/51

Also depends on https://github.com/kubernetes/kubernetes/pull/80375

**What testing is done?** 
I edited the script to point to my s3 bucket instead of s3://k8s-kops-csi-e2e. Then, I ran `make test-e2e` as I would expect prow to.

Skipping slow tests (by manually editing hack/run-e2e-test script):
Ran 11 of 142 Specs in 330.736 seconds
SUCCESS! -- 11 Passed | 0 Failed | 0 Pending | 131 Skipped
--- PASS: TestEFSCSI (330.74s)
PASS
ok  	github.com/kubernetes-sigs/aws-efs-csi-driver/test/e2e	331.139s

Running all tests:
Ran 23 of 142 Specs in 1310.545 seconds
SUCCESS! -- 23 Passed | 0 Failed | 0 Pending | 119 Skipped
--- PASS: TestEFSCSI (1310.55s)
PASS
ok  	github.com/kubernetes-sigs/aws-efs-csi-driver/test/e2e	1310.917s

I am not sure if the prow AWS account has permissions to create EFS file systems, I ran the test using an admin account. The plan is to add the prow job and tweak it later.


wongma7:
```
An error occurred (AccessDeniedException) when calling the CreateFileSystem operation: User: arn:aws:iam:::user/awstester is not authorized to perform: elasticfilesystem:CreateFileSystem on the specified resource
```
now my script is stuck in a loop trying to delete...need to add a timeout.

wongma7:
e2e passed! https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_aws-efs-csi-driver/55/pull-aws-efs-csi-driver-e2e/1159269511698845697

Ran 23 of 141 Specs in 1316.602 seconds
SUCCESS! -- 23 Passed | 0 Failed | 0 Pending | 118 Skipped 

However, it's not separated by testcase in testgrid, so I'd like to figure that out: https://testgrid.k8s.io/sig-aws-efs-csi-driver#e2e-test-single-az

leakingtapan:
>However, it's not separated by testcase in testgrid, so I'd like to figure that out

see https://gcsweb.k8s.io/gcs/kubernetes-jenkins/logs/ci-gcp-compute-persistent-disk-csi-driver-latest-k8s-master-integration/1158851998234185732/artifacts/

you need to generate some junit xml files from your job

wongma7:
Yay, working now. Changes needed:
1. `-report-dir=$ARTIFACTS` in the `go test` command https://github.com/kubernetes-sigs/aws-efs-csi-driver/pull/55/files#diff-db2615cb8e63f2d009711c2dfd5088e4R111
2. Add the junit reporter https://github.com/kubernetes-sigs/aws-efs-csi-driver/pull/55/files#diff-6413e1dbafd6dd63b5d907ea3ae6f6f9R54

hack/run-e2e-test
```diff
@@ -0,0 +1,137 @@
...
+  ret=$?
+  if [[ $ret -eq 0 ]]; then
+    break
+  else
+    echo "Waiting cluster to be created"
+    sleep 30
+  fi
+done;
+
+# Push test driver image
```
leakingtapan: Let's move this section till line 75 to be before kops create cluster. This is to avoid driver build failed after cluster creation. And turn on `set -e`. See https://github.com/kubernetes-sigs/aws-ebs-csi-driver/pull/335/files#diff-db2615cb8e63f2d009711c2dfd5088e4R57 for a similar change

hack/run-e2e-test
```diff
@@ -0,0 +1,137 @@
...
+IMAGE_NAME=$AWS_ACCOUNT_ID.dkr.ecr.$REGION.amazonaws.com/aws-efs-csi-driver
+docker build -t $IMAGE_NAME:$IMAGE_TAG .
+docker push $IMAGE_NAME:$IMAGE_TAG
+
+echo "Deploying driver"
+cat deploy/kubernetes/manifest.yaml | sed s,amazon/aws-efs-csi-driver:latest,$IMAGE_NAME:$IMAGE_TAG, > $TEST_DIR/manifest.yaml
+kubectl apply -f $TEST_DIR/manifest.yaml
+
+echo "Creating EFS file system"
+aws efs create-file-system --creation-token $TEST_ID --tags Key=KubernetesCluster,Value=$CLUSTER_NAME.k8s.local --region $REGION
```
leakingtapan: I think this will make transition to the new test framework harder. Could we create the EFS fs in the go test code?

hack/run-e2e-test
```diff
@@ -0,0 +1,137 @@
...
+# Download dep if not yet
+if [[ ! -e $DEP_PATH ]]; then
+  mkdir -p $TEST_DIR
+  echo "Downloading dep from $DEP_DOWNLOAD_URL to $DEP_PATH"
+  curl -L -X GET $DEP_DOWNLOAD_URL -o $DEP_PATH
+  chmod +x $DEP_PATH
+fi
+
+pushd ./test/e2e
+$DEP_PATH ensure
```
leakingtapan: why is go dep needed?

test/e2e/e2e_test.go
```diff
@@ -0,0 +1,157 @@
...
+}
+
+func (e *efsDriver) SkipUnsupportedTest(testpatterns.TestPattern) {}
+
+func (e *efsDriver) PrepareTest(f *framework.Framework) (*testsuites.PerTestConfig, func()) {
+	ginkgo.By("Deploying EFS CSI driver")
+	cancelPodLogs := testsuites.StartPodLogs(f)
+
+	/* TODO deploy the driver before the test.
+	cleanup, err := f.CreateFromManifests(nil, "deploy/kubernetes/manifest.yaml")
```
leakingtapan: Is this needed?

test/e2e/e2e_test.go
```diff
@@ -0,0 +1,157 @@
...
+			ginkgo.By("Cleaning up EFS CSI driver")
+			/* TODO deploy the driver before the test.
+			cleanup()
+			*/
+			cancelPodLogs()
+		}
+}
+
+func (e *efsDriver) CreateVolume(config *testsuites.PerTestConfig, volType testpatterns.TestVolType) testsuites.TestVolume {
+	// TODO it's assumed that the EFS volumeHandle referred to in GetPersistentVolumeSource already exists. Instead, it could be dynamically created per test here?
```
leakingtapan: How about creating EFS filesystem inside here?

test/e2e/e2e_test.go
```diff
@@ -0,0 +1,157 @@
...
+	testsuites.InitVolumeModeTestSuite,
+	testsuites.InitSubPathTestSuite,
+	testsuites.InitProvisioningTestSuite,
+	//testsuites.InitSnapshottableTestSuite,
+	testsuites.InitMultiVolumeTestSuite,
+}
+
+var _ = ginkgo.Describe("[efs-csi] EFS CSI", func() {
+	driver := InitEFSCSIDriver()
+	ginkgo.Context(testsuites.GetDriverNameWithFeatureTags(InitEFSCSIDriver()), func() {
```
leakingtapan: Do we need to `InitEFSCSIDriver` twice?

wongma7: no, good catch, fixed

test/e2e/e2e_test.go
wongma7: No, removed; much simpler to do the build+push+deploy of driver in the script beforehand.

test/e2e/e2e_test.go
wongma7: I decided not to as there isn't much point. It will take a long time to create one per test and there's no need for every test to use its own EFS filesystem, it's fine for them to reuse the same one

hack/run-e2e-test
wongma7: Only way to vendor k/k into the test/e2e directory. I don't want to check in huge vendor so have to download dep and do dep ensure during test.

hack/run-e2e-test
wongma7: That's the plan but atm it makes sense to keep this in the script to reuse the various env vars

hack/run-e2e-test
wongma7: done

hack/run-e2e-test
```diff
@@ -0,0 +1,137 @@
...
+SSH_KEY_PATH=$TEST_DIR/id_rsa
+ssh-keygen -P csi-e2e -f $SSH_KEY_PATH
+
+$KOPS_PATH create cluster --state $KOPS_STATE_FILE \
+  --zones $ZONES \
+  --node-count=3 \
+  --kubernetes-version=$K8S_VERSION \
+  --ssh-public-key=$SSH_KEY_PATH.pub \
+  $CLUSTER_NAME.k8s.local
+  $KOPS_PATH get cluster --state $KOPS_STATE_FILE $CLUSTER_NAME.k8s.local -o yaml > $CLUSTER_YAML_PATH
```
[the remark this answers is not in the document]
wongma7: done, oops

hack/run-e2e-test
leakingtapan: I think it works for EBS to consume k/k with just go mod. See [here](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/go.mod#L102)

test/e2e/e2e_test.go
leakingtapan: Makes sense

hack/run-e2e-test
wongma7: yeah, but that is with 1.14 k/k. After 1.15, k/k moved to modules and I can't use go mod anymore. (Seems like it doesn't make sense, I know)

wongma7: More details: when k/k used godeps, it just symlinked stuff in staging dir into vendor dir. Now with go.mod, it uses some weird tricks with v0.0.0 https://github.com/kubernetes/kubernetes/blob/master/go.mod#L144 and `replace` directives and I've not figured out a way to import it using go mod.

leakingtapan:
/hold 

let me try if I can figure out how to import 1.15 kubernetes from go mod

leakingtapan:
Here is how you can import 1.15 k/k: https://github.com/kubernetes-sigs/aws-ebs-csi-driver/pull/345

test/e2e/hack/update-gomod.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/sh
```
leakingtapan: Could you add a copyright head similar to other scripts?

leakingtapan:
seems the test reporter for prow ARTIFACT is not working. could you check on that? the e2e test failed too

wongma7:
tests green: https://testgrid.k8s.io/sig-aws-efs-csi-driver#e2e-test

hack/run-e2e-test
```diff
@@ -0,0 +1,127 @@
...
+done
+
+# Run the test
+export KUBECONFIG=$HOME/.kube/config
+
+pushd ./test/e2e
+go test -v -timeout 0 ./... -kubeconfig=$HOME/.kube/config -report-dir=$ARTIFACTS -ginkgo.focus="\[efs-csi\]" -ginkgo.skip="\[Disruptive\]" \
+  -file-system-id=$FILE_SYSTEM_ID
+popd
+TEST_PASS=$?
```
leakingtapan: this will be the exit code `popd`?

wongma7: yes, dang, fixed.

Issue #45, opened 2019-06-03
Update ServiceAccount, ClusterRole and ClusterRoleBinding name


leakingtapan:
**Is your feature request related to a problem? Please describe.**
In order to run multiple CSI drivers within the cluster, they should use different names for ServiceAccount, ClusterRole and ClusterRoleBinding for the driver controller service and node service.

**Describe the solution you'd like in detail**
See EBS CSI driver's [manifest](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/deploy/kubernetes/manifest.yaml) as a reference.

leakingtapan:
This is only needed when dynamic provisioning is implemented.

Issue #31, opened 2019-03-23
Integration with k8s liveness probe


leakingtapan:
**Is your feature request related to a problem? Please describe.**
See https://github.com/kubernetes-csi/livenessprobe for more details

PR #62, opened 2019-07-24
Add liveness probe

wongma7:

**What is this PR about? / Why do we need it?** fixes https://github.com/kubernetes-sigs/aws-efs-csi-driver/issues/31 by adding https://github.com/kubernetes-csi/livenessprobe sidecar

also fixes https://github.com/kubernetes-sigs/aws-efs-csi-driver/issues/45 by prepending efs- to serviceaccount because I am touching the manifest here anyway.

**What testing is done?** on 1.14 cluster, created manifest. all containers report ready.

### DIFF
--- a/deploy/kubernetes/manifest.yaml
+++ b/deploy/kubernetes/manifest.yaml
@@ -3,7 +3,7 @@
 apiVersion: v1
 kind: ServiceAccount
 metadata:
-  name: csi-node-sa
+  name: efs-csi-node-sa
   namespace: kube-system
 
 ---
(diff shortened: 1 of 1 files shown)


deploy/kubernetes/manifest.yaml
```diff
@@ -63,6 +75,15 @@ spec:
               mountPath: /csi
             - name: registration-dir
               mountPath: /registration
+        - name: liveness-probe
+          imagePullPolicy: Always
+          image: quay.io/k8scsi/livenessprobe:v1.1.0
+          args:
+            - --csi-address=/csi/csi.sock
+            - --health-port=9809
```
leakingtapan: The default [probe-timeout](https://github.com/kubernetes-csi/livenessprobe/blob/v1.1.0/cmd/livenessprobe/main.go#L37) is 1 sec. Have you tested if this is long enough?

wongma7: yes, had no issue. I think socket taking >1s to respond when they are in the same pod would be worrisome.

PR #129, opened 2020-02-05
Force use of AWS nameservers

adammw:
Bug fix/suggestion. Unsure if it's applicable for all parties (ie. if someone expects this CSI driver to work on non-EC2 instances), but might be a solid default.

Depending on your default DNS configuration, the EFS DNS names may or may not be resolvable, whereas they will always be resolvable on the AWS DNS resolver. This change hardcodes DNS resolution for the efs plugin daemonset to the AWS resolver on the link-local address.

**What testing is done?** 
Tested on our own setup where DNS forwards to a non-AWS resolver where EFS domains did not resolve until this change.

### DIFF
--- a/deploy/kubernetes/base/node.yaml
+++ b/deploy/kubernetes/base/node.yaml
@@ -17,6 +17,10 @@ spec:
       nodeSelector:
         beta.kubernetes.io/os: linux
       hostNetwork: true
+      dnsPolicy: "None"
+      dnsConfig:
+        nameservers:
+          - 169.254.169.253
       priorityClassName: system-node-critical
       tolerations:
         - key: CriticalAddonsOnly


DmitriyStoyanov:
Thank you for this change, it solved my issue with private dns servers and unresolvable efs dns name.

nxf5025:
Confirmed this solved my issue as well

deploy/kubernetes/base/node.yaml
```diff
@@ -17,6 +17,10 @@ spec:
       nodeSelector:
         beta.kubernetes.io/os: linux
       hostNetwork: true
+      dnsPolicy: "None"
+      dnsConfig:
+        nameservers:
+          - 169.254.169.253
```
jqmichael, line 23: For my own note: 

>You can use the Amazon DNS server IP address 169.254.169.253, though some servers don't allow its use. Windows Server 2008, for example, disallows the use of a DNS server located in the 169.254.x.x network range.

https://docs.aws.amazon.com/vpc/latest/userguide/VPC_DHCP_Options.html#AmazonDNS

jqmichael: @adammw

tomhuang12:
Tested in our environment and works! Thank you!

buptliuwei:
It solved my issue,but i don't know how it works

PR #148, opened 2020-04-21
Move EFS tests into e2e.go so that they can be imported from elsewhere

wongma7:

**What is this PR about? / Why do we need it?** This is a small refactor to move the test code to e2e.go away from the test entrypoint code e2e_test.go. Slightly cleaner and enables the use-case where I import package test/e2e in another package to access InitEFSCSIDriver, ClusterName, and Region.

**What testing is done?** go test command still works and still accepts cluster-name/region flags.
```
 go test ./test/e2e/ -v -timeout=0 -kubeconfig=$HOME/.kube/config --cluster-name=asdf --region=us-west-2 --report-dir=./ARTIFACTS -ginkgo.focus="\[efs-csi\]" -ginkgo.skip="\[Disruptive\]"

W0420 17:40:20.768014   43285 test_context.go:418] Unable to find in-cluster config, using default host : http://127.0.0.1:8080
Apr 20 17:40:20.768: INFO: The --provider flag is not set. Continuing as if --provider=skeleton had been used.
=== RUN   TestEFSCSI
2020-04-20 17:40:20.768572 I | Starting e2e run "f5a3c398-be21-4e4e-b49c-0ae849bde957" on Ginkgo node 1
Running Suite: EFS CSI Suite
============================
Random Seed: 1587429620 - Will randomize all specs
Will run 151 of 371 specs

STEP: Creating EFS filesystem at us-west-2 on asdf
...
```

### DIFF
--- a/test/e2e/e2e.go
+++ b/test/e2e/e2e.go
@@ -1,0 +1,119 @@
+package e2e
+
+import (
+	"fmt"
+
+	"github.com/onsi/ginkgo"
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/kubernetes/test/e2e/framework"
+	"k8s.io/kubernetes/test/e2e/storage/testpatterns"
+	"k8s.io/kubernetes/test/e2e/storage/testsuites"
+)
+
+var (
+	ClusterName  string
+	Region       string
+	fileSystemId string
+)
+
+type efsDriver struct {
+	driverInfo testsuites.DriverInfo
+}
+
+var _ testsuites.TestDriver = &efsDriver{}
+
+// TODO implement Inline (unless it's redundant) and DynamicPV
+// var _ testsuites.InlineVolumeTestDriver = &efsDriver{}
+var _ testsuites.PreprovisionedPVTestDriver = &efsDriver{}
+
+func InitEFSCSIDriver() testsuites.TestDriver {
+	return &efsDriver{
+		driverInfo: testsuites.DriverInfo{
+			Name:                 "efs.csi.aws.com",
+			SupportedFsType:      sets.NewString(""),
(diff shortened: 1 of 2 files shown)


test/e2e/e2e.go
```diff
@@ -0,0 +1,119 @@
...
+	if err != nil {
+		ginkgo.Fail(fmt.Sprintf("failed to create EFS filesystem: %s", err))
+	}
+	fileSystemId = id
+})
+
+var _ = ginkgo.AfterSuite(func() {
+	ginkgo.By(fmt.Sprintf("Deleting EFS filesystem %s", fileSystemId))
+	if len(fileSystemId) != 0 {
+		c := NewCloud(Region)
```
jqmichael, line 106: I think we can move the `fileSystemId` check to BeforeSuite. We shouldn't start tests if the `fileSystemId` is empty.

jqmichael:
lgtm

test/e2e/e2e.go
wongma7: BeforeSuite will always either create fileSystemId or fail (no tests will start) so I think it is OK. Anyway the followup pr complicates it more :/ https://github.com/kubernetes-sigs/aws-efs-csi-driver/pull/149/files#diff-53314d6a19fc22810420c6bee965f5b8R120

leakingtapan:
> Slightly cleaner and enables the use-case where I import package test/e2e in another package to access InitEFSCSIDriver, ClusterName, and Region.

Is there any pkg other than this driver itself will import those values?

wongma7:
> Is there any pkg other than this driver itself will import those values?

I am trying out a wrapper pkg that will import this, ebs, fsx, etc. and run all the tests at once in the same "suite" ( as opposed to doing `go test ebs/test/e2e`, then `go test fsx/test/e2e`, etc.) so I want the option to set these values via code instead of via flags.

PR #149, opened 2020-04-21
Move BeforeSuite EFS creation to BeforeEach and add driver deployment BeforeEach

wongma7:

**What is this PR about? / Why do we need it?** See second commit only ... another minor cleanup/refactor that depends on https://github.com/kubernetes-sigs/aws-efs-csi-driver/pull/148/files.

If I import these tests here into my own ginkgo test suite, `BeforeSuite` always runs even if I don't want to focus on these tests (e.g. I set --ginkgo.skip=EFS). To avoid this, move EFS creation to BeforeEach so that it only runs if one of these tests runs. To avoid multiple attempts creating the same EFS volume, check variable fileSystemId

**What testing is done?** Without this change, my test suite (` go test ... --ginkgo-skip=EFS`) fails with EFS creation failure. With this change, my test suite doesn't try to create an EFS since I am skipping EFS tests anyway.

### DIFF
--- a/test/e2e/e2e.go
+++ b/test/e2e/e2e.go
@@ -87,21 +87,14 @@ var csiTestSuites = []func() testsuites.TestSuite{
 	testsuites.InitMultiVolumeTestSuite,
 }
 
-var _ = ginkgo.BeforeSuite(func() {
-	ginkgo.By(fmt.Sprintf("Creating EFS filesystem in region %q for cluster %q", Region, ClusterName))
-	if Region == "" || ClusterName == "" {
-		ginkgo.Fail("failed to create EFS filesystem: both Region and ClusterName must be non-empty")
-	}
-	c := NewCloud(Region)
-	id, err := c.CreateFileSystem(ClusterName)
-	if err != nil {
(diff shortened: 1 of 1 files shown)


wongma7:
This is a workaround because ginkgo does not have a "BeforeAll" between "BeforeSuite" and "BeforeEach" :/ https://github.com/onsi/ginkgo/issues/70.

test/e2e/e2e.go
```diff
@@ -127,6 +134,28 @@ var _ = ginkgo.Describe("[efs-csi] EFS CSI", func() {
 		fileSystemId = id
 	})
 
+	var once sync.Once
```
wongma7, line 137: This synchronization turned out to be pointless because the other BeforeEach/AfterSuite blocks are not yet synchronized, so running tests in parallel usually fails with "AlreadyExists" type errors anyway. But I'm going to leave it here, I'm working on another PR to fix it. The CI runs tests serially so it makes no difference at the moment.

wongma7:
closing in favor of https://github.com/kubernetes-sigs/aws-efs-csi-driver/pull/160

PR #150, opened 2020-04-21
 Use provider-agnostic tags/names to identify node subnets and security group

wongma7:
**Is this a bug fix or adding new feature?** feature

**What is this PR about? / Why do we need it?** This makes it possible to run the tests on EKS which unfortunately doesn't tag node subnets/security group the same way kops does.

**What testing is done?** Ran test on EKS. Before it failed because it could not find the node security group to create mount targets with. Now it works.

### DIFF
--- a/test/e2e/cloud.go
+++ b/test/e2e/cloud.go
@@ -19,6 +19,7 @@ package e2e
 import (
 	"errors"
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/aws/aws-sdk-go/aws"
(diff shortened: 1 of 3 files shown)


wongma7:
Split getSecurityGroup in to getKopsSecurityGroup and getEksSecurityGroup to be more explicit about assumptions

wongma7:
rebased on top of #148 and #149 just to save future troubles... those will need to merge first of course.

wongma7:
closing in favor of https://github.com/kubernetes-sigs/aws-efs-csi-driver/pull/160

PR #153, opened 2020-04-27
Add example for Access Points

2uasimojo:
Adds a README and sample specs for exposing multiple independent data
sources on the same EFS volume by creating and mounting access points.

Neither, really. It's adding docs for a thing that already works.

There are questions about how to manage the permissions of the mounts, which by default are inaccessible because they're owned by `root:root` with 755 permissions. Changing permissions of the root export and creating subdirectories both require manual intervention by a powerful user. This provides an alternative that does not (though it does require the ability to administer EFS itself).

**What testing is done?** 
Well, this is a doc, so it's not tested per se. But I've been using the outlined procedure under OpenShift Dedicated 4.3.

### DIFF
--- a/examples/kubernetes/access_points/README.md
+++ b/examples/kubernetes/access_points/README.md
@@ -6,7 +6,10 @@ In this case, the separation is managed on the EFS side rather than the kubernet
 
 ### Create Access Points (in EFS)
 Following [this doc](https://docs.aws.amazon.com/efs/latest/ug/create-access-point.html), create a separate access point for each independent data store you wish to expose in your cluster, tailoring the ownership and permissions as desired.
-Note that there's no need to use different EFS volumes.
+There is no need to use different EFS volumes.
+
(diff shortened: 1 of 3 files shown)


examples/kubernetes/access_points/README.md
```diff
@@ -0,0 +1,78 @@
+## EFS Access Points
```
leakingtapan, line 1: I think certain IAM permission is required in order to mount access point, What's the IAM permission used in this case? And how is the IAM permission assigned? Using instance profile or IAM for SA?

2uasimojo: The driver (under the DaemonSet) is already running with godlike powers; I don't think anything additional is required. At least, I haven't run into any issues. But in any case I would welcome someone confirming my findings by walking through the doc :)

leakingtapan: Let’s confirm on that and document here. As a minimum permission, this driver doesn’t require any IAM permission for it to work

2uasimojo: We confirmed another way that -- by default -- no IAM permissions are required. I've noted that it's possible to set up additional policies.

leakingtapan: LGTM, thx for confirming!

Issue #152, opened 2020-04-27
Please tag a new version


2uasimojo:
/kind bug

Requesting a new tagged image (`release-0.4`) built against an `efs-utils` >= 1.23.

We will use `latest` for the time being, but would like to be able to use a "frozen" version as soon as feasible. Thanks in advance!

**What happened?**
When I try to use the `tls` `mountOption` the `mount` command fails with

```
  Warning  FailedMount  27s (x7 over 61s)  kubelet, ip-10-0-130-137.us-west-1.compute.internal  MountVolume.SetUp failed for volume "efs-pv" : kubernetes.io/csi: mounter.SetupAt failed: rpc error: code = Internal desc = Could not mount "fs-c3b5e0da:/" at "/var/lib/kubelet/pods/96f1ea12-4d2b-400c-85c8-72853cb8ef17/volumes/kubernetes.io~csi/efs-pv/mount": mount failed: exit status 1
Mounting command: mount
Mounting arguments: -t efs -o tls,accesspoint=fsap-01b2c12c8578dceff fs-c3b5e0da:/ /var/lib/kubelet/pods/96f1ea12-4d2b-400c-85c8-72853cb8ef17/volumes/kubernetes.io~csi/efs-pv/mount
Output: Traceback (most recent call last):
  File "/sbin/mount.efs", line 1375, in <module>
    main()
  File "/sbin/mount.efs", line 1369, in main
    mount_tls(config, init_system, dns_name, path, fs_id, mountpoint, options)
  File "/sbin/mount.efs", line 1302, in mount_tls
    with bootstrap_tls(config, init_system, dns_name, fs_id, mountpoint, options) as tunnel_proc:
  File "/lib64/python2.7/contextlib.py", line 17, in __enter__
    return self.gen.next()
  File "/sbin/mount.efs", line 705, in bootstrap_tls
    security_credentials, ap_id=ap_id, base_path=state_file_dir)
  File "/sbin/mount.efs", line 848, in create_certificate
    private_key = check_and_create_private_key(base_path)
  File "/sbin/mount.efs", line 911, in check_and_create_private_key
    do_with_lock(generate_key)
  File "/sbin/mount.efs", line 894, in do_with_lock
    return function()
  File "/sbin/mount.efs", line 907, in generate_key
    subprocess_call(cmd, 'Failed to create private key')
  File "/sbin/mount.efs", line 970, in subprocess_call
[...]

PR #154, opened 2020-04-28
Pin dependency library versions

2uasimojo:
To make image builds more predictable, this commit pins the versions of
the two libraries pulled in by the docker build:
- util-linux
- amazon-efs-utils

This way we don't get surprised by changes in the underlying libs -- we
move along explicitly and deliberately.

Neither, just hardening the image build process.

Recently when trying to debug a problem mounting access points (see https://github.com/kubernetes-sigs/aws-efs-csi-driver/issues/152) one of our wild goose chases involved figuring out what version of amazon-efs-utils was built into the driver image at various times during our testing. Having the version pinned in the dockerfile for a given driver release would have removed the guesswork.

And in general this is just a good idea.

**What testing is done?** 
Local `make image-release`, run (overriding entrypoint with a shell), verify installed versions with `yum`.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -24,7 +24,7 @@ ADD . .
 RUN make aws-efs-csi-driver
 
 FROM amazonlinux:2
-RUN yum install util-linux amazon-efs-utils -y
+RUN yum install util-linux-2.30.2-2.amzn2.0.4.x86_64 amazon-efs-utils-1.24-4.amzn2.noarch -y
 
 # Default client source is k8s which can be overriden with –build-arg when building the Docker image
 ARG client_source=k8s
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,7 +23,7 @@ RUN go mod download
 ADD . .
 RUN make aws-efs-csi-driver
 
-FROM amazonlinux:2
+FROM amazonlinux:2.0.20200406.0
 RUN yum install util-linux-2.30.2-2.amzn2.0.4.x86_64 amazon-efs-utils-1.24-4.amzn2.noarch -y
 
 # Default client source is k8s which can be overriden with –build-arg when building the Docker image


Dockerfile
```diff
@@ -18,7 +18,7 @@ ADD . .
 RUN make
 
 FROM amazonlinux:2
-RUN yum install util-linux amazon-efs-utils -y
+RUN yum install util-linux-2.30.2-2.amzn2.0.4.x86_64 amazon-efs-utils-1.24-4.amzn2.noarch -y
```
jharrington22, line 21: Moving forward these versions may be dropped from the repository. Do we need to pin the `FROM` image too assuming the repo URLs are specific for minor releases otherwise it wouldn't help.

2uasimojo:
Sorry for my ignorance @leakingtapan: I can't tell if [this build failure](https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_aws-efs-csi-driver/154/pull-aws-efs-csi-driver-e2e/1255532456350584833#1:build-log.txt%3A17) is spurious, or if it's happening because the checksum of the driver image has changed. The latter would be expected -- but I could use some help identifying where the expected checksum would need to be updated. (And I might also submit a PR to put that explanation into the text for this build failure to help future me.)

2uasimojo:
/test pull-aws-efs-csi-driver-e2e
...to see if this is for real.

2uasimojo:
> Sorry for my ignorance @leakingtapan: I can't tell if [this build failure](https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_aws-efs-csi-driver/154/pull-aws-efs-csi-driver-e2e/1255532456350584833#1:build-log.txt%3A17) is spurious, or if it's happening because the checksum of the driver image has changed.

Spurious. It passed on a retest. Disregard.

Dockerfile
wongma7: Right, it doesn't seem to help, for example try base image 20200207 released prior to 1.24 and it still needs to get updates from the amzn2-core repository where only 1 version is available:
```
sh-4.2# yum list available amazon-efs-utils
Loaded plugins: ovl, priorities
Available Packages
amazon-efs-utils.noarch                   1.24-4.amzn2                    amzn2-core
```

So any time a new package of amazon-efs-utils replaces the old, builds will fail (like as part of PR CI)? That will be painful for contributors.

Alternatively, we could log the amazon-efs-utils version with `mount.efs --version`  https://github.com/kubernetes-sigs/aws-efs-csi-driver/blob/be96b1d3f2c93633adf90899321c45f96c4fc70f/pkg/driver/version.go#L46 to help with the debugging/guesswork part of this issue.

2uasimojo: It sounds like there's not a perfect answer here; we're going to have to choose the lesser evil.

I think we really only need to worry about the dep versions being removed (do they really do that?) during the lifecycle of `latest`. Once we tag a driver release, we're effectively freezing at whatever's available at that time; the bits stick to the image even if the repo is changed. If something happens where we need to update a tagged version (backporting a bugfix, for example), we can bump the lib versions at the same time if necessary -- the point being that we're always using a specific known working version.

And yes, developers working on `latest` could be inconvenienced if a lib version goes away, but IMO that's a good thing. As it stands, if something breaks while I'm developing, I don't know whether it's my code or because the anonymous lib version moved underneath me, and that's hard to discover. If the library version is pinned, that source of error is eliminated. If that version disappears, the failure (in CI or locally) is easy to understand. Yes, it requires a new/separate PR to fix, but IMO that's worth the trouble.

2uasimojo:
I've rebased and added a commit to pin the amazonlinux image as well. Can we continue [this discussion](https://github.com/kubernetes-sigs/aws-efs-csi-driver/pull/154#discussion_r426686402)?

2uasimojo:
The pull-aws-efs-csi-driver-e2e job seems unhappy, but I don't think it's my PR. I'll wait a bit before retesting.

Dockerfile
wongma7: I forgot to set --show-duplicates:
```
$ yum --showduplicates list amazon-efs-utils
Loaded plugins: priorities, update-motd
Installed Packages
amazon-efs-utils.noarch                         1.25-2.amzn2                         @amzn2-core
Available Packages
amazon-efs-utils.noarch                         1.0-1.amzn2                          amzn2-core 
amazon-efs-utils.noarch                         1.1-1.amzn2                          amzn2-core 
amazon-efs-utils.noarch                         1.2-1.amzn2                          amzn2-core 
amazon-efs-utils.noarch                         1.3-1.amzn2                          amzn2-core 
amazon-efs-utils.noarch                         1.4-1.amzn2                          amzn2-core 
amazon-efs-utils.noarch                         1.5-1.amzn2                          amzn2-core 
amazon-efs-utils.noarch                         1.6-1.amzn2                          amzn2-core 
amazon-efs-utils.noarch                         1.7-1.amzn2                          amzn2-core 
amazon-efs-utils.noarch                         1.10-1.amzn2                         amzn2-core 
amazon-efs-utils.noarch                         1.18-1.amzn2                         amzn2-core 
amazon-efs-utils.noarch                         1.21-2.amzn2                         amzn2-core 
amazon-efs-utils.noarch                         1.23-2.amzn2                         amzn2-core 
amazon-efs-utils.noarch                         1.24-2.amzn2                         amzn2-core 
amazon-efs-utils.noarch                         1.24-3.amzn2                         amzn2-core 
amazon-efs-utils.noarch                         1.24-4.amzn2                         amzn2-core 
amazon-efs-utils.noarch                         1.25-1.amzn2                         amzn2-core 
amazon-efs-utils.noarch                         1.25-2.amzn2                         amzn2-core 
amazon-efs-utils.noarch                         1.25-3.amzn2                         amzn2-core
```
So my point is moot. They don't seem to remove old versions.

wongma7:
/lgtm
/approve

I'm going to create a follow up PR to bump to 1.25-3 as that's the latest and I want to get that tested ASAP.

2uasimojo:
Thanks @wongma7 

> I'm going to create a follow up PR to bump to 1.25-3 as that's the latest and I want to get that tested ASAP.

Cool, would you please @ me when you do? I'd like to watch that and test with it as well.

PR #156, opened 2020-05-06
WIP: Mount with tls mount option by default via new encryptInTransit VolumeContext field that defaults true

wongma7:

**What is this PR about? / Why do we need it?** We would like encryption of data in transit to be the default. Currently, encryption is turned on/off by adding/omitting the "tls" mount option to/from PV spec.mountOptions. With this PR, whether a "tls" mount option is present in PV spec.mountOptions or not, the value of `encryptInTransit` will ultimately determine whether encryption in transit is on, i.e. whether the *final* mount command contains the "tls" mount option or not.

Intended behavior:

| | encryptInTransit is true | encryptInTransit is false |
| ------------- | ------------- | -- |
| "tls" is in mountOptions  | encryption  | NO encryption |
| "tls" isn't in mountOptions | encryption  | NO encryption |

TODO: update docs/examples

**What testing is done?**  TODO: test

### DIFF
--- a/pkg/driver/node.go
+++ b/pkg/driver/node.go
@@ -22,6 +22,7 @@ import (
 	"os"
 	"path"
 	"path/filepath"
+	"strconv"
 	"strings"
 
 	"github.com/container-storage-interface/spec/lib/go/csi"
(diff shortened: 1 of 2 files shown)


wongma7:
closing in favor of https://github.com/kubernetes-sigs/aws-efs-csi-driver/pull/205

PR #158, opened 2020-05-07
Updated node.yaml to update deprecated node selectors

nithu0115:
Updated node.yaml to remove deprecated labels

As per documentation, `kubernetes-io-arch` and `kubernetes.io/os` node selectors are deprecated.
https://v1-14.docs.kubernetes.io/docs/reference/kubernetes-api/labels-annotations-taints/#beta-kubernetes-io-arch-deprecated. Removed deprecated node selectors. 

**What testing is done?** 
N/A

### DIFF
--- a/deploy/kubernetes/base/node.yaml
+++ b/deploy/kubernetes/base/node.yaml
@@ -15,8 +15,8 @@ spec:
         app: efs-csi-node
     spec:
       nodeSelector:
-        beta.kubernetes.io/os: linux
-        beta.kubernetes.io/arch: amd64
+        kubernetes.io/os: linux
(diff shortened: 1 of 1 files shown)


wongma7:
/lgtm

Since the yaml is for latest/stable(0.5.0) which isn't supported on kubernetes < 1.14 I don't think we need to worry about having the beta labels, can move to the new labels.

jqmichael:
Today EKS officially supports v1.14+ masters. However, since worker nodes can be 2 versions behind, we shouldn't break worker nodes who are still using 1.12 or 1.13 version, where `kubernetes.io/arch` wasn't available.

PR #160, opened 2020-05-09
Enable e2e testing on EKS, parallelism, and make optional EFS creation/driver deployment

wongma7:

**What is this PR about? / Why do we need it?** When I tried to run the tests on EFS in parallel I ran into a couple of issues. This is the sum of fixes I needed to make to get the tests to pass.

EFS creation queries kops-specific tags, so change cloud.go a bit to also check for EKS-style tags.

Since multiple tests can be run if I run by doing `ginkgo -p`, I found it's best to move EFS creation/driver deployment to SynchronizedBefore/AfterSuite. Moreover, I made it optional since outside of the CI environment there may be an existing EFS/driver that the test can use.

**What testing is done?** I ran the tests against my EKS cluster.


test/e2e/cloud.go
```diff
@@ -142,19 +179,63 @@ func (c *cloud) getSecurityGroup(clusterName string) (string, error) {
...
+	if len(response.SecurityGroups) == 0 {
+		return "", fmt.Errorf("no security group found for cluster %s", clusterName)
+	}
+
+	groupId := aws.StringValue(response.SecurityGroups[0].GroupId)
+	for _, sg := range response.SecurityGroups {
+		if strings.Contains(strings.ToLower(*sg.GroupName), "node") {
+			groupId = aws.StringValue(sg.GroupId)
+		}
+	}
```
jqmichael, line 227: break

test/e2e/e2e.go
```diff
@@ -87,27 +99,71 @@ var csiTestSuites = []func() testsuites.TestSuite{
 	testsuites.InitMultiVolumeTestSuite,
 }
 
-var _ = ginkgo.BeforeSuite(func() {
-	ginkgo.By(fmt.Sprintf("Creating EFS filesystem in region %q for cluster %q", Region, ClusterName))
-	if Region == "" || ClusterName == "" {
-		ginkgo.Fail("failed to create EFS filesystem: both Region and ClusterName must be non-empty")
+var _ = ginkgo.SynchronizedBeforeSuite(func() []byte {
+	// Validate parameters
```
jqmichael, line 103: For my notes:

>SynchronizedBeforeSuite blocks are primarily meant to solve the problem of setting up singleton external resources shared across nodes when running tests in parallel. For example, say you have a shared database that you can only start one instance of that must be used in your tests. When running in parallel, only one node should set up the database and all other nodes should wait until that node is done before running.

>SynchronizedBeforeSuite accomplishes this by taking *two* function arguments. The first is only run on parallel node #1. The second is run on all nodes, but *only* after the first function completes successfully. Ginkgo also makes it possible to send data from the first function (on Node 1) to the second function (on all the other nodes).

https://godoc.org/github.com/onsi/ginkgo#SynchronizedBeforeSuite

test/e2e/cloud.go
```diff
@@ -58,12 +59,19 @@ func (c *cloud) CreateFileSystem(clusterName string) (string, error) {
 		Tags:          tags,
 	}
 
+	var fileSystemId *string
 	response, err := c.efsclient.CreateFileSystem(request)
 	if err != nil {
-		return "", err
+		switch t := err.(type) {
+		case *efs.FileSystemAlreadyExists:
```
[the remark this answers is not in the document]
jqmichael: `Error()` is an an implementation to the pointer receiver of `FileSystemAlreadyExists`. 

`func (s *FileSystemAlreadyExists) Error() string {`

PR #165, opened 2020-05-14
Added client source for business tracking purpose

jqmichael:
This is a new feature request from EFS team to report client source for business tracking purpose by appending a client info section in the `/etc/amazon/efs/efs-utils.conf` file.  

**What testing is done?** 
1. When not passing in `--build-arg` flag, use the default client source. 
```
docker build -t 705114051414.dkr.ecr.us-west-2.amazonaws.com/amazon/aws-efs-csi-driver:latest .

# launch a container and check the content of `efs-utils.conf` file.
docker run --name aws-efs-csi-driver --rm  --entrypoint="" 705114051414.dkr.ecr.us-west-2.amazonaws.com/amazon/aws-efs-csi-driver cat /etc/amazon/efs/efs-utils.conf
```
```
...

# Set client auth/access point certificate renewal rate. Minimum value is 1 minute.
tls_cert_renewal_interval_min = 60

[client-info] 
source=k8s
```

2. When building the image and passing client_source from `--build-arg`.

```
docker build -t 705114051414.dkr.ecr.us-west-2.amazonaws.com/amazon/aws-efs-csi-driver:latest --build-arg client_source=eks.ec2 .

# launch a container and check the content of `efs-utils.conf` file.
docker run --name aws-efs-csi-driver --rm  --entrypoint="" 705114051414.dkr.ecr.us-west-2.amazonaws.com/amazon/aws-efs-csi-driver cat /etc/amazon/efs/efs-utils.conf
```
```
...

# Set client auth/access point certificate renewal rate. Minimum value is 1 minute.
tls_cert_renewal_interval_min = 60

[client-info] 
source=eks.ec2
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,11 +14,27 @@
 
 FROM golang:1.13.4-stretch as builder
 WORKDIR /go/src/github.com/kubernetes-sigs/aws-efs-csi-driver
+
+# Cache go modules
+COPY go.mod .
+COPY go.sum .
+RUN go mod download
+
 ADD . .
-RUN make
+RUN make aws-efs-csi-driver
 
 FROM amazonlinux:2
 RUN yum install util-linux amazon-efs-utils -y
+
+# Default client source is k8s which can be overriden with –build-arg when building the Docker image
+ARG client_source=k8s
+RUN echo "client_source:${client_source}"
+RUN printf "\n\
+\n\
+[client-info] \n\
+source=${client_source} \n\
+" >> /etc/amazon/efs/efs-utils.conf
+
 COPY --from=builder /go/src/github.com/kubernetes-sigs/aws-efs-csi-driver/bin/aws-efs-csi-driver /bin/aws-efs-csi-driver
 COPY THIRD-PARTY /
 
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -14,11 +14,27 @@
 
 FROM golang:1.13.4-stretch as builder
 WORKDIR /go/src/github.com/kubernetes-sigs/aws-efs-csi-driver
+
+# Cache go modules
+COPY go.mod .
+COPY go.sum .
+RUN go mod download
+
```
jqmichael, line 22: So that we don't have to download go modules every time building the application.

jqmichael: More details of this trick can be found in this article. 
https://medium.com/@petomalina/using-go-mod-download-to-speed-up-golang-docker-builds-707591336888

> Go 1.11 introduces the go mod downloadcommand, which takesgo.modand go.sumfiles and downloads the dependencies from them instead of using the source code. As these files don’t change frequently (unless you are updating the dependencies), they can be simply cached by the COPYcommand from Dockerfile.

```
COPY go.mod . # <- COPY go.mod and go.sum files to the workspace
COPY go.sum .

# Get dependancies - will also be cached if we won't change mod/sum
RUN go mod download
# COPY the source code as the last step
COPY . .
```

Issue #117, opened 2019-12-31
Established NFS connections hang after network connectivity interrupted


otakumike:
/kind bug

**What happened?**
One of our nodes had a very brief failure in which Calico seemed to restart. We're not sure of the root cause here, but momentarily Calico was lost to us and then healed itself: 

```
calico-node	2019-12-31 03:28:30.652 [WARNING][55] health.go 190: Reporter failed readiness checks name="int_dataplane" reporter-state=&health.reporterState{name:"int_dataplane", reports:health.HealthReport{Live:true, Ready:true}, timeout:20000000000, latest:health.HealthReport{Live:true, Ready:true}, timestamp:time.Time{wall:0xbf7acdf1ac9ae5a5, ext:186458634869, loc:(*time.Location)(0x2b29100)}}
calico-node	2019-12-31 03:28:30.652 [WARNING][55] health.go 184: Reporter failed liveness checks name="int_dataplane" reporter-state=&health.reporterState{name:"int_dataplane", reports:health.HealthReport{Live:true, Ready:true}, timeout:20000000000, latest:health.HealthReport{Live:true, Ready:true}, timestamp:time.Time{wall:0xbf7acdf1ac9ae5a5, ext:186458634869, loc:(*time.Location)(0x2b29100)}}
calico-node	2019-12-31 03:28:30.652 [INFO][55] health.go 178: Overall health status changed lastSummary=&health.HealthReport{Live:false, Ready:false}
```

After Calico failed, the EFS CSI daemonset running on this node reported an issue connecting to the unix socket: 

```
E1231 03:28:41.720414       1 connection.go:129] Lost connection to unix:///csi/csi.sock.
```

Even though Calico came back up a few seconds later and the node went on working like normal, all established EFS connections hung. On the host, things like `df` never completed running and the containers that were using these EFS mounts would fail to terminate. 

**What you expected to happen?**
When the EFS CSI driver loses connections to the Unix socket, it should try to recover lost NFS connections or destroy them, forcing pods to be rescheduled.

**How to reproduce it (as minimally and precisely as possible)?**
Interrupt network connectivity on a node briefly, then re-connect. However, I have not tested this.  

**Anything else we need to know?**:
I was able to recover from this issue by forcefully 'umount'ing the hanging NFS connections on the host. Nothing needed to be done to get new NFS connections running on the host, it was only those connections that were active when the socket failed which were broken. 

**Environment**
[...]

Issue #112, opened 2019-12-12
EFS volume cannot unmount because device is busy


houdinisparks:
/kind bug

**What happened?**
During rolling updated, 1 of the pods is stuck in terminating status due to error below"
```
Dec 12 18:57:22 ip-100-64-36-72.ap-southeast-1.compute.internal kubelet[2577]: I1212 18:57:22.720521
    2577 reconciler.go:181] operationExecutor.UnmountVolume started for volume "efs-claim" (UniqueNa
me: "kubernetes.io/csi/efs.csi.aws.com^fs-51aad710") pod "7a84d253-11c9-11ea-b310-066b5c788196" (UID
: "7a84d253-11c9-11ea-b310-066b5c788196")
Dec 12 18:57:22 ip-100-64-36-72.ap-southeast-1.compute.internal kubelet[2577]: I1212 18:57:22.720669
    2577 clientconn.go:440] parsed scheme: ""
Dec 12 18:57:22 ip-100-64-36-72.ap-southeast-1.compute.internal kubelet[2577]: I1212 18:57:22.720689
    2577 clientconn.go:440] scheme "" not registered, fallback to default scheme
Dec 12 18:57:22 ip-100-64-36-72.ap-southeast-1.compute.internal kubelet[2577]: I1212 18:57:22.720746
    2577 asm_amd64.s:1337] ccResolverWrapper: sending new addresses to cc: [{/var/lib/kubelet/plugins/efs.csi.aws.com/csi.sock 0  <nil>}]
Dec 12 18:57:22 ip-100-64-36-72.ap-southeast-1.compute.internal kubelet[2577]: I1212 18:57:22.720764    2577 clientconn.go:796] ClientConn switching balancer to "pick_first"
Dec 12 18:57:22 ip-100-64-36-72.ap-southeast-1.compute.internal kubelet[2577]: I1212 18:57:22.720813    2577 balancer_conn_wrappers.go:131] pickfirstBalancer: HandleSubConnStateChange: 0xc0003f24b0, CONNECTING
Dec 12 18:57:22 ip-100-64-36-72.ap-southeast-1.compute.internal kubelet[2577]: I1212 18:57:22.720846    2577 clientconn.go:1016] blockingPicker: the picked transport is not ready, loop back to repick
[...]

PR #166, opened 2020-05-15
Always enabled encryption in transit

jqmichael:
This PR is a feature request to always enable [encryption in transit](https://docs.aws.amazon.com/efs/latest/ug/encryption-in-transit.html) by always passing in `tls` as mount options. 

Using encryption in transit, data will be encrypted during its transition over the network to the EFS service. This provides an extra layer of defence-in-depth for applications that requires strict security compliance.

Encryption in transit was not always enabled in EFS mount helper because EFS can't control the client OS and some versions of Linux don't include a version of stunnel (a service providing TLS tunneling) that supports these TLS features by default. In the EFS CSI Driver case, we have the luxury to always run the driver in Amazon Linux 2 container which includes a newer version of stunnel. 

**What testing is done?** 
1. When running the static provisioner example where `tls` was not set as mount option, verified in EFS CSI Driver log `tls` was set. 
```
I0514 22:09:14.623830       1 node.go:108] NodePublishVolume: mounting fs-b6654c1c:/ at /var/lib/kubelet/pods/9de5c5d3-1a1f-4ecb-b522-2c538dc2709d/volumes/kubernetes.io~csi/efs-pv/mount with options [tls]
```
2. Exec into efs-csi-node pod and verified stunnel process was started. 

```
kubectl exec $(kubectl get po -l app=efs-csi-node -n kube-system -o jsonpath='{.items[0].metadata.name}') -n kube-system -c efs-plugin -it sh

sh-4.2# ps -ef
UID        PID  PPID  C STIME TTY          TIME CMD
root         1     0  0 23:08 ?        00:00:00 /bin/aws-efs-csi-driver --endpoint=unix:/csi/csi.sock --logtostderr --
root        14     1  0 23:08 ?        00:00:00 python /usr/bin/amazon-efs-mount-watchdog
root       102     1  0 23:11 ?        00:00:00 stunnel /var/run/efs/stunnel-config.fs-b6654c1c.var.lib.kubelet.pods.5
```

3. Set up in the efs filesytem to enforce in-transit encryption for all clients and verified the example still works.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -24,7 +24,9 @@ The following CSI interfaces are implemented:
 ### Encryption In Transit
 One of the advantages of using EFS is that it provides [encryption in transit](https://aws.amazon.com/blogs/aws/new-encryption-of-data-in-transit-for-amazon-efs/) support using TLS. Using encryption in transit, data will be encrypted during its transition over the network to the EFS service. This provides an extra layer of defence-in-depth for applications that requires strict security compliance.
 
-To enable encryption in transit, `tls` needs to be set in the `NodePublishVolumeRequest.VolumeCapability.MountVolume` object's `MountFlags` fields. For an example of using it in kubernetes, see the persistence volume manifest in [Encryption in Transit Example](../examples/kubernetes/encryption_in_transit/specs/pv.yaml)
+For EFS CSI Driver EFS CSI Driver v0.3.0 or earlier releases, this feature is optional. To enable encryption in transit, `tls` needs to be set in the `NodePublishVolumeRequest.VolumeCapability.MountVolume` object's `MountFlags` fields. For an example of using it in kubernetes, see the persistence volume manifest in [Encryption in Transit Example](../examples/kubernetes/encryption_in_transit/specs/pv.yaml)
+
+For new releases starting v0.4.0, encryption in transit is always enabled.
 
 **Note** Kubernetes version 1.13+ is required if you are using this feature in Kubernetes.
 
@@ -32,12 +34,12 @@ To enable encryption in transit, `tls` needs to be set in the `NodePublishVolume
 The following sections are Kubernetes specific. If you are a Kubernetes user, use this for driver features, installation steps and examples.
 
 ### Kubernetes Version Compability Matrix
-| AWS EFS CSI Driver \ Kubernetes Version| maturity | v1.11 | v1.12 | v1.13 | v1.14 | v1.15 |
-|----------------------------------------|----------|-------|-------|-------|-------|-------|
-| master branch                          | beta     | no    | no    | no    | yes   | yes   |
-| v0.3.0                                 | beta     | no    | no    | no    | yes   | yes   |
-| v0.2.0                                 | beta     | no    | no    | no    | yes   | yes   |
-| v0.1.0                                 | alpha    | yes   | yes   | yes   | no    | no    |
+| AWS EFS CSI Driver \ Kubernetes Version (both master and worker node)| maturity | v1.11 | v1.12 | v1.13 | v1.14 | v1.15 |
+|----------------------------------------------------------------------|----------|-------|-------|-------|-------|-------|
+| master branch                                                        | beta     | no    | no    | no    | yes   | yes   |
+| v0.3.0                                                               | beta     | no    | no    | no    | yes   | yes   |
+| v0.2.0                                                               | beta     | no    | no    | no    | yes   | yes   |
+| v0.1.0                                                               | alpha    | yes   | yes   | yes   | no    | no    |
 
 ### Container Images
 |EFS CSI Driver Version     | Image                               |
@@ -49,7 +51,7 @@ The following sections are Kubernetes specific. If you are a Kubernetes user, us
 
 ### Features
 * Static provisioning - EFS filesystem needs to be created manually first, then it could be mounted inside container as a persistent volume (PV) using the driver.
-* Mount Options - Mount options can be specified in the persistence volume (PV) to define how the volume should be mounted. Aside from normal mount options, you can also specify `tls` as a mount option to enable encryption in transit of the EFS filesystem.
+* Mount Options - Mount options can be specified in the persistence volume (PV) to define how the volume should be mounted. Aside from normal mount options, for EFS CSI Driver v0.3.0 or earlier releases, you can also specify `tls` as a mount option to enable encryption in transit of the EFS filesystem (this feature is always enabled starting release 0.4.0).
 
 **Notes**:
 * Since EFS is an elastic filesystem it doesn't really enforce any filesystem capacity. The actual storage capacity value in persistence volume and persistence volume claim is not used when creating the filesystem. However, since the storage capacity is a required field by Kubernetes, you must specify the value and you can use any valid value for the capacity.
--- a/examples/kubernetes/access_points/README.md
+++ b/examples/kubernetes/access_points/README.md
@@ -2,7 +2,7 @@
 Like [volume path mounts](../volume_path), mounting [EFS access points](https://docs.aws.amazon.com/efs/latest/ug/efs-access-points.html) allows you to expose separate data stores with independent ownership and permissions from a single EFS volume.
 In this case, the separation is managed on the EFS side rather than the kubernetes side.
 
-**Note**: Because access point mounts require TLS, this is not supported in driver versions at or before `0.3`.
+**Note**: Because access point mounts require TLS, this is not supported in driver versions at or before `0.3.0`.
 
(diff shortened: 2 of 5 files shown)


docs/README.md
```diff
@@ -24,20 +24,22 @@ The following CSI interfaces are implemented:
 ### Encryption In Transit
 One of the advantages of using EFS is that it provides [encryption in transit](https://aws.amazon.com/blogs/aws/new-encryption-of-data-in-transit-for-amazon-efs/) support using TLS. Using encryption in transit, data will be encrypted during its transition over the network to the EFS service. This provides an extra layer of defence-in-depth for applications that requires strict security compliance.
 
-To enable encryption in transit, `tls` needs to be set in the `NodePublishVolumeRequest.VolumeCapability.MountVolume` object's `MountFlags` fields. For an example of using it in kubernetes, see the persistence volume manifest in [Encryption in Transit Example](../examples/kubernetes/encryption_in_transit/specs/pv.yaml)
+For EFS CSI Driver EFS CSI Driver v0.3.0 or earlier releases, this feature is optional. To enable encryption in transit, `tls` needs to be set in the `NodePublishVolumeRequest.VolumeCapability.MountVolume` object's `MountFlags` fields. For an example of using it in kubernetes, see the persistence volume manifest in [Encryption in Transit Example](../examples/kubernetes/encryption_in_transit/specs/pv.yaml)
```
wongma7, line 27: s/persistence/persistent

docs/README.md
```diff
@@ -24,20 +24,22 @@ The following CSI interfaces are implemented:
...
+
+For new releases starting v0.4.0, encryption in transit is always enabled.
```
wongma7, line 29: are we sure it will be 0.4.0 and not 1.0.0 ? :p

docs/README.md
```diff
@@ -24,20 +24,22 @@ The following CSI interfaces are implemented:
...
 The following sections are Kubernetes specific. If you are a Kubernetes user, use this for driver features, installation steps and examples.
 
 ### Kubernetes Version Compability Matrix
-| AWS EFS CSI Driver \ Kubernetes Version| maturity | v1.11 | v1.12 | v1.13 | v1.14 | v1.15 |
-|----------------------------------------|----------|-------|-------|-------|-------|-------|
-| master branch                          | beta     | no    | no    | no    | yes   | yes   |
-| v0.3.0                                 | beta     | no    | no    | no    | yes   | yes   |
-| v0.2.0                                 | beta     | no    | no    | no    | yes   | yes   |
-| v0.1.0                                 | alpha    | yes   | yes   | yes   | no    | no    |
+| AWS EFS CSI Driver \ Kubernetes Version (both master and worker node)| maturity | v1.11 | v1.12 | v1.13 | v1.14 | v1.15 |
```
wongma7, line 37: thanks. we can also just say `(worker node)` which implies the master is at least the same version because our driver does not care about master (yet). Not sure what is least confusing.

examples/kubernetes/access_points/README.md
```diff
@@ -2,7 +2,7 @@
 Like [volume path mounts](../volume_path), mounting [EFS access points](https://docs.aws.amazon.com/efs/latest/ug/efs-access-points.html) allows you to expose separate data stores with independent ownership and permissions from a single EFS volume.
 In this case, the separation is managed on the EFS side rather than the kubernetes side.
 
-**Note**: Because access point mounts require TLS, this is not supported in driver versions at or before `0.3`.
+**Note**: Because access point mounts require TLS, this is not supported in driver versions at or before `0.3.0`.
```
wongma7, line 5: this seems wrong? Isn't it possible to use both tls and accesspoint mount options even on 0.3.0? If you have time to verify this statement would appreciate it ...

leakingtapan:
What if some user doesn’t want tls? Enforcing the option doesn’t seem a good feature to me. And there are many problems with tls. I also don’t think it ready to be turned on.

/hold

wongma7:
Of course the main argument is security, that "secure always" is better than "secure by default."

My suspicion (I'm stil trying to reproduce them :( ) on issues where mount hangs like #141 (#112 didn't have tls enabled, not sure about #117) is that tls is not the cause, although it's an exacerbating factor because now instead of just x mount processes hanging you have x mount + x stunnel processes.

Personally, my thought is that if users don't intend to enable tls then there is no reason to use this driver over NFS, because if you don't enable tls then mount.efs *is* mount.nfs: https://github.com/aws/efs-utils/blob/master/src/mount_efs/__init__.py#L1516.

docs/README.md
mikestef9: Should be 1.0.0

wochanda:
> What if some user doesn’t want tls? 
> 

Why wouldn't they? We've implemented TLS in a way that does not sacrifice functinality or performance. Agree with @wongma7 on root cause for those bugs - TLS isn't conclusively the cause. Further, having a single option for client connectivity allows us to focus our hardening and testing on this path, making it more resilient for customers.

jqmichael:
> What if some user doesn’t want tls? Enforcing the option doesn’t seem a good feature to me. And there are many problems with tls. Like #141 #117 #112. I don’t think it’s ready even
> 
> /hold

Yeah, we (EKS team) have been work closely with EFS team to ensure turning on TLS is performant and available solution. EFS team has much more experience with supporting customers who enabled TLS in the past. We will make sure there's no known issue before enabling this on EKS. 

I think we are on the same page if the TLS solution (specifically the way EKS CSI Driver and EFS Mount Helper support TLS) is safe for customer, we should enable it for everyone for the benefit of security. Major websites around the world enforced TLS for over a decade.  

We will look more closely on the bugs linked above to see if there's any correlation with enabling TLS(as @wongma7 mentioned #112 wasn't turning on TLS) before seeking for merge.

leakingtapan:
> Why wouldn't they? We've implemented TLS in a way that does not sacrifice functinality or performance.

Is there any data for what's the performance difference with/without tls? And what's the limitation and trade offs with and without?

> My suspicion (I'm stil trying to reproduce them :( ) on issues where mount hangs like #141 (#112 didn't have tls enabled, not sure about #117) is that tls is not the cause, although it's an exacerbating factor because now instead of just x mount processes hanging you have x mount + x stunnel processes.

Let's get the related bug fixed first to make sure tls feature is reliable enough

> Of course the main argument is security, that "secure always" is better than "secure by default."

As of now, there are still bugs, I would prefer the bug fixed before turning on tls. We can set it to be default on if required afterward. If you really want to ensure `tls`, I think this should be enforce at efs mount helper layer, as the driver is simply surfacing what's provided by efs mount helper.

stefansedich:
I would be 100% against this unless all the TLS issues we experienced in the past are 110% fixed. I had to turn it off in the end as it was not stable enough for our environments and we experienced frequent lockups or various forms.

wongma7:
Thx, @stefansedich your feedback is much appreciated. There is at least one bug on the efs mount helper side https://github.com/aws/efs-utils/issues/69 that I would like to see fixed first and *could* explain mounts being unresponsive, could you take a look and see if it sounds familiar? It should only occur when the driver container is restarted (like if you update the image), so I'm not sure it could cover all the cases of instability you've seen.

Plus there is one bug on our (driver) side that needs to be fixed first, something to do with how a newer version of the efs mount helper writes private keys that need to be persisted (but this one does not affect any released versions of the driver).

stefansedich:
> Thx, @stefansedich your feedback is much appreciated. There is at least one bug on the efs mount helper side [aws/efs-utils#69](https://github.com/aws/efs-utils/issues/69) that I would like to see fixed first and _could_ explain mounts being unresponsive, could you take a look and see if it sounds familiar? It should only occur when the driver container is restarted (like if you update the image), so I'm not sure it could cover all the cases of instability you've seen.
> 
> Plus there is one bug on our (driver) side that needs to be fixed first, something to do with how a newer version of the efs mount helper writes private keys that need to be persisted (but this one does not affect any released versions of the driver).

That error looks related, we would see that randomly the mount would hang, our pods health check would fail at some point but it would get stuck in "terminating" forever as kublet had begun the unmount but would just hang forever.

It has been a while, I turned off TLS as we experienced enough production outages due to this issue, would be happy to test any fixes but they were sparodic and hard to replicate, I would think we were out of the woods after weeks of stability only to have the issue rear it's head again.

jqmichael:
FYI, given all the feedback from community, EKS will not enforce encryption in transit for the moment. So I'm closing this PR. 

The alternative we are proposing is to enable encryption in transit by default (because it significantly enhances security for traffic going to EFS), but allow users to opt-out with a flag.

https://github.com/kubernetes-sigs/aws-efs-csi-driver/pull/156 

Feel free to share your thoughts and feedback on the PR linked above. 

Thanks

Issue #167, opened 2020-05-20
Unable to mount two access points on same EFS volume


2uasimojo:
/kind bug

This is just like #100, but for access points.

**What happened?**
Pod fails to come up, eventually timing out waiting for mounts. Logs show only one mount was attempted.

**What you expected to happen?**
All file systems mounted; pod comes up.

**How to reproduce it (as minimally and precisely as possible)?**
- Create two PVs using [access points](https://github.com/kubernetes-sigs/aws-efs-csi-driver/blob/f8e964198d1c90c5343dc1a1cefa6d1d10e3bffa/examples/kubernetes/access_points/README.md). Crucially, use access points on the *same* `[FileSystemId]`.
- Create claims for the above PVs
- Create a pod using *both* claims. It will fail to start, showing symptoms identical to those described in #100.

**Anything else we need to know?**:
- The document referred to above states that this setup is supported. (That's my bad.) It needs to be fixed until/unless this bug is resolved. I'll put up a PR for that in the next few minutes.
- If you use access points on *different* file systems, it works fine.
- The solution (#102) for #100 was to glom the subdirectory into the `volumeHandle`, separated by a colon. It is not out of the question that a similar fix could be implemented here. For example, support any of:
  - `volumeHandle: fs-XXX`
  - `volumeHandle: fs-XXX:/path`
  - `volumeHandle: fs-XXX:fsap-YYY` <== new
  The access point would be fed into the mount options by the driver, rather than having to specify it explicitly under `mountOptions` in the PV spec. (Note that I don't think `fs-XXX:fsap-YYY:/path` makes sense because the access point is already subpathed, but it could be looked at.)

**Environment**
- Kubernetes version (use `kubectl version`):
```
Client Version: openshift-clients-4.3-31-gc25fb9c7
Server Version: 4.3.19
Kubernetes Version: v1.16.2
```
- Driver version:
latest (as of 20200520)

2uasimojo:
> * The document referred to above states that this setup is supported. (That's my bad.) It needs to be fixed until/unless this bug is resolved. I'll put up a PR for that in the next few minutes.

#168 :arrow_up:

leakingtapan:
Yes, the access point ID has to be part of `volumeHandle` to make kubelet aware of there is a different volume to be mounted instead of the mounting the same volume twice (which will be ignored by kubelet).

We need to define a proper `volumeHandle` semantic for both use cases:
* just the filesystem ID
[...]

PR #168, opened 2020-05-20
Doc: Access Points: Require separate file systems

2uasimojo:
Per [issue #167](https://github.com/kubernetes-sigs/aws-efs-csi-driver/issues/167), it turns out that you *can't* use separate access points to the same EFS volume from the same pod. Correct the documentation pending a fix.

Docs only.

Docs are wrong until #167 is resolved.

**What testing is done?** 
Rendered doc. Looks peachy.


2uasimojo:
Reverted in #185, which fixes #167.

Issue #151, opened 2020-04-23
Bump test cluster version


leakingtapan:
**Is your feature request related to a problem? Please describe.**
Bump test cluster version

PR #174, opened 2020-06-01
Bump k8s dependencies to 1.17.6

wongma7:

**What is this PR about? / Why do we need it?** Routine bump of dependencies. Only thing of note is that k8s.io/utils/mount is now out of k8s.io/kubernetes

**What testing is done?**


wongma7:
need to also bump test cluster version (https://github.com/kubernetes-sigs/aws-efs-csi-driver/issues/151) as by bumping the deps to 1.17 we picked up a test that only makes sense when run against 1.17 clusters (volumeMode should not mount / map unused volumes in a pod)

Issue #178, opened 2020-06-08
All existing mounts hang after the EKS Driver container is replaced on a given node


jqmichael:
/kind bug

**What happened?**
After the EKS driver container is replaced (i.e. by terminating the driver process or upgrading the driver to a new image), all existing mounts on that node hang indefinitely. 

**What you expected to happen?**
Existing mounts should still be functioning after a new EKS driver container is launched to the same node. 

**How to reproduce it (as minimally and precisely as possible)?**
Deploy EFS driver from master and create a pod using the encryption_in_transit example. So far so good. 

exec into the EFS CSI driver container and kill the driver process. 
```
kubectl exec $(kubectl get po -l app=efs-csi-node -n kube-system -o jsonpath='{.items[0].metadata.name}') -n kube-system -it efs-plugin -- sh

kill 1
```

Reading file from the EFS mount hangs indefinitely. 
```
kubectl exec -ti efs-app -- tail -f /data/out.txt
```

**Anything else we need to know?**:
Watch dog log:
```
2020-06-03 20:13:23,191 - INFO - Starting TLS tunnel: "/usr/bin/stunnel /var/run/efs/stunnel-config.fs-b6654c1c.var.li
b.kubelet.pods.d20ab7da-b516-4d8c-a397-3fd3f6d8616c.volumes.kubernetes.io~csi.efs-pv1.mount.20309"
2020-06-03 20:13:23,222 - INFO - Started TLS tunnel, pid: 1996

2020-06-03 20:13:25,288 - WARNING - Child TLS tunnel process 1996 has exited, returncode=1
```
Manually invoking the stunnel command gave me the following error:
```
/etc/amazon/efs/privateKey.pem: No such file or directory (2)
Service [efs]: Failed to initialize SSL context
```

**Environment**
- Kubernetes version (use `kubectl version`): 1.16
- Driver version: master

wongma7:
mentioned this bug here "Plus there is one bug on our (driver) side that needs to be fixed first, something to do with how a newer version of the efs mount helper writes private keys that need to be persisted (but this one does not affect any released versions of the driver)." https://github.com/kubernetes-sigs/aws-efs-csi-driver/pull/166#issuecomment-640932009 (just linking for reference)

jqmichael:
[...]

PR #179, opened 2020-06-09
Persisted private key genereated by efs-util to host

jqmichael:
Fix https://github.com/kubernetes-sigs/aws-efs-csi-driver/issues/178

This PR persists the private key generated by efs-utils at https://github.com/aws/efs-utils/blob/7b1f2d0f90afc5a6bfdf5522d5231eb8a91c184d/src/mount_efs/__init__.py#L1046.

An side-effect of making `/etc/amazon/efs` folder  a hostpath is it cleans up the content written to that folder when installing efs-utils, including `efs-utils.conf` and `efs-utils-crt` (https://github.com/aws/efs-utils/blob/7b1f2d0f90afc5a6bfdf5522d5231eb8a91c184d/dist/efs-utils.conf and https://github.com/aws/efs-utils/blob/7b1f2d0f90afc5a6bfdf5522d5231eb8a91c184d/dist/efs-utils.crt), because hostpath added a new layer on top of the image that's built from Dockerfile. 

The workaround is 
1) Having EFS driver create the `efs-utils.conf` directly, which is the right thing to do anyway. 
2) Pointing stunnel to use a pre-existing root CA from the image, `/etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem`, which is where Golang looks up Root CA in CentOS. 

**What testing is done?** 
Killed the driver process and updates to EFS mount request still went through right after a new container is launched. 

```
$kubectl exec -ti efs-app -- tail -f /data/out.txt

Sat Jun 6 01:02:50 UTC 2020
Sat Jun 6 01:02:55 UTC 2020 # killed the driver process which terminated the old container
Sat Jun 6 01:03:06 UTC 2020
Sat Jun 6 01:03:11 UTC 2020
Sat Jun 6 01:03:16 UTC 2020
```

`kubectl exec` into the driver container. 

```
sh-4.2# aws-efs-csi-driver --version
{
  "driverVersion": "v0.4.0-dirty",
  "gitCommit": "5741533ed37c0bc44c3a639de50acb2fdf9f78a2",
  "buildDate": "2020-06-10T21:52:08Z",
  "efsClientSource": "k8s",
  "goVersion": "go1.13.4",
  "compiler": "gc",
  "platform": "linux/amd64"
}

sh-4.2# cat /etc/amazon/efs/efs-utils.conf 
[DEFAULT]
logging_level = INFO
... 
tls_cert_renewal_interval_min = 60

[client-info] 
source=k8s
```

### DIFF
--- a/pkg/driver/efs_watch_dog.go
+++ b/pkg/driver/efs_watch_dog.go
@@ -15,16 +15,72 @@ package driver
 
 import (
 	"fmt"
+	"os"
 	"os/exec"
 	"sync"
+	"text/template"
 
 	"k8s.io/klog"
 )
 
+// https://github.com/aws/efs-utils/blob/master/dist/efs-utils.conf
+const efsUtilsConfigTemplate = `
+[DEFAULT]
+logging_level = INFO
+logging_max_bytes = 1048576
+logging_file_count = 10
+# mode for /var/run/efs and subdirectories in octal
+state_file_dir_mode = 750
+
+[mount]
+dns_name_format = {fs_id}.efs.{region}.{dns_name_suffix}
+dns_name_suffix = amazonaws.com
+#The region of the file system when mounting from on-premises or cross region.
+#region = us-east-1
+stunnel_debug_enabled = true
+#Uncomment the below option to save all stunnel logs for a file system to the same file
+stunnel_logs_file = /var/log/amazon/efs/{fs_id}.stunnel.log
+# RootCA on AmazonLinux2/CentOS. https://golang.org/src/crypto/x509/root_linux.go
+stunnel_cafile = /etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem
+
+# Validate the certificate hostname on mount. This option is not supported by certain stunnel versions.
+stunnel_check_cert_hostname = true
+
+# Use OCSP to check certificate validity. This option is not supported by certain stunnel versions.
+stunnel_check_cert_validity = false
+
+# Define the port range that the TLS tunnel will choose from
+port_range_lower_bound = 20049
+port_range_upper_bound = 20449
+
+[mount.cn-north-1]
+dns_name_suffix = amazonaws.com.cn
+
+[mount.cn-northwest-1]
+dns_name_suffix = amazonaws.com.cn
+
+[mount.us-iso-east-1]
+dns_name_suffix = c2s.ic.gov
+
+[mount.us-isob-east-1]
+dns_name_suffix = sc2s.sgov.gov
+
+[mount-watchdog]
+enabled = true
+poll_interval_sec = 1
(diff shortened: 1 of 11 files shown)


pkg/driver/efs_watch_dog.go
```diff
@@ -39,22 +95,43 @@ type execWatchdog struct {
...
+	}
+
 	go w.runLoop(w.stopCh)
+
+	return nil
+}
+
+func (w *execWatchdog) updateConfig() error {
+	efsUtilsConfig := template.Must(template.New("efs-utils-config").Parse(efsUtilsConfigTemplate))
+	f, err := os.Create(w.efsUtilsCfgPath)
```
leakingtapan, line 127: do you need to close the file after write finishes?

pkg/driver/efs_watch_dog.go
```diff
@@ -39,22 +95,43 @@ type execWatchdog struct {
...
+	if err != nil {
+		return fmt.Errorf("cannot create config file %s for efs-utils. Error: %v", w.efsUtilsCfgPath, err)
+	}
+	if err = efsUtilsConfig.Execute(f, efsUtilsConfig); err != nil {
```
leakingtapan, line 131: Why pass `efsUtilsConfig` in as the template object ? Does the template have any substitution variable?

pkg/driver/efs_watch_dog.go
jqmichael: Nice catch! We should definitely close file after we are done with it to avoid having too many open files. 

Fixed.

pkg/driver/efs_watch_dog.go
jqmichael: Not in this PR. In a upcoming PR, I'll customize the client info section. Also in the future we may set the region in the config file as well.

pkg/driver/efs_watch_dog.go
```diff
@@ -15,16 +15,72 @@ package driver
...
+[mount]
+dns_name_format = {fs_id}.efs.{region}.{dns_name_suffix}
+dns_name_suffix = amazonaws.com
+#The region of the file system when mounting from on-premises or cross region.
+#region = us-east-1
+stunnel_debug_enabled = true
+#Uncomment the below option to save all stunnel logs for a file system to the same file
+stunnel_logs_file = /var/log/amazon/efs/{fs_id}.stunnel.log
+# RootCA on AmazonLinux2/CentOS. https://golang.org/src/crypto/x509/root_linux.go
+stunnel_cafile = /etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem
```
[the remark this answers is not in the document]
wongma7: for reference the CA created by efs-utils is https://github.com/aws/efs-utils/blob/master/dist/efs-utils.crt

PR #186, opened 2020-06-17
DRY TestNodePublishVolume

2uasimojo:

Neither, just a refactor for maintainability.

Refactor TestNodePublishVolume to be truly structure-driven and less
repetitive by collapsing the test logic into the `Run` call and
changing the driver struct to contain only the inputs, outputs, and
expected results.

This greatly reduces the amount of copy/pasted test code, thereby
reducing the overall size of the file. It also (at least IMHO) makes the
test cases more scrutable because, once you understand the `Run` logic,
you can quickly understand the test case by looking at the struct
values.

A similar thing could be done (later) to TestNodeUnpublishVolume.

**What testing is done?** 

N/A

Signed-off-by: Eric Fried <efried@redhat.com>

### DIFF
--- a/pkg/driver/node_test.go
+++ b/pkg/driver/node_test.go
@@ -44,392 +44,214 @@ func TestNodePublishVolume(t *testing.T) {
 	)
 
 	testCases := []struct {
-		name     string
-		testFunc func(t *testing.T)
+		name          string
+		req           *csi.NodePublishVolumeRequest
+		expectMakeDir bool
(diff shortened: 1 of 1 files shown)


wongma7:
/ok-to-test
/lgtm
/approve

Thanks. I tried this last year but was not smart enough to make `req` a parameter.

wongma7:
https://github.com/kubernetes-sigs/aws-efs-csi-driver/pull/49/files

PR #189, opened 2020-06-19
Fix TestNodeUnpublishVolume

2uasimojo:
**Is this a bug fix or adding new feature?**

Exposes a bug

**What is this PR about? / Why do we need it?**

Turns out a couple of these tests were broken because the finisher
wasn't being called on the mock controller. When the finisher is added:

```
[efried@efried aws-efs-csi-driver]$ go test ./pkg/driver -run TestNodeUnpublishVolume
--- FAIL: TestNodeUnpublishVolume (0.00s)
    --- FAIL: TestNodeUnpublishVolume/success:_unpublish_with_already_unmounted_target (0.00s)
        node_test.go:478: missing call(s) to *mocks.MockMounter.Unmount(is equal to /target/path) /home/efried/go/src/github.com/kubernetes-sigs/aws-efs-csi-driver/pkg/driver/node_test.go:472
        node_test.go:478: aborting test due to missing call(s)
    --- FAIL: TestNodeUnpublishVolume/fail:_mounter_failed_to_GetDeviceName (0.00s)
        node_test.go:557: missing call(s) to *mocks.MockMounter.Unmount(is equal to /target/path) /home/efried/go/src/github.com/kubernetes-sigs/aws-efs-csi-driver/pkg/driver/node_test.go:551
        node_test.go:557: aborting test due to missing call(s)
FAIL
coverage: 7.9% of statements
FAIL	github.com/kubernetes-sigs/aws-efs-csi-driver/pkg/driver	0.005s
FAIL
```

Fix.

**What testing is done?** 

This *is* testing :P

PR #190, opened 2020-06-19
Refactor and extend node tests

2uasimojo:

Neither; cleaning up and extending tests.

Make tests more maintainable. Make tests more thorough.

**What testing is done?** 

Extended testing of Node{P|Unp}upblishVolume to
validate the error codes and message strings. Not that it's super
important that these strings are spelled exactly right; more to make
sure we're erroring from the exact places we are expecting for each test
case.

### DIFF
--- a/pkg/driver/node.go
+++ b/pkg/driver/node.go
@@ -73,11 +73,11 @@ func (d *Driver) NodePublishVolume(ctx context.Context, req *csi.NodePublishVolu
 		case "path":
 			klog.Warning("Use of path under volumeAttributes is depracated. This field will be removed in future release")
 			if !filepath.IsAbs(v) {
-				return nil, status.Error(codes.InvalidArgument, fmt.Sprintf("Volume context property %q must be an absolute path", k))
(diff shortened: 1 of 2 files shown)


2uasimojo:
Between this and #189, one will need a rebase when the other merges.

wongma7:
/hold
to give you time to rebase this on #189

2uasimojo:
> /hold
> to give you time to rebase this on #189

Done. Also fixed some test fails I somehow missed, and de-linted. Thanks @wongma7.

2uasimojo:
Wow. I think the coverage is "reduced" because I trimmed one line out of node.go, thereby reducing both the numerator and denominator in the equation by one. That seems like something the coverage tool ought to take into account.

For now, I'll just blow that back into two lines :man_shrugging:

2uasimojo:
Oh, is the coverage job nonblocking?

wongma7:
Yeah it's non blocking, I guess we don't have to worry about it.

/lgtm
/hold cancel

Issue #192, opened 2020-06-22
MountVolume.SetUp failed for volume "efs-pv"


nomopo45:
/kind bug

Hello,

I followed this documentation : https://docs.aws.amazon.com/eks/latest/userguide/efs-csi.html, as written in the userguide i'm trying to use the Multiple Pods Read Write Many

I have an EKS cluster, and i have an issue with my pod creation, apparently he is unable to mount the EFS Volume.
 
here is some log that i found :

```
MountVolume.SetUp failed for volume "efs-pv" : kubernetes.io/csi: mounter.SetupAt failed: rpc error: code = Internal desc = Could not mount "fs-xxxxxxx:/" at "/var/lib/kubelet/pods/eec7379e-0d59-440d-87c3-050c59c27b45/volumes/kubernetes.io~csi/efs-pv/mount": mount failed: exit status 1 Mounting command: mount Mounting arguments: -t efs fs-xxxxxxx:/ /var/lib/kubelet/pods/eec7379e-0d59-440d-87c3-050c59c27b45/volumes/kubernetes.io~csi/efs-pv/mount Output: Traceback (most recent call last): File "/sbin/mount.efs", line 1375, in <module> main() File "/sbin/mount.efs", line 1355, in main bootstrap_logging(config) File "/sbin/mount.efs", line 1031, in bootstrap_logging raw_level = config.get(CONFIG_SECTION, 'logging_level') File "/lib64/python2.7/ConfigParser.py", line 607, in get raise NoSectionError(section) ConfigParser.NoSectionError: No section: 'mount'
```

and

```
Unable to attach or mount volumes: unmounted volumes=[persistent-storage], unattached volumes=[persistent-storage default-token-xvnhr]: timed out waiting for the condition
```

I checked many things, my SG are fully open in and out, my pv.yaml, pod1.yaml, classtorage.yml and claim.yaml are exactly the same as here https://github.com/kubernetes-sigs/aws-efs-csi-driver/tree/master/examples/kubernetes/multiple_pods

*Environment**
- Client Version: version.Info{Major:"1", Minor:"16+", GitVersion:"v1.16.6-beta.0", GitCommit:"e7f962ba86f4ce7033828210ca3556393c377bcc", GitTreeState:"clean", BuildDate:"2020-01-15T08:26:26Z", GoVersion:"go1.13.5", Compiler:"gc", Platform:"darwin/amd64"}
[...]

Issue #191, opened 2020-06-20
Using the latest yaml install file + v0.3.0 driver no longer works


jqmichael:
/kind bug

**What happened?**
issue is our command to install efs driver is downloading an image which is not working. 
kubectl apply -k "github.com/kubernetes-sigs/aws-efs-csi-driver/deploy/kubernetes/overlays/stable/?ref=master"

MountDevice failed to create newCsiDriverClient: driver name efs.csi.aws.com not found in the list of registered CSI drivers

**What you expected to happen?**

**How to reproduce it (as minimally and precisely as possible)?**

**Anything else we need to know?**:

**Environment**
- Kubernetes version (use `kubectl version`):
- Driver version:

jqmichael:
We added a hostpath to /etc/amazon/efs/ which override the efs-utils.conf file. The driver from master branch will install the config file itself, but since you are using a stable version of the driver, that didn't happen.

https://github.com/kubernetes-sigs/aws-efs-csi-driver/blob/master/deploy/kubernetes/base/node.yaml#L102

I think we should start branching the yaml installment of the driver.

wongma7:
We can maintain overlays for each version in master. And let stable always be a copy of the latest stable release. We should not change stable until after a new release when we do the post-release stuff, update the README etc.: https://github.com/kubernetes-sigs/aws-efs-csi-driver/pull/140/

wongma7:
Let's revert ad02e03 for now. Basically we should never touch stable overlay except in post-release stage where we will copy changes from dev overlay to stable overlay.

omriarieli:
Hi, FYI i am getting validation errors when installing with helm3 on eks kubernetes 1.15

null_resource.efs_csi_driver (local-exec): Error: rendered manifests contain a resource that already exists. Unable to continue with install: CSIDriver "efs.csi.aws.com" in namespace "" exists and cannot be imported into the current release: invalid ownership metadata; label validation error: missing key "app.kubernetes.io/managed-by": must be set to "Helm"; annotation validation error: missing key "meta.helm.sh/release-name": must be set to "aws-efs-csi-driver"; annotation validation error: missing key "meta.helm.sh/release-namespace": must be set to "default"

the part about CSIDriver existing is irrelevant i had to manually delete it since it was installed before with kubectl.
also after the helm chart is installed it reports it is installed in default namespace using helm ls, even though it is installed in kube-system namespace

wongma7:
[...]

PR #193, opened 2020-06-23
fix (helm): honor imagePullPolicy setting

ossareh:

bug fix, see #192 

_what is this PR?_

This change does the following:

 * (bug fix) `image.pullPolicy` is threaded to the `efs-plugin` container
 * (feature) added `pullPolicy` to the side car containers and threaded them through to their containers in the DaemonSet
 * (refactor) standardized on `print "%s:%s"` for generating image:tag in the DaemonSet template.

_why do we need it?_

The community of users of this chart have started to recommend the use of the `:latest` docker image. See #192. However if you specify `image.tag: "latest"` when using helm you are locked in at the version of `:latest` at the time of installation. Kubernetes provides the `imagePullPolicy` setting to work around this - but the helm chart prior to this change does not honor that value. This change will allow users of the helm chart to use `:latest` and get the desired outcome. It also adds the same `pullPolicy` options to the side car containers.

**What testing is done?** 

I've run this successfully on our production cluster. The salient work here is merely updating the helm chart to honor values in `helm/values.yaml` so the main test is that changing `pullPolicy` reflects the correct value in kubernetes.

### DIFF
--- a/deploy/kubernetes/base/node.yaml
+++ b/deploy/kubernetes/base/node.yaml
@@ -25,7 +25,8 @@ spec:
         - name: efs-plugin
           securityContext:
             privileged: true
-          image: amazon/aws-efs-csi-driver:latest
+          image: amazon/aws-efs-csi-driver:v0.3.0
+          imagePullPolicy: IfNotPresent
           args:
             - --endpoint=$(CSI_ENDPOINT)
             - --logtostderr
@@ -57,6 +58,7 @@ spec:
             failureThreshold: 5
         - name: csi-driver-registrar
           image: quay.io/k8scsi/csi-node-driver-registrar:v1.3.0
+          imagePullPolicy: IfNotPresent
           args:
             - --csi-address=$(ADDRESS)
             - --kubelet-registration-path=$(DRIVER_REG_SOCK_PATH)
@@ -77,6 +79,7 @@ spec:
               mountPath: /registration
         - name: liveness-probe
           image: quay.io/k8scsi/livenessprobe:v2.0.0
+          imagePullPolicy: IfNotPresent
           args:
             - --csi-address=/csi/csi.sock
             - --health-port=9809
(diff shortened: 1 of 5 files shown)


ossareh:
I've signed the CLA @k8s-ci-robot

ossareh:
@leakingtapan thanks for the go ahead on tests.

It seems the tests failed for reasons discussed in #191; the error line can be found here: https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_aws-efs-csi-driver/193/pull-aws-efs-csi-driver-e2e/1275819244449173504#1:build-log.txt%3A2567

I'm not familiar enough with how [testsuites](https://github.com/kubernetes/kubernetes/blob/master/test/e2e/storage/testsuites/subpath.go) work in the kubernetes project so I'm not clear on how to proceed; any direction would be appreciated.

wongma7:
Hmm, what about the stable overlay then, don't we also want that to have policy IfNotPresent? currently it's not set so it will default to Always. Makes sense to me that only the dev overlay should have Always as added by your other PR https://github.com/kubernetes-sigs/aws-efs-csi-driver/pull/195

ossareh:
Weird - this whole arc of changes was kicked off because I was getting `IfNotPresent` semantics while using the `:latest` tag for this CSI driver.

But y'all are right as far as the docs go: https://kubernetes.io/docs/concepts/containers/images/#updating-images - this should be unnecessary. 

I'll go and make the changes, and then attempt to establish what was actually happening.

ossareh:
@leakingtapan I've made the following change to `deploy/kubernetes..`:

 * default node.yaml to stable semantics
 * `overlay/stable` doesn't edit resources
 
So now; the helm chart semantics and the kustomize semantics are the same; where the user can opt into "dev-like" semantics by either specifying the dev overlay or by providing the correct values to helm.

I'm happy to move this new commit to another branch to honor "branches should change one thing" if you want; but since the conversation was here it seemed reasonable to make that change here since the merge of #195 has happened.

ossareh:
It's obvious to me now that this and #195 were not the correct solution to the issue; I'm on vacation right now - but when I'm back (next week) I'm going to draw up PR's that revert these commits.

PR #196, opened 2020-06-25
Revert "Persisted private key genereated by efs-util to host"

wongma7:
Reverts kubernetes-sigs/aws-efs-csi-driver#179

https://github.com/kubernetes-sigs/aws-efs-csi-driver/issues/191

We are still discussing a solution to maintaining *stability* of the *stable* overlay. In meantime, let's revert.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -27,6 +27,15 @@ RUN make aws-efs-csi-driver
 FROM amazonlinux:2.0.20200406.0
 RUN yum install util-linux-2.30.2-2.amzn2.0.4.x86_64 amazon-efs-utils-1.24-4.amzn2.noarch -y
 
+# Default client source is k8s which can be overriden with –build-arg when building the Docker image
+ARG client_source=k8s
+RUN echo "client_source:${client_source}"
+RUN printf "\n\
+\n\
+[client-info] \n\
+source=${client_source} \n\
+" >> /etc/amazon/efs/efs-utils.conf
+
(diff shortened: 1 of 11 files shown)


leakingtapan:
/ok-to-test

please make the final proposal public once you have any conclusions

wongma7:
Hmm coveralls says it's not required but still is blocking merge. @leakingtapan can you take a look? IMO coveralls shouldn't block merge or at least the threshold should be higher, like 1%.

2uasimojo:
:eyes: 

<$0.02>
IMO there's nothing wrong with pushing breaking changes to `latest`, especially now while major == 0. Just at any given time the `latest` image ought to work with the YAML from the tip of the master branch.

As for tagged versions, the same rule should apply: the image with $tag must work against the YAML from the repo at $tag. However, here we have the opportunity to make things a little nicer by providing specific overlays for each $tag, hardcoded into the repo and frozen at that level.

In any case, the process for upgrading (including from `latest` to `latest`, unfortunately) should always include swapping both the YAML and the image.

The real problem with `latest` moving is the infrequency of tagged releases (cf. #152). It would be really nice if releases could be tagged *much* more frequently -- basically any time something new/significant has happened and we've had a chance to vet a specific commit with those changes. If we did that, we would have far fewer consumers relying on using `latest` in the first place. And picking up new features would be the explicit, deliberate action of moving from one tag to another.
</$0.02>

jqmichael:
We should start branching the yaml files.

Issue #111, opened 2019-12-10
support for arm64 nodes


zhouziyang:
docker image and k8s manifest for arm64

leakingtapan:
High level changes could include:
* cross compile the driver for arm
* build container image for arm
* install efs mount helper for arm
* install nfs client for arm

@zhouziyang could you give some some use case about why do you need to run the driver on arm platform?

zhouziyang:
Hi @leakingtapan , I'm currently evaluating EKS Arm worker nodes (1.14); and efs csi seeming missing arm64 images. I created them myself ([docker hub] zhouziyang/aws-efs-csi-driver-arm64, zhouziyang/livenessprobe-arm64, zhouziyang/node-driver-registrar-arm64), and they worked as expected. 
It would be much more convenient if officially supported!
Thanks!

tsarna:
#184 is indeed the same issue. Sorry, I think I searched for aarch64 and not arm64 so I missed this.

> could you give some some use case about why do you need to run the driver on arm platform?

The arm64 based m6g family VMs are now generally available, and price/performance is reportedly much better than m5. I think many people would want to use them in unless they have a specific need to stay on x86_64.

leakingtapan:
I created a PR for arm64 and an arm64 image using similar command in #197. Since I don't have arm64 machine, mind testing using the image: chengpan/aws-efs-csi-driver:latest ?

leakingtapan:
@tsarna

paolofacchinetti:
Any update on a release for arm64 EC2 instances?

We want to try a c6g cluster, but the lack of support for arm64 means that we can't proceed in that direction.

tsarna:
Sorry this took so long to test, I had some other blockers come up.

chengpan/aws-efs-csi-driver:latest seems to work just fine for me on a t4g instance. Sad to see that #197 was closed

tsarna:
Thanks for working on this! It looks like the sticking point for #197 was the charts. Any chance the build portion could be committed and a separate issue split off for the charts portion?

Thanks!

Issue #184, opened 2020-06-16
Please provide an official aarch64 image for use on m6g, a1


tsarna:
**Is your feature request related to a problem?/Why is this needed**
The 64-bit ARM-based m6g family is generally available now and provides very attractive price/performance (and the older a1 has been available for a while). It should be possible to use these ARM-based instances just as easily as amd64, and to that end there should be official images on Docker hub for aarch64 as well.
/feature

**Describe the solution you'd like in detail**
Build the current version on ARM as well and offer it on Docker Hub alongside the existing amd64 images.

**Describe alternatives you've considered**
It should be possible to build my own image easily enough, and I may end up doing that in the interim, but it's an extra hassle, and I feel AWS should provide aarch64 support as a first-class feature given that they now have two GA amd64 instance type families on offer.

leakingtapan:
Similar request: https://github.com/kubernetes-sigs/aws-efs-csi-driver/issues/111

KorvinSzanto:
Is there a workaround today beyond building my own image?

lexagigabit:
Just a comment for whoever is here struggling with the same issue, I just saw that the github workflow for this project contains arm64 instructions, I used the master branch and it just worked perfectly
Take into account that in case you tried to use a link from AWS Docs with the 1.1 release and it failed, you should delete all the `efs-csi` related stuff running, including the pods, which are stuck in error state, because that is what I did not do and it did not work, then I cleaned it up and it worked. 
So check there is nothing efs-related anywhere, then just
`kubectl apply -k "github.com/kubernetes-sigs/aws-efs-csi-driver/deploy/kubernetes/overlays/stable/ecr/?ref=master"`
And it works

dbuos:
Does anyone know if the support for aarch64 already released and stable?

wongma7:
v1.3.2 is multiarch. The sidecar images (as installed in the daemonset/deployment by the helm chart/kustomize) should be multiarch as well.
```
~ $ docker manifest inspect amazon/aws-efs-csi-driver:v1.3.2
{
   "schemaVersion": 2,
   "mediaType": "application/vnd.docker.distribution.manifest.list.v2+json",
   "manifests": [
      {
         "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
         "size": 1582,
         "digest": "sha256:42fd397df352973e06c9cb8939cd031f205b42dc84590b8ded8eb6709fe6505b",
         "platform": {
[...]

PR #197, opened 2020-06-25
Add arm64 support

leakingtapan:
Fixes: #111 #184 

Add arm64 support

### DIFF
--- a/deploy/kubernetes/base/node.yaml
+++ b/deploy/kubernetes/base/node.yaml
@@ -16,7 +16,6 @@ spec:
     spec:
       nodeSelector:
         kubernetes.io/os: linux
-        kubernetes.io/arch: amd64
       hostNetwork: true
       priorityClassName: system-node-critical
       tolerations:
--- a/.github/workflows/container-image.yaml
+++ b/.github/workflows/container-image.yaml
@@ -8,9 +8,18 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v1
+    - name: Set up Docker Buildx
+      id: buildx
+      uses: crazy-max/ghaction-docker-buildx@v3
+      with:
+        buildx-version: latest
+        qemu-version: latest
     - name: Build container image
       run: |
-        docker build -t aws-efs-csi-driver .
+        docker buildx build \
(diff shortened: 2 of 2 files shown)


deploy/kubernetes/base/node.yaml
```diff
@@ -16,7 +16,6 @@ spec:
     spec:
       nodeSelector:
         kubernetes.io/os: linux
-        kubernetes.io/arch: amd64
```
jqmichael, line 19: Updating this file directly will break users who use the older driver, doesn't it?  

I have a proposal of starting branching install yaml files. Will update the link once I write it down.

leakingtapan: Why will it break?

IIUC your question. Applying new manifest on old driver is not expected. Each driver release comes with it's dedicated manifest and helm chart. 

User should get the manifest or chart using the version tag. eg. here is the manifest for v0.2 driver: https://raw.githubusercontent.com/kubernetes-sigs/aws-efs-csi-driver/release-0.2.0/deploy/kubernetes/manifest.yaml
And there is a v0.3.0 tag which contains to v0.3.0 driver manifest.

And the versioned chart is located at: https://github.com/kubernetes-sigs/aws-efs-csi-driver/releases/download/{version}/helm-chart.tgz

That being said, this change will only be part of v0.4.0 release, which user only suppose to apply when upgrading / creating v0.4.0 driver

jqmichael: The current recommended way to install stable driver version according to the README is 

> kubectl apply -k "github.com/kubernetes-sigs/aws-efs-csi-driver/deploy/kubernetes/overlays/stable/?ref=master"

And that will pick up this code with 0.3.0 driver image. 

And that's is also why we had bug like https://github.com/kubernetes-sigs/aws-efs-csi-driver/issues/191. 

Please take a look at a proposal to solve this problem by branching the yaml install file. 

https://github.com/kubernetes-sigs/aws-efs-csi-driver/issues/207

jqmichael:
@leakingtapan 

Do you still plan to merge this? 

BTW, did you get a chance to test the image on ARM nodes?

PR #205, opened 2020-06-30
Add encryptInTransit volume attribute

wongma7:

**What is this PR about? / Why do we need it?** We would like encryption of data in transit to be the default. Currently, encryption is turned on/off by adding/omitting the "tls" mount option to/from PV spec.mountOptions. With this PR the value of `encryptInTransit` will determine whether encryption in transit is on, i.e. whether the *final* mount.efs command contains the "tls" mount option or not.

Intended behavior:

|  | encryptInTransit is unset | encryptInTransit is true | encryptInTransit is false |
| ------------- | ------------- | ------------- | ------------- |
| "tls" is in mountOptions  | encryption + deprecation warning  | encryption + deprecation warning | error |
| "tls" isn't in mountOptions | encryption  | encryption | NO encryption |

**What testing is done?**  the unit tests I added pass. The e2e tests I added pass on EKS and will be run on kops by PR CI.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -4,7 +4,7 @@
 
 ## Amazon EFS CSI Driver
 
-The [Amazon Elastic File System](https://aws.amazon.com/efs/) Container Storage Interface (CSI) Driver implements the [CSI](https://github.com/container-storage-interface/spec/blob/master/spec.md) specification for container orchestrators to manage the lifecycle of Amazon EFS filesystems.
+The [Amazon Elastic File System](https://aws.amazon.com/efs/) Container Storage Interface (CSI) Driver implements the [CSI](https://github.com/container-storage-interface/spec/blob/master/spec.md) specification for container orchestrators to manage the lifecycle of Amazon EFS file systems.
 
 ### CSI Specification Compatibility Matrix
 | AWS EFS CSI Driver \ CSI Spec Version  | v0.3.0| v1.1.0 | v1.2.0 |
@@ -15,7 +15,7 @@ The [Amazon Elastic File System](https://aws.amazon.com/efs/) Container Storage 
 | v0.1.0                                 | yes   | no     | no     |
 
 ## Features
-Currently only static provisioning is supported. This means an AWS EFS filesystem needs to be created manually on AWS first. After that it can be mounted inside a container as a volume using the driver.
+Currently only static provisioning is supported. This means an AWS EFS file system needs to be created manually on AWS first. After that it can be mounted inside a container as a volume using the driver.
 
 The following CSI interfaces are implemented:
 * Node Service: NodePublishVolume, NodeUnpublishVolume, NodeGetCapabilities, NodeGetInfo, NodeGetId
@@ -24,7 +24,7 @@ The following CSI interfaces are implemented:
 ### Encryption In Transit
 One of the advantages of using EFS is that it provides [encryption in transit](https://aws.amazon.com/blogs/aws/new-encryption-of-data-in-transit-for-amazon-efs/) support using TLS. Using encryption in transit, data will be encrypted during its transition over the network to the EFS service. This provides an extra layer of defence-in-depth for applications that requires strict security compliance.
 
-To enable encryption in transit, `tls` needs to be set in the `NodePublishVolumeRequest.VolumeCapability.MountVolume` object's `MountFlags` fields. For an example of using it in kubernetes, see the persistence volume manifest in [Encryption in Transit Example](../examples/kubernetes/encryption_in_transit/specs/pv.yaml)
+Encryption in transit is enabled by default in the master branch version of the driver. To disable it and mount volumes using plain NFSv4, set `volumeAttributes` field `encryptInTransit` to `"false"` in your persistent volume manifest. For an example manifest, see [Encryption in Transit Example](../examples/kubernetes/encryption_in_transit/specs/pv.yaml).
 
 **Note** Kubernetes version 1.13+ is required if you are using this feature in Kubernetes.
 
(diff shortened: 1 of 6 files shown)


jqmichael:
@stefansedich @leakingtapan 

Please take a look.

examples/kubernetes/encryption_in_transit/specs/pv.yaml
```diff
@@ -10,8 +10,8 @@ spec:
     - ReadWriteOnce
   persistentVolumeReclaimPolicy: Retain
   storageClassName: efs-sc
-  mountOptions:
-    - tls
   csi:
     driver: efs.csi.aws.com
     volumeHandle: fs-4af69aab
+    volumeAttributes:
+      encryptInTransit: "true"
```
jqmichael, line 17: Emm, if this is enabled by default, do we still need to set it explicitly in example? 

I think examples should provide a recommended usage pattern for users and I believe we encourage users to come with the mind set that encryptInTransit is enabled without them doing anything?

pkg/driver/node.go
```diff
@@ -140,6 +148,17 @@ func (d *Driver) NodePublishVolume(ctx context.Context, req *csi.NodePublishVolu
 				// Fall through; the code below will uniq for us.
 			}
 
+			if f == "tls" {
+				klog.Warning(
+					"Use of 'tls' under mountOptions is deprecated with this driver. " +
+						"Set encrypt in transit in the volumeContext instead, e.g. 'encryptInTransit: true'")
```
jqmichael, line 154: How about 

`Use of 'tls' under mountOptions is deprecated with this driver since tls is enabled by default`.

test/e2e/e2e.go
```diff
@@ -258,11 +258,64 @@ var _ = ginkgo.Describe("[efs-csi] EFS CSI", func() {
...
+			// 127.0.0.1:/ on /mnt/volume1 type nfs4 (rw,relatime,vers=4.1,rsize=1048576,wsize=1048576,namlen=255,hard,noresvport,proto=tcp,port=20052,timeo=600,retrans=2,sec=sys,clientaddr=127.0.0.1,local_lock=none,addr=127.0.0.1)
+			// Note the tls option is not actually there. The proof that tls is passed
+			// to mount.efs is the server is 127.0.0.1
+			// (stunnel proxy running on localhost)
+			// instead of the EFS DNS name
+			// (file-system-id.efs.aws-region.amazonaws.com)
+			command := "mount && mount | grep /mnt/volume1 | grep 127.0.0.1"
+			if encryptInTransit != nil {
+				if !*encryptInTransit {
+					command = fmt.Sprintf("mount && mount | grep /mnt/volume1 | grep %v", FileSystemId)
```
jqmichael, line 286: What does the first `mount` do from `mount && mount` ?

wongma7: That is just to print it for debugging

examples/kubernetes/encryption_in_transit/specs/pv.yaml
wongma7: IDK, depends how we wish to differentiate this example from the static_provisioning example. If we do not show how to set encryptInTransit then this example is pointless. I think it's fine just to have it in the pv yaml so users can easily toggle it by doing s/true/false/ & vice versa.

pkg/driver/node.go
wongma7: fixed. I still mention encryptInTransit to disable .

test/e2e/e2e.go
wongma7: added msg

wongma7:
Updated README

docs/README.md
```diff
@@ -24,7 +24,7 @@ The following CSI interfaces are implemented:
 ### Encryption In Transit
 One of the advantages of using EFS is that it provides [encryption in transit](https://aws.amazon.com/blogs/aws/new-encryption-of-data-in-transit-for-amazon-efs/) support using TLS. Using encryption in transit, data will be encrypted during its transition over the network to the EFS service. This provides an extra layer of defence-in-depth for applications that requires strict security compliance.
 
-To enable encryption in transit, `tls` needs to be set in the `NodePublishVolumeRequest.VolumeCapability.MountVolume` object's `MountFlags` fields. For an example of using it in kubernetes, see the persistence volume manifest in [Encryption in Transit Example](../examples/kubernetes/encryption_in_transit/specs/pv.yaml)
+Encryption in transit is enabled by default. To disable it and mount volumes using plain NFSv4, set `volumeAttributes` field `encryptInTransit` to `"false"` in your persistent volume manifest. For an example manifest, see [Encryption in Transit Example](../examples/kubernetes/encryption_in_transit/specs/pv.yaml).
```
jqmichael, line 27: > Encryption in transit is enabled by default.

Should we mention `starting from x.x version`?

wongma7: I said "enabled by default in the master branch version of the driver" since the README in charts and such makes reference to "master branch"

examples/kubernetes/encryption_in_transit/specs/pv.yaml
jqmichael: I think it might make sense to set `encryptInTransit` to `false` in the static_provisioning example. Will send a PR after this is merged.

examples/kubernetes/encryption_in_transit/README.md
```diff
@@ -18,21 +18,21 @@ spec:
...
-  mountOptions:
-    - tls
   csi:
     driver: efs.csi.aws.com
     volumeHandle: [FileSystemId] 
+    volumeAttributes:
+      encryptInTransit: "true"
 ```
-Note that encryption in transit is enabled using mount option `tls`. Replace `VolumeHandle` value with `FileSystemId` of the EFS filesystem that needs to be mounted.
+Note that encryption in transit is configured using volume attribute `encryptInTransit`. By default, encryption in transit is enabled and there is no need to set `encryptInTransit` true. Replace `VolumeHandle` value with `FileSystemId` of the EFS filesystem that needs to be mounted.
```
nckturner, line 27: I think you should add the table from the PR to this README file, I think its more clear than this description.

wongma7: PTAL. I just replaced it with "The following table illustrates how the setting of encryptInTransit determines whether encryption in transit is enabled or not: ..."

wongma7:
> What's the reason of leveraging `volumeAttributes` for opt-out encryption in transit? One thing feels confusing and overlay complicated to me is the behavior matrix for different `tls` and `encryptionInTransit` combinations. Is `volumeAttribes` intended to be used as dynamic value such as mount option? I understand `volumeAttributes` as static attributes that tires to a specific value no matter how is it mounted. And `mountOptions` is used to provide dynamic values such as how the same volume should be mounted, etc. But I might be wrong, have you checked with CSI spec?

I don't really see it like this, both mountOptions and volumeAttributes tell the system how to mount the volume, neither is  "dynamic" or "static." Once a PV is created, its PVsource including mountOptions and volumeAttributes is immutable, so I think by your definition they are both "static"

> Another way could be introducing a new mount option `no-tls`, and we do not set `tls` mount helper option when `no-tls` is set, and vice versa. This feels much easier and consistent to the way it works currently

On one hand I would like mount.efs to provide a "notls" option similar to how nfs has lock/nolock, acl/noacl, etc. But I don't want to trick users into thinking that "no-tls" is a real mount option. mountOptions is a field set by people who know what they're doing to pass options to the actual mount command, if we abuse it to accept fake mount options it will cause even more confusion.

jqmichael:
> But I don't want to trick users into thinking that "no-tls" is a real mount option. mountOptions is a field set by people who know what they're doing to pass options to the actual mount command, if we abuse it to accept fake mount options it will cause even more confusion.

+1 

@leakingtapan

PR #206, opened 2020-07-01
Update amazon-efs-utils to 1.26-2

wongma7:

**What is this PR about? / Why do we need it?** Routine bump in preparation for release + I want specifically the fix for bug https://github.com/aws/efs-utils/issues/69 which can be triggered pretty reliably by the driver Pod being restarted.

**What testing is done?** Package exists in AL2 yum, I will let CI exercise tests.
TODO: validate that https://github.com/aws/efs-utils/issues/69 has been fixed by restarting driver after it has done a couple of tls mounts. Pretty hard to automate this test.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -25,7 +25,7 @@ ADD . .
 RUN make aws-efs-csi-driver
 
 FROM amazonlinux:2.0.20200406.0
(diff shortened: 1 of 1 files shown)


wongma7:
/hold
while I do my manual validation that efs-utils bug was fixed

wongma7:
/close

We are going to wait for 1.26-3 which will also have fix https://github.com/aws/efs-utils/issues/74

PR #212, opened 2020-07-09
Started treating the efs-utils config dir stateful and also handles the static files installed at image build time

jqmichael:
/feature

This PR added https://github.com/kubernetes-sigs/aws-efs-csi-driver/pull/196 back by only patching node.yaml in the `dev` overlay. 

This PR also start treating the efs-utils config directory stateful and handles static files installed at image build time. 

At image build time, static files installed by efs-utils in the config directory, i.e. CAs file, need to be saved in another place so that the other stateful files created at runtime, i.e. private key for client certificate, in the same config directory can be persisted to host with a host path volume. Otherwise creating a host path volume for that directory will clean up everything inside at the first time. Those static files need to be copied back to the config directory when the driver starts up.

**What testing is done?** 
Tested by verifying both tls and non-tls mount installed by 0.3.0 driver can be safely upgraded to the latest driver code without much interruption (other than during the downtime of the old driver).

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -27,8 +27,9 @@ import (
 
 func main() {
 	var (
-		endpoint = flag.String("endpoint", "unix://tmp/csi.sock", "CSI Endpoint")
-		version  = flag.Bool("version", false, "Print the version and exit")
+		endpoint        = flag.String("endpoint", "unix://tmp/csi.sock", "CSI Endpoint")
+		version         = flag.Bool("version", false, "Print the version and exit")
+		efsUtilsCfgPath = flag.String("efs-utils-config-path", "/etc/amazon/efs/efs-utils.conf", "The path to efs-utils config")
 	)
 	klog.InitFlags(nil)
 	flag.Parse()
@@ -42,7 +43,7 @@ func main() {
 		os.Exit(0)
 	}
 
-	drv := driver.NewDriver(*endpoint)
+	drv := driver.NewDriver(*endpoint, *efsUtilsCfgPath)
 	if err := drv.Run(); err != nil {
 		klog.Fatalln(err)
 	}
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -21,15 +21,17 @@ import (
 	"fmt"
 	"os"
 
-	"github.com/kubernetes-sigs/aws-efs-csi-driver/pkg/driver"
 	"k8s.io/klog"
+
+	"github.com/kubernetes-sigs/aws-efs-csi-driver/pkg/driver"
 )
 
 func main() {
 	var (
-		endpoint        = flag.String("endpoint", "unix://tmp/csi.sock", "CSI Endpoint")
-		version         = flag.Bool("version", false, "Print the version and exit")
-		efsUtilsCfgPath = flag.String("efs-utils-config-path", "/etc/amazon/efs/efs-utils.conf", "The path to efs-utils config")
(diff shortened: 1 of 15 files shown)


deploy/kubernetes/overlays/stable/ecr/kustomization.yaml
```diff
@@ -0,0 +1,14 @@
...
+images:
+  - name: amazon/aws-efs-csi-driver
+    newName: 602401143452.dkr.ecr.us-west-2.amazonaws.com/eks/aws-efs-csi-driver
+    newTag: v0.3.0
+  - name: quay.io/k8scsi/livenessprobe
+    newName: 602401143452.dkr.ecr.us-west-2.amazonaws.com/eks/csi-liveness-probe
+    newTag: v1.1.0
+  - name: quay.io/k8scsi/csi-node-driver-registrar
+    newName: 602401143452.dkr.ecr.us-west-2.amazonaws.com/eks/csi-node-driver-registrar
+    newTag: v1.1.0
```
wongma7, line 14: should this be different PR? (I don't particularly mind including this, just want to make sure you didn't commit stuff by accident)

jqmichael: Yeah that was intentionally. I agree that should be at least a separate commit.

pkg/driver/efs_watch_dog.go
```diff
@@ -39,22 +101,116 @@ type execWatchdog struct {
...
+			}
+		} else {
+			klog.Infof("Skip copying %s since it exists already", dst)
+		}
+	}
+	return nil
+}
+
+func copyFile(src, dst string) error {
+	source, err := os.Open(src)
```
wongma7, line 186: If src is a directory, this code will fail, which is good because we will catch it fast. If for some reason efs-utils in a new version starts persisting more stuff into /etc/amazon/efs.

cmd/main.go
```diff
@@ -21,14 +21,17 @@ import (
...
+	"github.com/kubernetes-sigs/aws-efs-csi-driver/pkg/driver"
 )
 
 func main() {
 	var (
-		endpoint = flag.String("endpoint", "unix://tmp/csi.sock", "CSI Endpoint")
-		version  = flag.Bool("version", false, "Print the version and exit")
+		endpoint                = flag.String("endpoint", "unix://tmp/csi.sock", "CSI Endpoint")
+		version                 = flag.Bool("version", false, "Print the version and exit")
+		efsUtilsCfgDirPath      = flag.String("efs-utils-config-dir-path", "/etc/amazon/efs/", "The path to efs-utils config directory")
```
wongma7, line 33: hmmm are these flags needed, would a user ever want to change them?

wongma7:
lgtm, just having second thoughts about the flags if they are needed only for debugging purposes or what

pkg/driver/efs_watch_dog.go
jqmichael: That's a great point. Added a unit test to ensure it failed in this case.

cmd/main.go
jqmichael: I was expecting users would like to change them when running the driver locally for testing (in order to do that we should not talk to IMDS always, instead passing instanceId and region from a flag).

wongma7: efs-utils will always read from /etc/amazon/efs, I think this is just going to confuse people because changing it doesn't actually do anything. All that happens is the driver will copy the files from static path to a different path. But I only know that from looking at the code, a user is not going to know or care

PR #213, opened 2020-07-16
Added e2e test for driver version upgrade

jqmichael:
feature 

Added e2e test for driver version upgrade

**What testing is done?** 
```
go test ./test/e2e/ -v -kubeconfig=$HOME/.kube/config --region=us-west-2 --report-dir="./results" -ginkgo.focus="\[restart\]|\[upgrade\]" --cluster-name="eks-11602" --file-system-id fs-b6654c1c 
```

### DIFF
--- a/pkg/driver/efs_watch_dog.go
+++ b/pkg/driver/efs_watch_dog.go
@@ -71,7 +71,7 @@ dns_name_suffix = sc2s.sgov.gov
 [mount-watchdog]
 enabled = true
 poll_interval_sec = 1
-unmount_grace_period_sec = 30
+unmount_grace_period_sec = 5
 
 # Set client auth/access point certificate renewal rate. Minimum value is 1 minute.
 tls_cert_renewal_interval_min = 60
--- a/test/e2e/e2e.go
+++ b/test/e2e/e2e.go
@@ -5,7 +5,9 @@ import (
 	"time"
 
 	"github.com/onsi/ginkgo"
+	appsv1 "k8s.io/api/apps/v1"
 	v1 "k8s.io/api/core/v1"
+
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -13,6 +15,7 @@ import (
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/util/sets"
 	clientset "k8s.io/client-go/kubernetes"
+	clientgoappsv1 "k8s.io/client-go/kubernetes/typed/apps/v1"
 	"k8s.io/kubernetes/test/e2e/framework"
 	e2enode "k8s.io/kubernetes/test/e2e/framework/node"
 	e2epod "k8s.io/kubernetes/test/e2e/framework/pod"
@@ -182,7 +185,7 @@ var _ = ginkgo.Describe("[efs-csi] EFS CSI", func() {
 	f := framework.NewDefaultFramework("efs")
 
 	ginkgo.Context(testsuites.GetDriverNameWithFeatureTags(driver), func() {
-		ginkgo.It("should mount different paths on same volume on same node", func() {
+		ginkgo.It("[efs] [nonTls] should mount different paths on same volume on same node", func() {
 			ginkgo.By(fmt.Sprintf("Creating efs pvc & pv with no subpath"))
 			pvcRoot, pvRoot, err := createEFSPVCPV(f.ClientSet, f.Namespace.Name, f.Namespace.Name+"-root", "/")
 			framework.ExpectNoError(err, "creating efs pvc & pv with no subpath")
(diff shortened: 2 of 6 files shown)


pkg/driver/efs_watch_dog.go
```diff
@@ -71,7 +71,7 @@ dns_name_suffix = sc2s.sgov.gov
 [mount-watchdog]
 enabled = true
 poll_interval_sec = 1
-unmount_grace_period_sec = 30
+unmount_grace_period_sec = 5
```
wongma7, line 74: What are the implications of changing this? I'm OK with changing it for a test, but it seems wrong to change it for everyone.

test/e2e/e2e.go
```diff
@@ -275,8 +396,12 @@ func createEFSPVCPV(c clientset.Interface, namespace, name, path string) (*v1.Pe
 }
 
 func makeEFSPVCPV(namespace, name, path string) (*v1.PersistentVolumeClaim, *v1.PersistentVolume) {
+	pv := makeEFSPvWithoutTLS(name, path)
+	return makePVCWithPV(pv, namespace, name)
+}
+
+func makePVCWithPV(pv *v1.PersistentVolume, namespace, name string) (*v1.PersistentVolumeClaim, *v1.PersistentVolume) {
```
wongma7, line 403: This function name is confusing. "With" could mean "make a PVC *along* with a PV".

I suggest separating the binding code below (setting volumeName of a given PVC and claimRef of a given PV) to a separate function.

test/e2e/e2e.go
```diff
@@ -211,58 +214,176 @@ var _ = ginkgo.Describe("[efs-csi] EFS CSI", func() {
...
+			defer func() { f.ClientSet.CoreV1().PersistentVolumes().Delete(pv.Name, &metav1.DeleteOptions{}) }()
+
+			node, pod := launchPod(f, pv, pvc)
+
+			deleteDriverPodOnNode(f, node)
+
+			validatePodVolumeAccessible(f, pod)
+		})
+
+		ginkgo.It("[efs] [github] [tls] [upgrade] should continue reading/writing after the driver pod is upgraded from stable version", func() {
```
wongma7, line 247: What is the meaning of the new tags? It's not necessary to have an [efs] tag because everything is already tagged "[efs-csi] EFS CSI"

test/e2e/e2e.go
```diff
@@ -211,58 +214,176 @@ var _ = ginkgo.Describe("[efs-csi] EFS CSI", func() {
...
+	}
+}
+
+func launchPod(f *framework.Framework, pv *v1.PersistentVolume, pvc *v1.PersistentVolumeClaim) (*v1.Node, *v1.Pod) {
+	node, err := e2enode.GetRandomReadySchedulableNode(f.ClientSet)
+	framework.ExpectNoError(err, "getting random ready schedulable node")
+	command := fmt.Sprintf("touch /mnt/volume1/%s-%s && trap exit TERM; while true; do sleep 1; done", f.Namespace.Name, time.Now().Format(time.RFC3339))
+
+	isPrivileged := false
+	var nodeSelector map[string]string
```
wongma7, line 363: nodeSelector is not set correctly, something got lost in the refactoring.

wongma7: It seems like nodeSelector was never set correctly?

test/e2e/e2e.go
```diff
@@ -211,58 +214,176 @@ var _ = ginkgo.Describe("[efs-csi] EFS CSI", func() {
...
+	framework.ExpectNoError(err, "getting driver pod")
+	framework.ExpectEqual(len(podList.Items), 1, "expected 1 efs csi node pod but got %d", len(podList.Items))
+	driverPod := podList.Items[0]
+
+	ginkgo.By(fmt.Sprintf("Deleting driver pod %q on node %q", driverPod.Name, node.Name))
+	err = e2epod.DeletePodWithWaitByName(f.ClientSet, driverPod.Name, "kube-system")
+	framework.ExpectNoError(err, "deleting driver pod")
+}
+
+func createPvPvcWithPv(f *framework.Framework, pv *v1.PersistentVolume) (*v1.PersistentVolume, *v1.PersistentVolumeClaim) {
```
wongma7, line 332: Sorry but I hate this function name

test/e2e/e2e.go
```diff
@@ -211,58 +214,176 @@ var _ = ginkgo.Describe("[efs-csi] EFS CSI", func() {
...
+	_, err = daemonSetClient.Update(driverUpgradeTo)
+	framework.ExpectNoError(err, "Updating to the EFS CSI driver")
+}
+
 func createEFSPVCPV(c clientset.Interface, namespace, name, path string) (*v1.PersistentVolumeClaim, *v1.PersistentVolume, error) {
 	pvc, pv := makeEFSPVCPV(namespace, name, path)
+	return createEFSPVPVC(c, namespace, pvc, pv)
+}
+
+func createEFSPVPVC(c clientset.Interface, namespace string, pvc *v1.PersistentVolumeClaim, pv *v1.PersistentVolume) (*v1.PersistentVolumeClaim, *v1.PersistentVolume, error) {
```
wongma7, line 386: createEFSPVCPV vs createEFSPVPVC?

test/e2e/e2e.go
wongma7: Yeah I never set it: https://github.com/kubernetes-sigs/aws-efs-csi-driver/blob/master/test/e2e/e2e.go#L225 so with 3 nodes we had a 1/3 chance of actually running the test as intended.

I wouldn't use nodeSelector anyway ,I would rather just set pod.spec.Nodename

test/e2e/e2e.go
```diff
@@ -211,58 +214,176 @@ var _ = ginkgo.Describe("[efs-csi] EFS CSI", func() {
...
+	framework.ExpectNoError(e2epod.WaitForPodNameRunningInNamespace(f.ClientSet, pod.Name, f.Namespace.Name), "waiting for pod running")
+
+	return node, pod
+}
+
+func updateDriverTo(daemonSetClient clientgoappsv1.DaemonSetInterface, driverUpgradeTo *appsv1.DaemonSet) {
+	ginkgo.By("Updating to the EFS CSI driver")
+	currentEfsDriver, err := daemonSetClient.Get("efs-csi-node", metav1.GetOptions{})
+	driverUpgradeTo.ResourceVersion = currentEfsDriver.ResourceVersion
+	_, err = daemonSetClient.Update(driverUpgradeTo)
```
wongma7, line 377: Need to wait for daemonset to be totally replaced. Or at least the driver pod on the node we care about

wongma7:
I might pick this up depending on whether it's easier to manually do the test or fix it : p

PR #215, opened 2020-07-16
Build and push every master commit to master tag

wongma7:

**What is this PR about? / Why do we need it?** dev overlay still points to tag latest but we stopped pushing to tag latest. So the dev overlay is out of sync with the image. To put it back in sync, use tag `master`.

**What testing is done?** N/A

### DIFF
--- a/deploy/kubernetes/overlays/dev/master_image.yaml
+++ b/deploy/kubernetes/overlays/dev/master_image.yaml
@@ -8,4 +8,5 @@ spec:
     spec:
       containers:
         - name: efs-plugin
-          image: amazon/aws-efs-csi-driver:latest
+          image: amazon/aws-efs-csi-driver:master
+          imagePullPolicy: Always
--- a/docs/README.md
+++ b/docs/README.md
@@ -42,7 +42,7 @@ The following sections are Kubernetes specific. If you are a Kubernetes user, us
 ### Container Images
 |EFS CSI Driver Version     | Image                               |
 |---------------------------|-------------------------------------|
-|master branch              |amazon/aws-efs-csi-driver:latest     |
+|master branch              |amazon/aws-efs-csi-driver:master     |
 |v0.3.0                     |amazon/aws-efs-csi-driver:v0.3.0     |
 |v0.2.0                     |amazon/aws-efs-csi-driver:v0.2.0     |
 |v0.1.0                     |amazon/aws-efs-csi-driver:v0.1.0     |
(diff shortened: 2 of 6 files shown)


hack/run-e2e-test
```diff
@@ -78,7 +78,7 @@ done;
 
 echo "Deploying driver"
 sed -i'' "s,amazon/aws-efs-csi-driver,$IMAGE_NAME," deploy/kubernetes/overlays/dev/kustomization.yaml
```
jqmichael, line 80: This script doesn't seem to be used anyway.

docs/README.md
```diff
@@ -42,7 +42,7 @@ The following sections are Kubernetes specific. If you are a Kubernetes user, us
 ### Container Images
 |EFS CSI Driver Version     | Image                               |
 |---------------------------|-------------------------------------|
-|master branch              |amazon/aws-efs-csi-driver:latest     |
+|master branch              |amazon/aws-efs-csi-driver:master     |
```
jqmichael, line 45: Out of scope of this PR, but I think we should discourage users from using the latest/"master" tag by not mentioning it anywhere in the README. If users find it by reading the code, that's OK. But that doesn't apply to the majority of the users.

deploy/kubernetes/overlays/dev/master_image.yaml
```diff
@@ -8,4 +8,4 @@ spec:
     spec:
       containers:
         - name: efs-plugin
-          image: amazon/aws-efs-csi-driver:latest
+          image: amazon/aws-efs-csi-driver:master
```
jqmichael, line 11: We might need to add ImagePullPolicy: Always since only kubelet only always pull the ":latest" tag  without a policy. 

> If you would like to always force a pull, you can do one of the following:

> set the imagePullPolicy of the container to Always.
omit the imagePullPolicy and use :latest as the tag for the image to use.
omit the imagePullPolicy and the tag for the image to use.
enable the AlwaysPullImages admission controller.


https://kubernetes.io/docs/concepts/containers/images/#updating-images

wongma7: good point, fixed

hack/run-e2e-test
wongma7: You're right, I did a grep of latest and made changes everywhere. Technically someone could still use it (instead of using https://github.com/kubernetes-sigs/aws-efs-csi-driver/blob/master/tester/e2e-test-config.yaml which requires the k8stester binary), but probably we should remove it so we don't have to maintain it.

PR #224, opened 2020-07-23
Fix restart test creating pod on random, potentially wrong, node

wongma7:
**Is this a bug fix or adding new feature?** Bug fix.

**What is this PR about? / Why do we need it?** The test is supposed to create a pod on a certain node and then restart the driver pod on that same node to test that read/write still works. But the test did not specify a node to create the pod on, so actually, since our CI creates 3 nodes, there was only a 33% chance the test was working as intended. Otherwise 66% of the time we were getting false positives.

**What testing is done?**  CI


wongma7:
coveralls bug: I did not change any code under pkg/driver but it thinks coverage decreased.

I'm going to manually merge this.

PR #226, opened 2020-07-24
Actually use the existing nodeSelector value for the deamonset

Quarky9:
This allows to select only workers having the EFS mounted via labels.

Bug fix I think.

Improves helm deployment

**What testing is done?** 
Tested on personal test cluster


wongma7:
/ok-to-test
/lgtm
/approve

thanks

wongma7:
Going to manual merge...coverlay is bugged so it's preventing the bot from merging, but all required tests passed.

PR #230, opened 2020-07-28
Fix mounting /etc/amazon/efs in the node-driver-registrar container instead of the efs-plugin one

wongma7:
**Is this a bug fix or adding new feature?** bug fix.

**What is this PR about? / Why do we need it?** Since I put the volume in the wrong container, the contents of /etc/amazon/efs are not persisted across driver restarts. Notably, the privateKey.pem. Which can cause tls mounts to hang.

**What testing is done?** /etc/amazon/efs contains persisted files with this change to the YAML


wongma7:
Note helm install is not affected by this.

PR #238, opened 2020-08-10
Implement NodeGetVolumeStats

kbasv:
New Feature

* Implements NodeGetVolumeStats using `du` under the hood and enables GET_VOLUME_STATS node capability.
* Added a cache (map) to store the results of `du` since `du` can take longer than the default kubelet timeout for NodeGetVolumeStats rpc. 
* Provide a refreshRate option to control the rate at which `du` is invoked.

**What testing is done?** 
Added Unit tests to confirm the behavior. 
Tested the implementation E2E with EKS on EC2.

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -32,6 +32,9 @@ func main() {
 		version                 = flag.Bool("version", false, "Print the version and exit")
 		efsUtilsCfgDirPath      = flag.String("efs-utils-config-dir-path", "/etc/amazon/efs/", "The path to efs-utils config directory")
 		efsUtilsStaticFilesPath = flag.String("efs-utils-static-files-path", "/etc/amazon/efs-static-files/", "The path to efs-utils static files directory")
+		volMetricsOptIn         = flag.Bool("vol-metrics-opt-in", false, "Opt in to emit volume metrics")
+		volMetricsRefreshPeriod = flag.Float64("vol-metrics-refresh-period", 240, "Refresh period for volume metrics in minutes")
+		volMetricsFsRateLimit   = flag.Int("vol-metrics-fs-rate-limit", 5, "Volume metrics routines rate limiter per file system")
 	)
 	klog.InitFlags(nil)
 	flag.Parse()
@@ -45,7 +48,7 @@ func main() {
 		os.Exit(0)
 	}
 
-	drv := driver.NewDriver(*endpoint, *efsUtilsCfgDirPath, *efsUtilsStaticFilesPath)
+	drv := driver.NewDriver(*endpoint, *efsUtilsCfgDirPath, *efsUtilsStaticFilesPath, *volMetricsOptIn, *volMetricsRefreshPeriod, *volMetricsFsRateLimit)
 	if err := drv.Run(); err != nil {
 		klog.Fatalln(err)
 	}
--- a/pkg/driver/node.go
+++ b/pkg/driver/node.go
@@ -32,11 +32,11 @@ import (
 )
 
 var (
-	nodeCaps             = []csi.NodeServiceCapability_RPC_Type{}
 	volumeCapAccessModes = []csi.VolumeCapability_AccessMode_Mode{
 		csi.VolumeCapability_AccessMode_SINGLE_NODE_WRITER,
 		csi.VolumeCapability_AccessMode_MULTI_NODE_MULTI_WRITER,
 	}
+	volumeIdCounter = make(map[string]int)
 )
 
 func (d *Driver) NodeStageVolume(ctx context.Context, req *csi.NodeStageVolumeRequest) (*csi.NodeStageVolumeResponse, error) {
@@ -176,6 +176,15 @@ func (d *Driver) NodePublishVolume(ctx context.Context, req *csi.NodePublishVolu
 	}
 	klog.V(5).Infof("NodePublishVolume: %s was mounted", target)
 
+	//Increment volume Id counter
+	if d.volMetricsOptIn {
+		if value, ok := volumeIdCounter[req.GetVolumeId()]; ok {
+			volumeIdCounter[req.GetVolumeId()] = value + 1
+		} else {
+			volumeIdCounter[req.GetVolumeId()] = 1
+		}
+	}
+
 	return &csi.NodePublishVolumeResponse{}, nil
 }
 
@@ -211,11 +220,55 @@ func (d *Driver) NodeUnpublishVolume(ctx context.Context, req *csi.NodeUnpublish
 	}
 	klog.V(5).Infof("NodeUnpublishVolume: %s unmounted", target)
 
+	//TODO: If `du` is running on a volume, unmount waits for it to complete. We should stop `du` on unmount in the future for NodeUnpublish
+	//Decrement Volume ID counter and evict cache if counter is 0.
+	if d.volMetricsOptIn {
+		if value, ok := volumeIdCounter[req.GetVolumeId()]; ok {
+			value -= 1
+			if value < 1 {
+				klog.V(4).Infof("Evicting vol ID: %v, vol path : %v from cache", req.VolumeId, target)
+				d.volStatter.removeFromCache(req.VolumeId)
+				delete(volumeIdCounter, req.GetVolumeId())
+			} else {
+				volumeIdCounter[req.GetVolumeId()] = value
+			}
+		}
+	}
+
 	return &csi.NodeUnpublishVolumeResponse{}, nil
 }
 
 func (d *Driver) NodeGetVolumeStats(ctx context.Context, req *csi.NodeGetVolumeStatsRequest) (*csi.NodeGetVolumeStatsResponse, error) {
-	return nil, status.Error(codes.Unimplemented, "")
+	klog.V(4).Infof("NodeGetVolumeStats: called with args %+v", req)
+
+	volId := req.GetVolumeId()
+	if volId == "" {
+		return nil, status.Error(codes.InvalidArgument, "Volume ID not provided")
+	}
+
+	target := req.GetVolumePath()
+	if target == "" {
+		return nil, status.Error(codes.InvalidArgument, "Volume Path not provided")
+	}
+
+	_, err := os.Stat(target)
+	if err != nil {
+		if os.IsNotExist(err) {
+			return nil, status.Errorf(codes.NotFound, "Volume Path %s does not exist", target)
+		}
+
+		return nil, status.Errorf(codes.Internal, "Failed to invoke stat on volume path %s: %v", target, err)
+	}
+
+	volMetrics, err := d.volStatter.computeVolumeMetrics(volId, target, d.volMetricsRefreshPeriod, d.volMetricsFsRateLimit)
+
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "Could not get metrics: %v ", err)
+	}
+
+	return &csi.NodeGetVolumeStatsResponse{
+		Usage: volMetrics.volUsage,
+	}, nil
 }
 
 func (d *Driver) NodeExpandVolume(ctx context.Context, req *csi.NodeExpandVolumeRequest) (*csi.NodeExpandVolumeResponse, error) {
@@ -225,7 +278,7 @@ func (d *Driver) NodeExpandVolume(ctx context.Context, req *csi.NodeExpandVolume
 func (d *Driver) NodeGetCapabilities(ctx context.Context, req *csi.NodeGetCapabilitiesRequest) (*csi.NodeGetCapabilitiesResponse, error) {
 	klog.V(4).Infof("NodeGetCapabilities: called with args %+v", req)
 	var caps []*csi.NodeServiceCapability
-	for _, cap := range nodeCaps {
+	for _, cap := range d.nodeCaps {
 		c := &csi.NodeServiceCapability{
 			Type: &csi.NodeServiceCapability_Rpc{
 				Rpc: &csi.NodeServiceCapability_RPC{
--- a/pkg/driver/node_test.go
+++ b/pkg/driver/node_test.go
@@ -19,8 +19,11 @@ package driver
 import (
 	"context"
 	"fmt"
+	"os"
+	"reflect"
 	"strings"
 	"testing"
+	"time"
 
 	"github.com/container-storage-interface/spec/lib/go/csi"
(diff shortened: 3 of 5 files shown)


kbasv:
I signed it

cmd/main.go
```diff
@@ -45,7 +47,8 @@ func main() {
 		os.Exit(0)
 	}
 
-	drv := driver.NewDriver(*endpoint, *efsUtilsCfgDirPath, *efsUtilsStaticFilesPath)
+	drv := driver.NewDriver(*endpoint, *efsUtilsCfgDirPath, *efsUtilsStaticFilesPath, *volMetricsOptIn, *volMetricsRefreshRate)
+	drv.SetNodeCapOptInFeatures(*volMetricsOptIn)
```
wongma7, line 51: y not hide this in NewDriver since NewDriver takes in volMetricsOptIn anyway? not sure the benefit of a separate function

pkg/driver/node.go
```diff
@@ -187,6 +186,11 @@ func (d *Driver) NodeUnpublishVolume(ctx context.Context, req *csi.NodeUnpublish
 		return nil, status.Error(codes.InvalidArgument, "Target path not provided")
 	}
 
+	if d.volMetricsOptIn {
+		klog.V(4).Infof("Evicting vol ID: %v, vol path : %v from cache", req.VolumeId, target)
```
wongma7, line 190: it's possible to publish the same VolumeId to different targets.

I don't want to complicate this too much, but do we need some kind of ref counter to account for that so we only evict from cache if counter == 0?

pkg/driver/vol_statter.go
```diff
@@ -0,0 +1,134 @@
...
+
+	used, err := fs.DiskUsage(volPath)
+
+	if err != nil {
+		klog.Errorf("Failed to compute volume usage on path %s: %v",volPath, err)
+		return
+	}
+
+	volUsed, ok := used.AsInt64()
+
```
wongma7, line 104: nitpick, would like to see whitespace/newlines tightened up like here. Doesn't need to be a newline between the call and error checking. gofmt may help: https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_aws-efs-csi-driver/238/pull-aws-efs-csi-driver-verify/1293265088089690114#1:build-log.txt%3A278

pkg/driver/vol_statter.go
```diff
@@ -0,0 +1,134 @@
...
+	} else {
+		volStatterJobTracker[volId] = true
+		go v.computeDiskUsage(volId, volPath)
+	}
+}
+
+func (v VolStatterImpl) computeDiskUsage(volId, volPath string) {
+	klog.V(5).Infof("Compute Volume Metrics invoked for Vol ID: %v", volId)
+
+	used, err := fs.DiskUsage(volPath)
```
wongma7, line 96: what happens if Unpublish is called on this volume in the middle of this routine? Would DiskUsage block the Unpublish unmount for the duration it's reading the volume?

cmd/main.go
```diff
@@ -32,6 +32,8 @@ func main() {
 		version                 = flag.Bool("version", false, "Print the version and exit")
 		efsUtilsCfgDirPath      = flag.String("efs-utils-config-dir-path", "/etc/amazon/efs/", "The path to efs-utils config directory")
 		efsUtilsStaticFilesPath = flag.String("efs-utils-static-files-path", "/etc/amazon/efs-static-files/", "The path to efs-utils static files directory")
+		volMetricsOptIn         = flag.Bool("volMetricsOptIn", false, "Opt in to emit volume metrics")
```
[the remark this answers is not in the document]
kbasv: Sure, I'll update this to reflect other args

pkg/driver/node.go
kbasv: I think counter will be a bit complex here. Do you know if a volume Id and volume path combination is unique? If yes, I can use it as the cache key

pkg/driver/vol_statter.go
kbasv: Sure, will update this

pkg/driver/node.go
wongma7: It will be unique but if we do that then we will be recalculating the same volume multiple times. Like if I have a volume fs-abcd:/root mounted to both /pod1/target and /pod2/target, the routine will calculate the disk usage of /root in fs-abcd twice, once for /pod1/target and again for /pod2/target which is unnecessary.

pkg/driver/vol_statter.go
kbasv: The logs from my testing indicate UnPublish will not succeed when DiskUsage is running. I see "device is busy" associated with Unpublish command in the logs. Do you think it's okay to wait it out?

pkg/driver/node.go
kbasv: I now see what you mean with the counter, sure we can add a counter to evict only when all other volumeIds are unpublished.

cmd/main.go
```diff
@@ -32,6 +32,8 @@ func main() {
...
+		volMetricsRefreshRate   = flag.Float64("volMetricsRefreshRate", 5, "Refresh rate for volume metrics in minutes")
```
wongma7, line 36: I think this should be renamed to period instead of rate

pkg/driver/vol_statter.go
wongma7: Darn, well I think it's okay to wait it out. We can leave it as a future improvement because we're relying on a library (fs.DiskUsage) and there's no way to cancel it.

In the worst-case, Unpublish could be delayed indefinitely if ` volume metrics period < computeDiskUsage execution time` and computeDiskUsage ends up running 100% of the time.

In a less-bad case, Unpublish could be delayed by `time it takes for DiskUsage to complete + 2 minutes`. The 2 minutes comes from the exponential back off for Unpublish retries. https://github.com/kubernetes/kubernetes/blob/323f34858de18b862d43c40b2cced65ad8e24052/pkg/util/goroutinemap/exponentialbackoff/exponential_backoff.go#L33

kbasv: I'll add a TODO to figure out a way to cancel running DU process in the future

pdrakeweb:
Generally, the code here makes sense to me, however the implementation seems a bit problematic.  If I understand correctly, the underlying tool (`du`) essentially stats every file to count the size.  For EFS file systems with many files, this may dramatically consume the available IOPS.  As many volumes may be created on a single underlying EFS, many executions of `du` may then occur in parallel, further exacerbating the performance impacts.  An ideal implementation seems like it would instead call an AWS API to query the disk usage for a specific directory on an EFS.  I recognize such an API does not exist and AWS would need to create it.

wongma7:
You're right on all counts. This PR has an optimization to avoid reading the same volumehandle (fs-abcd:/a/b) concurrently on the same node but there's certainly room for more. If we think that the performance penalty of enabling this feature will outweigh the benefit more often than not, we can do more optimizations first. Personally I'm inclined to merge and collect feedback, but I don't want people to enable the feature and then be surprised if their credits get exhausted, so I guess we ought to test & evaluate the impact more first.

Some brainstorming:
- We don't account for double reading of parent/child directories (fs-abcd:/a/b/c and fs-abcd:/a/b). Maybe we could only read leaf directories.
- We don't account for the same file system being read from different nodes which we obviously have to since 1 of the main selling points of EFS is you can do that. This is not easy: would you have each daemonset member hold a per-filesystem lock (`lease` object) in kubernetes, would they write a file in the EFS file system itself as a lock, etc.?
- EFS metered size has a pretty generous eventual consistency guarantee, they say "the value represents the actual size only if the file system is not modified for a period longer than a couple of hours." https://docs.aws.amazon.com/efs/latest/ug/API_FileSystemSize.html I imagine most users care only about metered size and they want this feature to find culprits. So it follows that the driver only needs to be about as accurate as metered size to be useful, i.e. default 5 minutes refresh is probably excessive.

kbasv:
I agree with concerns on the underlying implementation's potential to consume all the available IOPS in a File System. 
From my observation while testing this implementation on a cluster, in one minute `du` stats ~32k files with 2 Percent IO consumption. I think the following two options will limit Volume Statter's IOPS consumption:
* Adding a jittered start to goroutine execution. This will minimize the number of volume statter routines executing simultaneously. 
* Introducing a rate limiter to limit the number of simultaneous executions per file system.

The above two options will address concerns for mounts at a per node basis. However, it will not address concerns of a file system mounted on multiple nodes. For a multi node mount, given how Daemonsets work, it is not easy to share states between nodes and any solution for multi node mounts seems out of scope for the current implementation.

pdrakeweb:
A start jitter and rate limiter could be helpful.  The test uncovered roughly what I would expect - that `du` manages to consume a large portion of the available IOPS.  Without QoS across clients, this may have a significant impact on applications using the volume.  This is of particular concern for larger file systems (think about volumes exceeding 10M files).  If an AWS provided API for this metadata is not possible, another option would be to introduce a mechanism for limiting the IOPS consumed by `du`.  The kernel has mechanisms (cgroups) to limit block I/O but I presume these would not apply as the underlying EFS volume is not a block device.  An alternative `du` which includes an internal IOP throttle could be used (although it would need to be written).

pkg/driver/vol_statter.go
```diff
@@ -0,0 +1,169 @@
...
+		},
+	}
+
+	volMetrics := &volMetrics{
+		volPath:   volPath,
+		timeStamp: time.Now(),
+		volUsage:  usage}
+
+	volUsageCache[volId] = volMetrics
+	delete(volStatterJobTracker, volId)
```
kerbyhughes, line 150: The map updates from this function are data races since the function is always called in a goroutine. The check in `launchVolStatsRoutine` to see if the `volId` is currently being processed is not sufficient - the check could be called at the same time as the delete here. I would suggest wrapping them in a sync mechanism. Removing the `time.Sleep(waitTime)` and stubbing out the external calls to the `fs` library will expose the data races in the test cases.

kerbyhughes: I'd say if it's invoked infrequently, then the overhead of a lock would be negligible. What's the recovery path if this panics?

kbasv: I see what you mean. Didn't realize go panics on concurrent reads and writes. Alternative would be to either use `locks` or `sync.map`. I guess `locks` are better given how kubelet works?

kerbyhughes: Correct, they are [not safe for concurrent use](https://golang.org/doc/faq#atomic_maps) and will panic if under certain conditions. I'm not overly familiar with `sync.map` but I believe it was created for optimization [under certain conditions](https://golang.org/pkg/sync/#Map) when regular locks are not meeting your performance needs.

kbasv:
@kerbyhughes I made some changes based on your comments. Can you please take a look at it?

kerbyhughes:
@kbasv I ran your updated code and the changes fixed the data races in volStatter. The race detector still shows a race where the test code itself does not wrap the map updates in a lock, but the rest looks good, thanks.

kbasv:
> @kbasv I ran your updated code and the changes fixed the data races in volStatter. The race detector still shows a race where the test code itself does not wrap the map updates in a lock, but the rest looks good, thanks.

@kerbyhughes Thanks for checking, fixed the test code with locks.

pkg/driver/vol_statter.go
```diff
@@ -0,0 +1,175 @@
...
+
+type VolStatterImpl struct {
+}
+
+func NewVolStatter() VolStatter {
+	return &VolStatterImpl{}
+}
+
+func (v VolStatterImpl) computeVolumeMetrics(volId, volPath string, refreshRate float64, fsRateLimit int) (*volMetrics, error) {
+	if value, ok := v.retrieveFromCache(volId); ok {
```
wongma7, line 52: What if I have two PVs mounted on this node with the same volume ID but different target path. Say volume A fs-123:/a and volume B fs-123:/b. It will be random which metric I get back.

wongma7: Maybe not random, but one of A or B will be reported for both A and B. 

However if you index the results cache by volume ID + path but the tracker just by volume ID, there might be situations where the metric for B is never computed. Let's say kubelet always calls get stats on A before B, then B will never get computed since the computation for A was just started and is in progress...

kbasv: IIRC, volumeId from above consists of file system ID + subpath. So `fs-123:/a` and `fs-123:/b` will be two different volumes and metric will be computed for both.

The cache will ensure that a metric for volumeId `fs-123:/a` mounted separately as two volumes will not be computed.

wongma7: ok, makes sense, i see now you distinguish between `fsid` for fs rate limit and `volid` for metrics cache

kbasv:
@wongma7 Let me know if you have more concerns on this PR. Otherwise can you please merge this? Thanks!

wongma7:
Please squash the commits and check coveralls. The bot will not merge if coveralls says coverage decreased

/lgtm
/approve

pkg/driver/node_test.go
```diff
@@ -616,8 +653,159 @@ func TestNodeUnpublishVolume(t *testing.T) {
...
+				{
+					Unit:      csi.VolumeUsage_BYTES,
+					Available: 1,
+					Used:      1,
+					Total:     2,
+				},
+			},
+		}
+	)
+	makeDir(validPath)
```
wongma7, line 684: test environment doesn't like this, not sure if https://golang.org/pkg/io/ioutil/#TempDir will work either

kbasv:
@wongma7 Looks like the coveralls test environment's temp directory creation is being flaky and that is pulling down the coverage. When running the test locally, temp directory creation works locally and the overall coverage is greater than 82%. The test environment succeeded in creating temp directory once here - https://coveralls.io/builds/32722046

wongma7:
@kbasv OK thanks for looking into it. Since the test works locally and in pull-aws-efs-csi-driver-unit (actually the output looks crazy because I see efs_watch_dog.go:236] stopping... being spammed infinitely but that's unrelated...), and technically coveralls is not supposed to block merge, I am inclined to merge this after the rest of the tests pass and leave the fix for coveralls as a followup

/lgtm

wongma7:
Manually merging due to coveralls being flaky. I'll open an issue to track it

PR #241, opened 2020-08-26
Added multi-arch support including ARM.

jqmichael:
feature. 

Added multi-arch support including ARM.

**What testing is done?** 

Verify the `/bin/aws-efs-csi-driver` binary was ARM when simulating an ARM environment with docker run. 
```
docker run --name aws-efs-csi-driver --platform linux/arm64 --rm --network host  --entrypoint="" -it jqmichael/aws-efs-csi-driver:2 /bin/bash -c "yum -y install file; file /bin/aws-efs-csi-driver"

---
/bin/aws-efs-csi-driver: ELF 64-bit LSB executable, ARM aarch64, version 1 (SYSV), statically linked, not stripped
```

Launched an `a1` instance, which is ARM architecture, and verified the driver worked with the TLS example (had to use csi-node-driver-registrar from https://github.com/raspbernetes/multi-arch-images since upstream image isn't available yet).

### DIFF
--- a/.github/workflows/container-image.yaml
+++ b/.github/workflows/container-image.yaml
@@ -8,9 +8,12 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v1
-    - name: Build container image
-      run: |
-        docker build -t aws-efs-csi-driver .
+    - name: Set up Docker Buildx
+      id: buildx
+      uses: crazy-max/ghaction-docker-buildx@v3
+      with:
+        buildx-version: latest
+        qemu-version: latest
     - name: Push to Github registry
       run: |
         USER=$(echo $GITHUB_REPOSITORY | cut -d'/' -f1)
@@ -23,6 +26,7 @@ jobs:
           TAG=$BRANCH
         fi
         docker login docker.pkg.github.com -u $USER -p ${{ secrets.REGISTRY_TOKEN }}
+        docker build -t aws-efs-csi-driver .
         docker tag aws-efs-csi-driver docker.pkg.github.com/$GITHUB_REPOSITORY/$IMAGE:$TAG
         docker push docker.pkg.github.com/$GITHUB_REPOSITORY/$IMAGE:$TAG
         if [ "$BRANCH" = "master" ]; then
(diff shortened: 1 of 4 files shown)


.github/workflows/container-image.yaml
```diff
@@ -8,9 +8,12 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v1
-    - name: Build container image
-      run: |
-        docker build -t aws-efs-csi-driver .
```
jqmichael, line 13: No need for a separate build step because `docker buildx` build and push are run at the same step.

.github/workflows/container-image.yaml
```diff
@@ -40,9 +44,16 @@ jobs:
...
         fi
         docker login -u ${{ secrets.DOCKERHUB_USER }} -p ${{ secrets.DOCKERHUB_TOKEN }}
-        docker tag aws-efs-csi-driver $REPO:$TAG
-        docker push $REPO:$TAG
+
+        docker buildx build \
+              -t $REPO:$TAG \
+              --platform=linux/amd64,linux/arm64 \
+              --progress plain \
+              --push .
```
jqmichael, line 52: `docker buildx` build and push are done as a single step.

.github/workflows/container-image.yaml
jqmichael: > The docker buildx build subcommand has a number of flags which determine where the final image will be stored. By default, i.e. if none of the flags are specified, the resulting image will remain captive in docker’s internal build cache. This is unlike the regular docker build command which stores the resulting image in the local docker images list. 

> --push: This flag tells docker to push the resulting image to a docker registry. This currently is the best way to store multi-architecture images.

https://medium.com/@artur.klauser/building-multi-architecture-docker-images-with-buildx-27d80f7e2408

.github/workflows/container-image.yaml
jqmichael: > The docker buildx build subcommand has a number of flags which determine where the final image will be stored. By default, i.e. if none of the flags are specified, the resulting image will remain captive in docker’s internal build cache. This is unlike the regular docker build command which stores the resulting image in the local docker images list. 

> --push: This flag tells docker to push the resulting image to a docker registry. This currently is the best way to store multi-architecture images.

https://medium.com/@artur.klauser/building-multi-architecture-docker-images-with-buildx-27d80f7e2408

wongma7:
merging because coveralls is bugged

PR #251, opened 2020-09-11
Update broken import vbom.ml/util with github.com/fvbommel/util

kbasv:
Bug Fix

Docker build fails at `go mod download` step with the following message:

> go: k8s.io/kubernetes@v1.17.6 requires
	k8s.io/kubectl@v0.0.0 requires
	vbom.ml/util@v0.0.0-20160121211510-db5cfe13f5cc: unrecognized import path "vbom.ml/util" (https fetch: Get https://vbom.ml/util?go-get=1: dial tcp: lookup vbom.ml on 192.168.65.1:53: no such host)

Issue is same as kubernetes/kubectl#925. 

**What testing is done?** 
Tested locally and docker build succeeds

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,6 @@
 *.swp
 bin/
 vendor/
+*.DS_Store
+*.idea
+test/e2e/results/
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -1,3 +1,6 @@
 *.swp
 bin/
 vendor/
+*.DS_Store
+*.idea
+test/e2e/results/
```
jqmichael, line 6: Could you prepend this with `/` so that it's clear only the root folder is filtered not ./abc/test/e2e/results?

jqmichael (approved):
Nice fix! Appreciate jumping on this

Issue #235, opened 2020-08-05
Update helm chart for 1.0.0 release


albertmichaelj:
**Is your feature request related to a problem?/Why is this needed**
I tried to install the helm chart today for the `aws-efs-csi-driver`, but I realized that the helm chart only install version `v0.1.0`. This is a very outdated version given that the current release is `v1.0.0`

**Describe the solution you'd like in detail**
I'd like to see an update to the helm chart so that the helm chart tracks the latest release.

Saberos:
The [index.yaml](https://github.com/kubernetes-sigs/aws-efs-csi-driver/blob/gh-pages/index.yaml) used by Helm has not yet been updated to point to support the v1.0.0 release.

wongma7:
We are aware, but there is some weird issue with the digest (which I understand is not really used by helm for verification anyway) and I don't have permission to push a new index: https://github.com/kubernetes-sigs/aws-efs-csi-driver/pull/223 a

wongma7:
By the way, 0.1.0 refers to the version of the helm chart, not the `appVersion` of the CSI driver. Can you double check the version of the driver that was installed by inspecting the aws-efs-csi-driver DaemonSet?

albertmichaelj:
I realized after I opened this issue that the helm chart version did not track with the `appVersion`, but it was an outdated version because it wouldn't let me use TLS encryption. It gave me an error that TLS encryption was not supported on the version. Unfortunately, I don't have it installed anymore, and I am hesitant to put it on my production cluster again. I uninstalled the helm chart and then installed this from just applying the `yaml` files method of installation, and then pods started failing to start, I'm still not sure what that issue was. I've switched back to standard NFS pvs for now.

Saberos:
> By the way, 0.1.0 refers to the version of the helm chart, not the `appVersion` of the CSI driver. Can you double check the version of the driver that was installed by inspecting the aws-efs-csi-driver DaemonSet?

Can confirm that I'm getting v0.3.0 csi driver daemonset containers when using helm chart v0.1.0.

arvin6msd:
Same issue. helm update does not update 0.3 app version to 1.0. Can someone update the chart?

antonosmond:
[...]

PR #255, opened 2020-09-25
Bump chart version as appVersion was updated to 1.0.0

antonosmond:
This fixes https://github.com/kubernetes-sigs/aws-efs-csi-driver/issues/235

The helm chart was updated without bumping the chart version. A chart should never be updated like this as it will cause problems with the digest in the index and breaks versioning i.e. someone that expects version 0.1.0 to be consistent and static would have the chart changing underneath them if they ever redeploy the same chart version. To avoid this, changes to the chart require the chart version to be bumped, hence this PR.

### DIFF
--- a/helm/Chart.yaml
+++ b/helm/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: "1.0.0"
 name: aws-efs-csi-driver
 description: A Helm chart for AWS EFS CSI Driver
-version: 0.1.0
+version: 0.2.0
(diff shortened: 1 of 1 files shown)


chrisjohnson00:
> /lgtm

Worth a shot.
:+1: on this one!

wongma7:
/lgtm
Thanks, I am pretty ignorant about helm but everything said here makes sense.
https://github.com/kubernetes-sigs/aws-efs-csi-driver/blob/gh-pages/index.yaml needs to be fixed as well

PR #258, opened 2020-10-06
Added jqmichael as approver

jqmichael:

Added jqmichael as approver

**What testing is done?**


wongma7:
You can't have your @jqmichael in the commit message I guess because it would send you a notification every time someone makes the commit in their fork or something like that.

PR #266, opened 2020-10-16
Add E2E test flag for EFS file system ID

chrishenzie:

Bug fix.

The E2E test suite documentation suggests users supply a `-file-system-id` flag, but no flag exists for this.

**What testing is done?** 

Running E2E test suite against cluster with pre-provisioned EFS file system.


wongma7:
/lgtm
/approve

that's odd, ty

Issue #100, opened 2019-10-30
Unable to mount two folders on same EFS volume


stefansedich:
/kind bug

**What happened?**

When attempting to mount two different folders on the same EFS volume using multiple PV and PVCs the pod fails to start showing a timeout binding the volumes:

```
Unable to mount volumes for pod "efs_default": timeout expired waiting for volumes to attach or mount for pod "default"/"efs". list of unmounted volumes=[efs-test2]. list of unattached volumes=[efs-test efs-test2 default-token-nffl4]
```

Binding works fine if I mount only one of the volumes, doesn't matter which one but as soon as there is >1 it fails.

**What you expected to happen?**

I would expect this to work as it did when I was previously using an NFS PV.

**How to reproduce it (as minimally and precisely as possible)?**

Use the following manifests, it will require an EFS volume with the `/efs-test` and `/efs-test2` folders created.
```
apiVersion: v1
kind: PersistentVolume
metadata:
  name: efs-test
spec:
  storageClassName: efs
  capacity:
    storage: 1Gi
  accessModes:
    - ReadWriteMany
  persistentVolumeReclaimPolicy: Delete
  mountOptions:
    - tls
  csi:
    driver: efs.csi.aws.com
    volumeHandle: fs-xxx
    volumeAttributes:
      path: "/efs-test"
---
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: efs-test
spec:
  accessModes:
    - ReadWriteMany
  resources:
    requests:
      storage: 1Gi
  storageClassName: efs
  volumeName: efs-test
---
apiVersion: v1
kind: PersistentVolume
metadata:
  name: efs-test2
spec:
  storageClassName: efs
  capacity:
    storage: 1Gi
  accessModes:
    - ReadWriteMany
  persistentVolumeReclaimPolicy: Delete
  mountOptions:
    - tls
  csi:
    driver: efs.csi.aws.com
    volumeHandle: fs-xxx
    volumeAttributes:
      path: "/efs-test2"
---
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: efs-test2
spec:
  accessModes:
    - ReadWriteMany
  resources:
    requests:
      storage: 1Gi
  storageClassName: efs
  volumeName: efs-test2
---
apiVersion: v1
kind: Pod
metadata:
  name: efs
spec:
  containers:
    - name: ubuntu
      image: ubuntu:latest
[...]

PR #272, opened 2020-10-31
Don't fail making mount dirs if they already exist

chrishenzie:
Bug fix.

When the `should mount different paths on same volume on same node` e2e test is re-ran more than once against the same pre-provisioned EFS file system, the test fails because the first pod in the test repeatedly crashloops. The pod crashloops because `mkdir` fails if the directory already exists. This fixes the issue by ignoring if the directory already exists.

A followup to this might include adding a cleanup step that deletes the files / directories placed on the EFS file system in these tests. Thoughts on this?

**What testing is done?**
```sh
# Pre-provision an EFS file system

# Make the binary.
make test-e2e-bin

# Run test against pre-provisioned EFS file system.
bin/test-e2e \
	-kubeconfig=$HOME/.kube/config \
	-ginkgo.focus="should mount different paths on same volume on same node" \
	-ginkgo.skip="\[Disruptive\]" \
	-ginkgo.dryRun \
	-file-system-id fs-0c7a3009 \
	-region us-west-2

# Run it a second time and observe the test not fail.
bin/test-e2e \
	-kubeconfig=$HOME/.kube/config \
	-ginkgo.focus="should mount different paths on same volume on same node" \
	-ginkgo.skip="\[Disruptive\]" \
	-ginkgo.dryRun \
	-file-system-id fs-0c7a3009 \
	-region us-west-2
```

@msau42 @wongma7

### DIFF
--- a/test/e2e/e2e.go
+++ b/test/e2e/e2e.go
@@ -204,7 +204,7 @@ var _ = ginkgo.Describe("[efs-csi] EFS CSI", func() {
 			}()
 
 			ginkgo.By(fmt.Sprintf("Creating pod to make subpaths /a and /b"))
-			pod := e2epod.MakePod(f.Namespace.Name, nil, []*v1.PersistentVolumeClaim{pvcRoot}, false, "mkdir /mnt/volume1/a && mkdir /mnt/volume1/b")
+			pod := e2epod.MakePod(f.Namespace.Name, nil, []*v1.PersistentVolumeClaim{pvcRoot}, false, "mkdir -p /mnt/volume1/a && mkdir -p /mnt/volume1/b")
 			pod, err = f.ClientSet.CoreV1().Pods(f.Namespace.Name).Create(context.TODO(), pod, metav1.CreateOptions{})
 			framework.ExpectNoError(err, "creating pod")
 			framework.ExpectNoError(e2epod.WaitForPodSuccessInNamespace(f.ClientSet, pod.Name, f.Namespace.Name), "waiting for pod success")


test/e2e/e2e.go
```diff
@@ -189,7 +189,7 @@ var _ = ginkgo.Describe("[efs-csi] EFS CSI", func() {
 			defer func() { _ = f.ClientSet.CoreV1().PersistentVolumes().Delete(pvRoot.Name, &metav1.DeleteOptions{}) }()
 
 			ginkgo.By(fmt.Sprintf("Creating pod to make subpaths /a and /b"))
-			pod := e2epod.MakePod(f.Namespace.Name, nil, []*v1.PersistentVolumeClaim{pvcRoot}, false, "mkdir /mnt/volume1/a && mkdir /mnt/volume1/b")
+			pod := e2epod.MakePod(f.Namespace.Name, nil, []*v1.PersistentVolumeClaim{pvcRoot}, false, "mkdir -p /mnt/volume1/a && mkdir -p /mnt/volume1/b")
```
msau42, line 192: Maybe the test should write to a randomly generated directory name so that we can validate in the test that read/write works?

msau42: Also how similiar is this test case to https://github.com/kubernetes/kubernetes/blob/5ead4974e0133286b801915b2d8de24bb64e36aa/test/e2e/storage/testsuites/subpath.go#L291?

msau42: Oh I guess this test case is creating two separate pvs for each subpath

wongma7: Either way I think the test should clean up after itself, i.e. not have weird sideffects like this. I am fine with current form of PR since this part is basically just setting up for the "real " test and if write is failing here then another test should cover that.

wongma7: this test is testing the workaround to issue https://github.com/kubernetes-sigs/aws-efs-csi-driver/issues/100, where due to how kubernetes treats volumeHandle , two volumemounts w/ save volumeHandle won't be mounted, so we create pvs with different volumehandles like fs:/a and fs:/b (probably I should have linked it in comment of test)

PR #273, opened 2020-10-31
Add flag for label selectors for EFS driver in e2e test

chrishenzie:
New feature.

The `should continue reading/writing without hanging after the driver pod is restarted` e2e test assumes the EFS driver running on each node is labeled with `app=efs-csi-node`. This PR adds a flag to the e2e test that allows users to provide custom label selectors for the EFS driver running in their cluster.

Another followup to this may be adding a flag to specify the namespace this pod may run in (if it's not `kube-system`). Thoughts on this?

**What testing is done?**
```sh
# Make the binary.
make test-e2e-bin

# Show labels on existing EFS pods.
kubectl -n kube-system describe ds aws-efs-csi-driver-node
Name:           aws-efs-csi-driver-node
...
Labels:         key1=value1
                key2=value2
...

# Run test against pre-provisioned EFS file system with these labels and observe test pass.
bin/test-e2e \
	-kubeconfig=$HOME/.kube/config \
	-ginkgo.focus="should continue reading/writing without hanging after the driver pod is restarted" \
	-ginkgo.skip="\[Disruptive\]" \
	-file-system-id fs-0c7a3009 \
	-efs-driver-label-selectors='key1=value1,key2=value2' \
	-region us-west-2

# Run test against pre-provisioned EFS file system without these labels and observe test fail with pod not found.
bin/test-e2e \
	-kubeconfig=$HOME/.kube/config \
	-ginkgo.focus="should continue reading/writing without hanging after the driver pod is restarted" \
	-ginkgo.skip="\[Disruptive\]" \
	-file-system-id fs-0c7a3009 \
	-region us-west-2
```

@msau42 @wongma7

### DIFF
--- a/test/e2e/e2e.go
+++ b/test/e2e/e2e.go
@@ -25,9 +25,11 @@ var (
 	// Parameters that are expected to be set by consumers of this package.
 	// If FileSystemId is not set, ClusterName and Region must be set so that a
 	// file system can be created
-	ClusterName  string
-	Region       string
-	FileSystemId string
+	ClusterName             string
+	Region                  string
+	FileSystemId            string
+	EfsDriverNamespace      string
+	EfsDriverLabelSelectors map[string]string
 
 	deleteFileSystem = false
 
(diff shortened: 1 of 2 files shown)


test/e2e/e2e.go
```diff
@@ -28,6 +28,11 @@ var (
 	ClusterName  string
 	Region       string
 	FileSystemId string
+	// Combined label selectors in the form of key1=value1,key2=value2.
+	// Supplied by users.
+	CombinedEfsDriverLabelSelectors string
```
wongma7, line 33: can unexport and move this to e2e_test.go since basically it is an intermediate value that is not useful until parsed

wongma7:
> Another followup to this may be adding a flag to specify the namespace this pod may run in (if it's not kube-system). Thoughts on this?

yes that would be nice, currently test assumes user has  just done the kustomize install in the README

wongma7:
/lgtm
/approve
thank you

PR #274, opened 2020-11-01
Introduce Controller with Create and Delete Volume capability. Adds support for Dynamic Provisioning via access points

### LINKED ISSUE #6: Implement dynamic provision support
# Option 1
User create PVC, its creation will trigger the driver to create a NEW EFS filesystem. The storageclass will be defined as follows:

## StorageClass
```
kind: StorageClass
apiVersion: storage.k8s.io/v1
metadata:
  name: efs-sc
provisioner: efs.csi.aws.com
parameters:
  encrypted: "true",
  kmsKeyId: {keyId},
  performanceMode: "generalPurpose",
  provisionedThroughputInMibps: "512",
  throughputMode: "provisioned"
```

## Open question
* How to configure the driver to create EFS mount targets for each availability with user defines subnets and security groups?

# Option 2 (Preferred)
User creates PVC, its creation will trigger the driver to provision a subdirectory within an existing EFS filesystem as the volume. The storageclass will be defined as follows:

```
kind: StorageClass
apiVersion: storage.k8s.io/v1
metadata:
  name: efs-sc
provisioner: efs.csi.aws.com
parameters:
  filesystemId: {EFS filesystem ID}
```

## Provisioing EFS filesystem
Since we model the storageclass as the configuration for creating new directories inside existing EFS filesystem, user will need a different way to provision EFS system. This can be achieved by create a new service operator for EFS using [service operator framework](https://github.com/awslabs/aws-service-operator). User will need to create a custom resource that defines a EFS filesystem, and the operator will create the EFS filesystem by watching the custom resource.

For various configurations supported by EFS, eg performanceMode, throughputMode, kmsKeyId, etc, they will be a field within the EFS CRD.

Ref:
* https://docs.aws.amazon.com/sdk-for-go/api/service/efs/#CreateFileSystemInput
* https://docs.aws.amazon.com/sdk-for-go/api/service/efs/#CreateMountTargetInput

### LINKED ISSUE #269: Auto creation of Accesspoints

When using the current driver in a multi-tenant cluster, tenants require access to resources in the AWS account in order to create EFS Accesspoints. 
The 'old' efs-provisioner would create a subpath for each new pv, while the CSI driver does not.

/feature

Being able to create a new EFS Accesspoint for new pvs.

Stick to efs-provisioner for now.

kbasv:
New Feature

* Adds support for Dynamic Provisioning via access points. Provisions a new volume by creating an access point.

* Fix for the following issues:
 Fixes #6 
 Fixes #269 
 https://github.com/aws/containers-roadmap/issues/891

* Sample Storage class:
```
kind: StorageClass
apiVersion: storage.k8s.io/v1
metadata:
  name: efs-sc
provisioner: efs.csi.aws.com
parameters:
  provisioningMode: efs-ap
  fileSystemId: fs-01234567
  gidRangeStart: "1000"  //Optional
  gidRangeEnd: "2000"   //Optional
  directoryPerms: "777"
  basePath: "/data"  //Optional
```

* Adds an opt-in flag to delete access virtual point directory and it's contents. By default, deleting a access point will not delete its virtual root directory or its content.

**What testing is done?** 

* Tested Manually on my EKS cluster.
* E2E integration test for dynamic provisioning


pkg/driver/controller.go
```diff
@@ -0,0 +1,311 @@
...
+	}
+
+	// Check if file system exists. Describe FS handles appropriate error codes
+	if _, err = d.cloud.DescribeFileSystem(ctx, accessPointsOptions.FileSystemId); err != nil {
+		if err == cloud.ErrNotFound {
+			return nil, status.Errorf(codes.InvalidArgument, "File System does not exist: %v", err)
+		}
+		return nil, status.Errorf(codes.Internal, "Failed to fetch File System info: %v", err)
+	}
+	gid, err := d.gidAllocator.getNextGid(accessPointsOptions.FileSystemId, gidMin, gidMax)
```
wongma7, line 158: Say the directory permission is 770 or 740, how will the user using the PV discover the GID and ensure their Pod has it? The old provisioner put a special GID annotation on all PVs https://github.com/kubernetes-retired/external-storage/blob/master/aws/efs/cmd/efs-provisioner/efs-provisioner.go#L190. This PV annotation is basically a field in kubernetes that means: add this group to the consuming Pod's supplemental groups. (Consuming pod meaning the Pod in relation Pod->PVC->PV). It doesn't seem possible in CSI to make this annotation show up on the final PV without updating https://github.com/kubernetes-csi/external-provisioner to do it somehow. I am not sure the usefulness of assigning a random GID without also setting the spcial annotation

kbasv: I ran some tests to see how K8s and the pods behave with different directory permissions. By default, it looks like pods get root(owner) permissions and pods will be able to access the PV.

The random GIDs help with restricting access on EFS. Each GID will be used to create a root directory for access points and the user using a PV will not be able to access other root directories behind different access points which serve other PVs. Also GIDs here serve as unique `ID` for an access point root directory. This prevents access points from re-using existing root directories and prevents data from being due to overwritten/corruption.

wongma7: The UID/GID a pod gets is dependent on the image, PSP enforced on the pod, and pod's security context. There is no guarantee that a pod will get root, this assumption is wrong.

I do not think there is any risk of users accessing directories other than what is under their EFS access point nor do I think GID is a good mechanism for controlling access. Besides, if you are going with the assumption that all pods run as root, and efs does not squash root, then it doesn't matter what GID the directories are. Anyway, all directories provisioned by the driver should be at the same level in the tree, there should never be one directory a child of another.

For a unique ID, I would prefer something truly unique and more useful like PVC UID

wongma7: If I am reading https://docs.aws.amazon.com/efs/latest/ug/accessing-fs-nfs-permissions.html#accessing-fs-nfs-permissions-access-points correctly, the access point will be readable no matter what the uid/gid/psp/etc. is since "When user enforcement is enabled, Amazon EFS replaces the NFS client's user and group IDs with the identity configured on the access point for all file system operations."

Can you test that if the Pod does not have root uid (using one of the ways i mentioned, securitycontext is easiest) that it can still read/write the access point no matter what the access point's uid/gid is

pkg/driver/node.go
```diff
@@ -213,6 +222,15 @@ func (d *Driver) NodeUnpublishVolume(ctx context.Context, req *csi.NodeUnpublish
 		return &csi.NodeUnpublishVolumeResponse{}, nil
 	}
 
+	deleteCount, err := decrementMapCounter(volumeContentDeleteTracker, req.GetVolumeId())
+	if err != nil {
+		klog.Errorf("Failed to check if volume content should be deleted: %v", err)
+	}
+
+	if deleteCount == 0 {
+		deleteDirectoryContent(req.GetTargetPath())
```
wongma7, line 231: why is this needed, seems a bit dangerous!

wongma7: Say I NodePublish the same volume 3x to different paths. driver restarts. I NodePublish 1 more time. I NodeUnpublish it. Won't it delete everything even though there are still 3 other consumers?

pkg/driver/controller.go
kbasv: To provide additional context, I configured the pod's security context to runAsUser and runAsGroup to be different from that of the GID of the root directory with permissions 700, the pod was able to access the PV. May be it the UID/GID is a determined by a combination of all image,  PSP and security context? - I'll do additional research on this.

Posix UID,GID and directory perms are required parameters to create an [access point root directory](https://docs.aws.amazon.com/efs/latest/ug/efs-access-points.html#enforce-root-directory-access-point).
The driver ensures all root directories created for access points will be under the same base path and one will not be a child of another.

In any case, your concern is valid and currently there doesn't seem to be a way to propagate the GIDs to pods from CSI driver.

Unique ID is a combination of a prefix(efs-csi-ap), GID & a UUID. PVC UID is used as the creation token for access point. It can also be used as the directory name, however, felt the former would be a good choice to identify the directories created for access points.

pkg/driver/node.go
kbasv: There is currently no way to delete a root directory created by access points and the driver will delete only the access point inside `DeleteVolume`.
DeleteContent option will allow users/administrators to specify if they want to delete the data behind the root directory when they want to delete a volume. If the option is set, the driver will delete the data prior to unmount inside NodeUnpublish

This counter specifically tracks the if an access point (volume Handle [fs_id]::[ap_id]) is mounted in more than one volume, lets say Volume A and B. If A is deleted and B is still in use, driver doesn't delete the data till B is deleted.

kbasv: > Say I NodePublish the same volume 3x to different paths. driver restarts. I NodePublish 1 more time. I NodeUnpublish it. Won't it delete everything even though there are still 3 other consumers?

Ah good catch on this one! I did not think of the case where the driver restarts

chrishenzie (changes requested):
Thank you for your work on this! One other nit: Can you run [`goimports`](https://godoc.org/golang.org/x/tools/cmd/goimports) on the files you modified here?
```
go get golang.org/x/tools/cmd/goimports
find . -name \*.go -not -path './vendor/*' -exec goimports -w {} \;
```

Makefile
```diff
@@ -49,9 +49,8 @@ verify:
 
 .PHONY: test
 test:
-	go test -v -race $$(go list ./pkg/... | grep -v /driver)
-	# TODO stop skipping controller tests when controller is implemented
-	go test -v -race ./pkg/driver/... -ginkgo.skip='\[Controller.Server\]'
+	go test -v -race $$(go list ./pkg/...)
+	go test -v -race ./pkg/driver/...
```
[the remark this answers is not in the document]
kbasv: Can't remember why I did this, you are right, I'll update this line.

pkg/cloud/cloud.go
```diff
@@ -17,19 +17,69 @@ limitations under the License.
...
+}
+
+type AccessPoint struct {
+	AccessPointId string
+	FileSystemId  string
+	CapacityGiB   int64
+}
+
+type AccessPointOptions struct {
+	CapacityGiB    int64
```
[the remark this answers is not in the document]
kbasv: Added this to satisfy CSI SanityTests. I'll add a comment that it is used only for testing purposes. You are right, EFS file systems do not have a capacity

pkg/cloud/cloud.go
```diff
@@ -43,11 +93,131 @@ func NewCloud() (Cloud, error) {
...
+	deleteAccessPointInput := &efs.DeleteAccessPointInput{AccessPointId: &accessPointId}
+	_, err = c.efs.DeleteAccessPointWithContext(ctx, deleteAccessPointInput)
+	if err != nil {
+		if isAccessDenied(err) {
+			return ErrAccessDenied
+		}
+		if isAccessPointNotFound(err) {
+			return ErrNotFound
+		}
+		return fmt.Errorf("Failed to delete access point: %v", accessPointId)
```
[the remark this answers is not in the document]
kbasv: Sure, I'll add the error to the log.

pkg/cloud/cloud.go
```diff
@@ -43,11 +93,131 @@ func NewCloud() (Cloud, error) {
...
+			return nil, ErrAccessDenied
+		}
+		if isAccessPointNotFound(err) {
+			return nil, ErrNotFound
+		}
+		return nil, fmt.Errorf("describe Access Point failed: %v", err)
+	}
+
+	return &AccessPoint{
+		AccessPointId: *res.AccessPoints[0].AccessPointId,
```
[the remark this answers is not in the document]
kbasv: I do not think there will ever be two access points with the same Id. However, in the event something unexpected happened, this has a potential to fail, I'll add some error handling on the size of the slice.

pkg/cloud/cloud_test.go
```diff
@@ -0,0 +1,370 @@
...
+				ctx := context.Background()
+				mockEfs.EXPECT().DeleteAccessPointWithContext(gomock.Eq(ctx), gomock.Any()).Return(output, nil)
+				err := c.DeleteAccessPoint(ctx, accessPointId)
+				if err != nil {
+					t.Fatalf("Delete Access Point failed: %v", err)
+				}
+
+				mockctl.Finish()
+			},
+		},
```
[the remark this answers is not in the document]
kbasv: Yeah, adding a test case for that.

pkg/cloud/cloud_test.go
```diff
@@ -0,0 +1,370 @@
...
+			name: "Fail: Access Point Not Found",
+			testFunc: func(t *testing.T) {
+				mockctl := gomock.NewController(t)
+				mockEfs := mocks.NewMockEfs(mockctl)
+				c := &cloud{efs: mockEfs}
+
+				ctx := context.Background()
+				mockEfs.EXPECT().DeleteAccessPointWithContext(gomock.Eq(ctx), gomock.Any()).Return(nil, awserr.New(efs.ErrCodeAccessPointNotFound, "Access Point not found", errors.New("DeleteAccessPointWithContext failed")))
+				err := c.DeleteAccessPoint(ctx, accessPointId)
+				if err == nil {
```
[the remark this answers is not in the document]
kbasv: Sure

pkg/cloud/fakes.go
```diff
@@ -0,0 +1,78 @@
...
+
+func (c *FakeCloudProvider) DescribeFileSystem(ctx context.Context, fileSystemId string) (fileSystem *FileSystem, err error) {
+	if fs, ok := c.fileSystems[fileSystemId]; ok {
+		return fs, nil
+	}
+
+	fs := &FileSystem{
+		FileSystemId: fileSystemId,
+	}
+	c.fileSystems[fileSystemId] = fs
```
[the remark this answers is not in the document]
kbasv: I did think of updating the `fileSystems` Map inside `CreateAccessPoint()`. However, the `CreateVolume()` code first calls `DescribeFileSystem()` and then `CreateAccessPoint()`. If I didn't do this, then DescribeFileSystem() will fail which will cause the some of the sanity tests around `CreateVolume` to fail.

pkg/driver/controller.go
```diff
@@ -0,0 +1,326 @@
...
+)
+
+func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest) (*csi.CreateVolumeResponse, error) {
+	klog.V(4).Infof("CreateVolume: called with args %+v", *req)
+	volName := req.GetName()
+	if volName == "" {
+		return nil, status.Error(codes.InvalidArgument, "Volume name not provided")
+	}
+
+	volSize := req.GetCapacityRange().GetRequiredBytes()
```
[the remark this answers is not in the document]
kbasv: Agreed, I'll add a comment

pkg/driver/controller.go
```diff
@@ -0,0 +1,326 @@
...
+		volContext       = make(map[string]string)
+		err              error
+	)
+
+	//Parse parameters
+	volumeParams := req.GetParameters()
+	if value, ok := volumeParams[ProvisioningMode]; ok {
+		provisioningMode = value
+		//TODO: Add FS provisioning mode check when implemented
+		if provisioningMode != "efs-ap" {
```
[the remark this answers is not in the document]
kbasv: Yeah, will make it a constant

pkg/driver/efs_watch_dog.go
```diff
@@ -233,7 +233,7 @@ func (w *execWatchdog) runLoop(stopCh <-chan struct{}) {
 	for {
 		select {
 		case <-stopCh:
-			klog.Info("stopping...")
+			//klog.Info("stopping...")
```
kbasv, line 236: This line spams the test out and makes the build logs unreadable. @wongma7 LMK if you want me to uncomment before merge

pkg/driver/node.go
```diff
@@ -178,11 +186,12 @@ func (d *Driver) NodePublishVolume(ctx context.Context, req *csi.NodePublishVolu
...
-			volumeIdCounter[req.GetVolumeId()] = value + 1
-		} else {
-			volumeIdCounter[req.GetVolumeId()] = 1
-		}
+		incrementMapCounter(volumeIdCounter, req.GetVolumeId())
+	}
+
+	//Increment Delete content counter
+	if volumeDeleteContent {
+		incrementMapCounter(volumeContentDeleteTracker, req.GetVolumeId())
```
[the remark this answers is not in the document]
kbasv: I think all the requests to CSI driver is sequential/handled by a single thread. Let me know if this isn't the case, I'll add a lock around this

pkg/cloud/fakes.go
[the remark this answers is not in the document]
chrishenzie: Thank you for the clarification, this may be worth leaving as a comment for context.

pkg/driver/controller.go
```diff
@@ -0,0 +1,326 @@
...
+		if err = d.cloud.DeleteAccessPoint(ctx, accessPointId); err != nil {
+			if err == cloud.ErrNotFound {
+				klog.V(5).Infof("DeleteVolume: Access Point not found, returning success")
+				return &csi.DeleteVolumeResponse{}, nil
+			}
+			return nil, status.Errorf(codes.Internal, "Failed to Delete volume %v: %v", volId, err)
+		}
+	} else {
+		return nil, status.Errorf(codes.Internal, "Failed to find access point for volume: %v", volId)
+	}
```
jqmichael, line 227: Is this an internal failure (assuming `codes.Internal` means it's 500x)?

pkg/driver/node.go
```diff
@@ -82,6 +86,10 @@ func (d *Driver) NodePublishVolume(ctx context.Context, req *csi.NodePublishVolu
 				return nil, status.Errorf(codes.InvalidArgument, "Volume context property %q must be an absolute path", k)
 			}
 			subpath = filepath.Join(subpath, v)
+		case "storage.kubernetes.io/csiprovisioneridentity":
+			continue
```
jqmichael, line 90: Curious why this would be set in volume context.

pkg/driver/node.go
jqmichael: Yea, I'm more comfortable of doing this *only if* `volumeDeleteContent` is set explicitly, rather than relying on a map whose initial state would delete content. 

I'm surprised NodeUnpublishVolume request doesn't expose VolumeContext. But since we cache the result in `volumeDeleteContent`, we should explicitly check if it is still set to true before deletion.

jqmichael (commented):
I assume you are working on e2e tests at least for the happy path?

deploy/kubernetes/base/clusterrole-provisioner.yaml
```diff
@@ -0,0 +1,35 @@
...
+    verbs: ["get", "list"]
+  - apiGroups: ["storage.k8s.io"]
+    resources: ["csinodes"]
+    verbs: ["get", "list", "watch"]
+  - apiGroups: [""]
+    resources: ["nodes"]
+    verbs: ["get", "list", "watch"]
+  - apiGroups: ["coordination.k8s.io"]
+    resources: ["leases"]
+    verbs: ["get", "watch", "list", "delete", "update", "create"]
```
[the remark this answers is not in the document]
kbasv: Yes, I copied it from an upstream template

deploy/kubernetes/base/controller.yaml
```diff
@@ -0,0 +1,77 @@
...
+      labels:
+        app: efs-csi-controller
+        app.kubernetes.io/name: aws-efs-csi-driver
+    spec:
+      nodeSelector:
+        kubernetes.io/os: linux
+      serviceAccountName: efs-csi-controller-sa
+      priorityClassName: system-cluster-critical
+      tolerations:
+        - operator: Exists
```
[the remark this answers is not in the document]
kbasv: I copied it from an upstream template, I'm not too sure of this one.

deploy/kubernetes/base/controller.yaml
```diff
@@ -0,0 +1,77 @@
...
+          livenessProbe:
+            httpGet:
+              path: /healthz
+              port: healthz
+            initialDelaySeconds: 10
+            timeoutSeconds: 3
+            periodSeconds: 10
+            failureThreshold: 5
+        - name: csi-provisioner
+          image: quay.io/k8scsi/csi-provisioner:v1.6.0
```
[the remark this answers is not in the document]
kbasv: I think this was the latest when I started working on it. I'll update to use the latest image.

pkg/driver/controller.go
kbasv: This will be an internal error from AWS, so I guess `codes.Internal`?
Or should we encapsulate all unexpected errors under `codes.Unknown`?

pkg/driver/node.go
kbasv: I'm not sure how to get rid of this. Kubernetes sets this for all dynamically provisioned volumes

pkg/driver/node.go
kbasv: Thinking about this, caching this option is a bad idea. I'll get rid of this and provide users another way to retain their data.

pkg/cloud/cloud.go
```diff
@@ -17,19 +17,69 @@ limitations under the License.
...
 	"github.com/aws/aws-sdk-go/aws/ec2metadata"
+	"github.com/aws/aws-sdk-go/aws/request"
 	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/efs"
+	"k8s.io/klog"
+)
+
+const (
+	// For testing purpose only
+	DefaultVolSize        = 5368709120
```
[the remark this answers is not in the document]
kbasv: Updated the csi-test to a newer version, this is no longer required, I'll remove it

pkg/driver/efs_watch_dog.go
wongma7: no let's remove it, I noticed it in a previous PR too. I'll create an issue to track it, it oculd be a real bug although tests  pass despite it

pkg/driver/controller.go
jqmichael: Why is failing to find access point a 500x not 400x?

vendor/github.com/kubernetes-csi/csi-test/pkg/sanity/controller.go
```diff
@@ -182,199 +171,10 @@ var _ = DescribeSanity("Controller Service [Controller Server]", func(sc *Sanity
...
-					},
-				},
-				Secrets: sc.Secrets.CreateVolumeSecret,
-			}
-			vol, err := c.CreateVolume(context.Background(), req)
-			Expect(err).NotTo(HaveOccurred())
-			Expect(vol).NotTo(BeNil())
-			Expect(vol.Volume).NotTo(BeNil())
-			existing_vols[vol.Volume.VolumeId] = true
+		// TODO: Add test to test for tokens
```
jqmichael, line 174: Was the old test failing with the new change? Should we concern?

pkg/driver/controller.go
```diff
@@ -0,0 +1,359 @@
...
+			target := TempMountPathPrefix + "/" + accessPointId
+			if err := d.mounter.MakeDir(target); err != nil {
+				return nil, status.Errorf(codes.Internal, "Could not create dir %q: %v", target, err)
+			}
+			if err := d.mounter.Mount(fileSystemId, target, "efs", mountOptions); err != nil {
+				os.Remove(target)
+				return nil, status.Errorf(codes.Internal, "Could not mount %q at %q: %v", fileSystemId, target, err)
+			}
+			os.RemoveAll(target + accessPoint.AccessPointRootDir)
+			err = d.mounter.Unmount(target)
```
jqmichael, line 240: This is saying on an arbitrary node (where the controller pod is assigned with), mount the EFS volume, delete the content, and unmount after. 

I think this is OK but would like @wongma7 to double check.

vendor/github.com/kubernetes-csi/csi-test/pkg/sanity/controller.go
jqmichael: Sorry, missed it was in the vendor package.

pkg/driver/controller.go
kbasv: Sorry, I mistook this to be inside the `if` loop above. I'll update this to `codes.NotFound`

pkg/driver/controller.go
```diff
@@ -0,0 +1,359 @@
...
+const (
+	AccessPointMode     = "efs-ap"
+	FsId                = "fileSystemId"
+	GidMin              = "gidRangeStart"
+	GidMax              = "gidRangeEnd"
+	DirectoryPerms      = "directoryPerms"
+	BasePath            = "basePath"
+	ProvisioningMode    = "provisioningMode"
+	DefaultGidMin       = 50000
+	DefaultGidMax       = 7000000
```
wongma7, line 42: isn't the max 65534 ? (depends on Os maybe?)

Any docs for the new parameters or example storageclass? Looksl ike ProvisioningMode is required. As for GID/direcotryPerms, I guess most will leave the GID parameters default given that it appears the GID has no effect on permissions for access points within the cluster as we discussed below (any Pod that mounts the access point will have access no matter their GID).

pkg/driver/controller.go
wongma7: It looks OK to me as long as the behavior is behind a flag and well documented.

deploy/kubernetes/overlays/dev/master_image.yaml
```diff
@@ -3,10 +3,27 @@ apiVersion: apps/v1
...
   namespace: kube-system
+  labels:
+    app.kubernetes.io/name: aws-efs-csi-driver
 spec:
   template:
     spec:
       containers:
         - name: efs-plugin
-          image: amazon/aws-efs-csi-driver:master
+          image: amazon/aws-efs-csi-driver:latest
```
wongma7, line 13: this change looks unintentional

pkg/cloud/cloud.go
```diff
@@ -17,19 +17,74 @@ limitations under the License.
...
+	// Capacity is used for testing purpose only.
+	// EFS does not consider capacity while provisioning new file systems or access points
+	// Capacity is used to satisfy this test: https://github.com/kubernetes-csi/csi-test/blob/v3.1.1/pkg/sanity/controller.go#L559
+	CapacityGiB    int64
+	FileSystemId   string
+	Uid            int64
+	Gid            int64
+	DirectoryPerms string
+	DirectoryPath  string
+	DeleteContent  bool
```
wongma7, line 66: this is unused

pkg/driver/controller.go
```diff
@@ -0,0 +1,359 @@
...
+		Volume: &csi.Volume{
+			CapacityBytes: volSize,
+			VolumeId:      accessPointsOptions.FileSystemId + "::" + accessPointId.AccessPointId,
+			VolumeContext: map[string]string{},
+		},
+	}, nil
+}
+
+func (d *Driver) DeleteVolume(ctx context.Context, req *csi.DeleteVolumeRequest) (*csi.DeleteVolumeResponse, error) {
+	klog.V(4).Infof("DeleteVolume: called with args %+v", *req)
```
wongma7, line 197: the gid is not released in this function

pkg/driver/controller.go
```diff
@@ -0,0 +1,359 @@
...
+			return nil, status.Errorf(codes.Unauthenticated, "Access Denied. Please ensure you have the right AWS permissions: %v", err)
+		}
+		if err == cloud.ErrNotFound {
+			return nil, status.Error(codes.NotFound, "Volume not found")
+		}
+		return nil, status.Errorf(codes.Internal, "Could not get volume with ID %q: %v", volId, err)
+	}
+
+	if accessPointId != "" {
+		if _, err := d.cloud.DescribeAccessPoint(ctx, accessPointId); err != nil {
```
wongma7, line 301: making cloud calls here is not necessary, this function should just be concerned with volume capabilities like SINGLE_NODE_WRITER, spec here https://github.com/container-storage-interface/spec/blob/master/spec.md#validatevolumecapabilities

wongma7 (commented):
some minor comments, and missing e2e tests and docs but otherwise lgtm. It's not that straightforward how to enable the dynamic pv e2e tests but check out https://github.com/kubernetes-sigs/aws-efs-csi-driver/blob/master/test/e2e/e2e.go#L53 . It will be easier to debug if the tests are enabled in the same PR as the code.

pkg/driver/controller.go
wongma7: IMO deleteAccessPointRootDir should default true to match kubernetes expectations where Delete means data is gone.

pkg/driver/controller.go
kbasv: Added this number based on what EFS supports. More info the default max supported by EFS here - https://docs.aws.amazon.com/efs/latest/ug/create-access-point.html

deploy/kubernetes/overlays/dev/master_image.yaml
kbasv: Oops, will revert this.

pkg/cloud/cloud.go
kbasv: Will get rid of this

pkg/driver/controller.go
kbasv: This is by design, idea is to not re-allocate a GID once it is used

pkg/driver/controller.go
kbasv: ah thanks for pointing this out, I'll update the function accordingly

kbasv:
Looks like the test account used for running integ tests is having problems. I ran the test separately on my account and the integ tests succeeded - 

`• [SLOW TEST:140.165 seconds]
[efs-csi] EFS CSI
/home/ec2-user/go/aws-efs-csi-driver/test/e2e/e2e.go:210
  [Driver: efs.csi.aws.com]
  /home/ec2-user/go/aws-efs-csi-driver/test/e2e/e2e.go:212
    [Testpattern: Pre-provisioned PV (block volmode)] volumeMode
    /home/ec2-user/go/pkg/mod/k8s.io/kubernetes@v1.18.10/test/e2e/storage/testsuites/base.go:94
      should fail to create pod by failing to mount volume [Slow]
      /home/ec2-user/go/pkg/mod/k8s.io/kubernetes@v1.18.10/test/e2e/storage/testsuites/volumemode.go:194
------------------------------
STEP: Deleting EFS filesystem "fs-25bebb20"
STEP: Deleted EFS filesystem "fs-25bebb20"

JUnit report was created: /home/ec2-user/go/aws-efs-csi-driver/test/e2e/report-dir/junit_01.xml

Ran 34 of 158 Specs in 1794.669 seconds
SUCCESS! -- 34 Passed | 0 Failed | 0 Pending | 124 Skipped`

kbasv:
@jqmichael @wongma7 added integ tests, please review

deploy/kubernetes/base/controller.yaml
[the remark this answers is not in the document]
wongma7: "An empty key with operator Exists matches all keys, values and effects which means this will tolerate everything."

https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/

pkg/driver/controller.go
```diff
@@ -0,0 +1,337 @@
...
+			mountOptions := []string{"tls"}
+			target := TempMountPathPrefix + "/" + accessPointId
+			if err := d.mounter.MakeDir(target); err != nil {
+				return nil, status.Errorf(codes.Internal, "Could not create dir %q: %v", target, err)
+			}
+			if err := d.mounter.Mount(fileSystemId, target, "efs", mountOptions); err != nil {
+				os.Remove(target)
+				return nil, status.Errorf(codes.Internal, "Could not mount %q at %q: %v", fileSystemId, target, err)
+			}
+			os.RemoveAll(target + accessPoint.AccessPointRootDir)
```
wongma7, line 239: Should this error be checked for fine to ignore? What happens if the dir has been deleted but access point still exists? Say if the controller gets killed at line 247

cmd/main.go
```diff
@@ -28,13 +28,15 @@ import (
...
-		volMetricsRefreshPeriod = flag.Float64("vol-metrics-refresh-period", 240, "Refresh period for volume metrics in minutes")
-		volMetricsFsRateLimit   = flag.Int("vol-metrics-fs-rate-limit", 5, "Volume metrics routines rate limiter per file system")
+		endpoint                 = flag.String("endpoint", "unix://tmp/csi.sock", "CSI Endpoint")
+		version                  = flag.Bool("version", false, "Print the version and exit")
+		efsUtilsCfgDirPath       = flag.String("efs-utils-config-dir-path", "/etc/amazon/efs/", "The path to efs-utils config directory")
+		efsUtilsStaticFilesPath  = flag.String("efs-utils-static-files-path", "/etc/amazon/efs-static-files/", "The path to efs-utils static files directory")
+		volMetricsOptIn          = flag.Bool("vol-metrics-opt-in", false, "Opt in to emit volume metrics")
+		volMetricsRefreshPeriod  = flag.Float64("vol-metrics-refresh-period", 240, "Refresh period for volume metrics in minutes")
+		volMetricsFsRateLimit    = flag.Int("vol-metrics-fs-rate-limit", 5, "Volume metrics routines rate limiter per file system")
+		deleteAccessPointRootDir = flag.Bool("delete-access-point-root-dir", false,
```
wongma7, line 38: what do we think about default true? I don't want users to be surprised by efs usage since usually delete means delete data. I considered default false before to "gate" the feature, but there is no point, the whole driver is beta.

pkg/driver/controller.go
```diff
@@ -0,0 +1,337 @@
...
+	volumeParams := req.GetParameters()
+	if value, ok := volumeParams[ProvisioningMode]; ok {
+		provisioningMode = value
+		//TODO: Add FS provisioning mode check when implemented
+		if provisioningMode != AccessPointMode {
+			errStr := "Provisioning mode " + provisioningMode + " is not supported. Only Access point provisioning: 'efs-ap' is supported"
+			return nil, status.Error(codes.InvalidArgument, errStr)
+		}
+	} else {
+		return nil, status.Errorf(codes.InvalidArgument, "Missing %v parameter", ProvisioningMode)
```
wongma7, line 92: would it be better to default it to efs-ap?

wongma7:
tests look to be passing.

I think a  test case for delete-access-point-root-dir will be needed in the future if we ever want to make changes to that code 

I am still looking for some kind of README update or example doc under https://github.com/kubernetes-sigs/aws-efs-csi-driver/tree/master/examples/kubernetes but can be done in follow up

/lgtm
/approve

pkg/driver/controller.go
wongma7: but if the driver gets restarted incidentally ( rescheduled, or updated) then the GID will be reused anyway, so this has no real security benefit.

Also this introduces an artifical limit on # of volumes that can be provisioned over the driver's lifetime which needs to be documented. Of course it may take a long time to hit 7million volumes if the user keeps the gid parameters at the default but it is a possibility that must be documented

deploy/kubernetes/base/controller.yaml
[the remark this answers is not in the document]
wongma7: i think ther has been discussion in the past that toleration all taints causes issues for users trying to drain nodes and such expecting NoSchedule to actually mean NoSchedule but let's just copy the template

wongma7:
I checked all existing comments have been resolved, did another pass, and left 3 more comments.

cmd/main.go
kbasv: I think we should keep this as false and not delete data by default and not delete data accidentally. This also ensures the administrator/user is aware that by turning on this feature, data is deleted.

pkg/driver/controller.go
kbasv: For now defaulting would be okay, however, we would have to bring this check back after we introduce 'efs-fs'. So lets keep it as is.

pkg/driver/controller.go
kbasv: Agreed, I'll add documentation for the possibility of running out of GIDs.

I couldn't think of a reliable way for now to recover the GIDs that have been used. I'll add a task to figure out and capture the used GIDs reliably in the event the driver is updated or restarted.

pkg/driver/controller.go
kbasv: It should be fine, the controller will proceed to delete access point once it is restarted - [Os.RemoveAll](https://golang.org/pkg/os/#RemoveAll) returns nil if the path doesn't exist.

kbasv:
I'll consolidate and add a readme file documenting dynamic provisioning and limitations of the controller and access points. I'll also provide examples for deploying dynamic provisioning in a follow up PR.

pkg/driver/controller.go
wongma7: ok, but we don't check if the err is nil.

pkg/driver/controller.go
wongma7: okay, as long as we provide an example storageclass and document that it is required then I dont mind

cmd/main.go
wongma7: well, everywhere else in kubernetes when PVC is deleted then the intention is that the data and backing volume are deleted. I am still convinced we need to default true but I will let @jqmichael tie break. nfs client provisioner had a similar issue where the default behaivor was "archiveOnDelete" and that conflicted with user expectations. If we commit to defaulting false now, it will be even harder to change it in the future.

cep21: I was also suprised by this, but neutral about the default.  It would be great to add strong documentation about how different this is than expectations with other drivers and the downsides (infinite accumulating data) in involves that will require out of bound maintenance.

test/e2e/e2e.go
```diff
@@ -102,6 +103,26 @@ func (e *efsDriver) GetPersistentVolumeSource(readOnly bool, fsType string, volu
 	return &pvSource, nil
 }
 
+func (e *efsDriver) GetDynamicProvisionStorageClass(config *testsuites.PerTestConfig, fsType string) *storagev1.StorageClass {
```
jqmichael, line 106: Is implementing `GetDynamicProvisionStorageClass` how to enable upstream dynamic provision tests?

jqmichael (commented):
Can we update example how to run dynamic provisioning with efs driver?

test/e2e/e2e.go
kbasv: Yeah, this allows the test suites to detect and run dynamic provision tests

kbasv:
@jqmichael I will be creating adding examples and documentation in a follow up PR

wangycc:
kubectl apply -f serviceaccount-csi-controller.yaml node.yaml csidriver.yaml controller.yaml clusterrole-provisioner.yaml clusterrolebinding-provisioner.yaml

After all apply, the controller component fails to start

controller logs:
```shell
I0106 03:02:23.058741       1 connection.go:186] GRPC error: rpc error: code = Unimplemented desc = unknown service csi.v1.Controller
F0106 03:02:23.058758       1 csi-provisioner.go:188] Error getting CSI driver capabilities: rpc error: code = Unimplemented desc = unknown service csi.v1.Controller
```

wongma7:
@wangycc what is the image in the controller.yaml? it should be `amazon/aws-efs-csi-driver:master` since no release contains this commit yet

wangycc:
@wongma7 amazon/aws-efs-csi-driver:latest,When do you plan to push  the controller image? We started to try.

Skaronator:
I'm a bit confused. This is not part of the 1.1 release, right? Will there be a new release any time soon?

wongma7:
@Skaronator right, this was merged into master branch but excluded from 1.1 release. I don't have ETA for when this will be released

Eternal21:
I'm following the link to this issue from this [document](https://aws.amazon.com/blogs/containers/running-stateful-workloads-with-amazon-eks-on-aws-fargate-using-amazon-efs/).  Is it now part of the official release?

rnishtala-sumo:
it would help to have an update on when support for dynamic provisioning will be released. Creating multiple accesspoints/PVs manually seems like a large overhead.

ncgisudo:
Is this actually available?  The feature was merged in Jan 2021 and it's now December 2022 (23 months later) and there's no information about it.

amej:
> Is this actually available? The feature was merged in Jan 2021 and it's now December 2022 (23 months later) and there's no information about it.

As a customer of this efs csi driver, I confirm that Dynamic provisioning via access point is available.  Delete volume capability does not fully work as mentioned in the README.

PR #279, opened 2020-11-12
Add psp rbac helm

quizy101:
New feature.

This was needed in my setup when our cluster required ServiceAccounts and a PodSecurityPolicy to allow access due to our restricted default PSP.

Added:
- Pod security policy which grants the appropriate permissions needed by the container such as root filesystem access, volumes, and host ports.
- ServiceAccount to tie the PSP and the Role.
- RBAC with a Role for accessing the PSP and a RoleBinding for the Role, PSP, and SA.

Configuration:
- I've left RBAC, PSP, and SA creation disabled for backwards compatibility. To use these kinds, you must enable them in values.yaml.

**What testing is done?** 
Currently using in our own cluster.

```bash
$ helm upgrade --install aws-efs-csi-driver -f values.yaml . --tiller-namespace kube-system --namespace kube-system
Release "aws-efs-csi-driver" has been upgraded.
LAST DEPLOYED: Thu Nov 12 15:23:32 2020
NAMESPACE: kube-system
STATUS: DEPLOYED

RESOURCES:
==> v1/DaemonSet
NAME          DESIRED  CURRENT  READY  UP-TO-DATE  AVAILABLE  NODE SELECTOR  AGE
efs-csi-node  3        3        3      3           3          <none>         3h19m

==> v1/Pod(related)
NAME                READY  STATUS   RESTARTS  AGE
efs-csi-node-579f7  3/3    Running  0         3h19m
efs-csi-node-dzvrp  3/3    Running  0         3h19m
efs-csi-node-shdc5  3/3    Running  0         3h19m

==> v1/Role
NAME                AGE
aws-efs-csi-driver  3h19m

==> v1/RoleBinding
NAME                AGE
aws-efs-csi-driver  3h19m

==> v1/ServiceAccount
NAME                SECRETS  AGE
aws-efs-csi-driver  1        3h19m

==> v1beta1/PodSecurityPolicy
NAME                PRIV  CAPS      SELINUX   RUNASUSER  FSGROUP   SUPGROUP  READONLYROOTFS  VOLUMES
aws-efs-csi-driver  true  RunAsAny  RunAsAny  RunAsAny   RunAsAny  false     hostPath,secret


NOTES:
To verify that aws-efs-csi-driver has started, run:

    kubectl get pod -n kube-system -l "app.kubernetes.io/name=aws-efs-csi-driver,app.kubernetes.io/instance=aws-efs-csi-driver"
```


flipstone42:
Hi all, what's the status of this PR?

helm/templates/serviceaccount.yaml
```diff
@@ -0,0 +1,8 @@
+{{- if .Values.serviceAccount.create -}}
+apiVersion: v1
+kind: ServiceAccount
+metadata:
```
Skaronator, line 4: namespace seems to be missing?

Skaronator (changes requested):

helm/templates/serviceaccount.yaml
Skaronator: Ah actually no. Helm just includes it if the value is empty.

nxf5025:
@quizy101 - Any objection to pulling in the changes from this PR? - https://github.com/kubernetes-sigs/aws-efs-csi-driver/pull/288

helm/templates/daemonset.yaml
```diff
@@ -28,36 +30,40 @@ spec:
...
-        {{- end }}
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
       hostNetwork: true
+      priorityClassName: {{ .Values.priorityClassName }}
+    {{- with .Values.affinity }}
+      affinity:
+        {{- toYaml . | nindent 8 }}
+    {{- end }}
+    {{- with .Values.tolerations }}
```
krmichel, line 50: I think it is probably the result of another merge but I don't think it was intentional to only include the DNS policy and config if tolerations are specified.

helm/templates/daemonset.yaml
```diff
@@ -16,8 +18,8 @@ spec:
         app: efs-csi-node
         app.kubernetes.io/name: {{ include "aws-efs-csi-driver.name" . }}
         app.kubernetes.io/instance: {{ .Release.Name }}
-      {{- if .Values.node.podAnnotations }}
-      annotations: {{ toYaml .Values.node.podAnnotations | nindent 8 }}
```
krmichel, line 20: If node.podAnnotations and node.tolerations aren't going to be used any longer they should be removed from the values.yaml.

helm/Chart.yaml
```diff
@@ -1,8 +1,8 @@
 apiVersion: v1
 appVersion: "1.0.0"
-name: aws-efs-csi-driver
 description: A Helm chart for AWS EFS CSI Driver
-version: 0.3.0
+name: aws-efs-csi-driver
+version: 0.3.1
```
krmichel, line 5: Since values that used to work in the templates are no longer being used the version should go to 0.4.0 if the project isn't ready to move to 1.0.0

helm/templates/_helpers.tpl
```diff
@@ -43,3 +43,14 @@ app.kubernetes.io/version: {{ .Chart.AppVersion | quote }}
 {{- end }}
 app.kubernetes.io/managed-by: {{ .Release.Service }}
 {{- end -}}
+
+{{/*
+Create the name of the service account to use
+*/}}
+{{- define "aws-efs-csi-driver.serviceAccountName" -}}
+{{- if .Values.serviceAccount.create -}}
```
krmichel, line 51: I believe this could just be `default (ternary .Values.serviceAccount.create (include "aws-efs-csi-driver.fullname" .) "default") .Values.serviceAccount.name` but I haven't tested it

helm/templates/daemonset.yaml
```diff
@@ -3,7 +3,9 @@ kind: DaemonSet
 apiVersion: apps/v1
 metadata:
   name: efs-csi-node
-  namespace: kube-system
+  namespace: {{ .Values.namespace }}
```
krmichel, line 6: I am not sure about this project but in general you either hard code a namespace because it is required to run in a specific namespace or you leave it out completely and allow the user to decide which namespace is goes in based on the namespace flag with the helm command.  I would generally not expect to have to set a value to control the namespace.  Maybe an owner would weigh in on whether to leave it hardcoded or remove it in favor of typical helm behavior.

@d-nishi @leakingtapan @justinsb @jsafrane @wongma7 @nckturner @jqmichael

helm/templates/daemonset.yaml
```diff
@@ -87,8 +93,11 @@ spec:
             timeoutSeconds: 3
             periodSeconds: 2
             failureThreshold: 5
+          resources:
+            {{- toYaml .Values.resources | nindent 12 }}
         - name: cs-driver-registrar
-          image: {{ printf "%s:%s" .Values.sidecars.nodeDriverRegistrarImage.repository .Values.sidecars.nodeDriverRegistrarImage.tag }}
```
krmichel, line 91: Is there a reason to move away from the printf functions for these sidecar images?  I would recommend keeping them.

helm/templates/daemonset.yaml
```diff
@@ -28,36 +30,40 @@ spec:
...
-        - name: {{ . }}
-      {{- end }}
+        {{- toYaml . | nindent 8 }}
     {{- end }}
+      serviceAccountName: {{ template "aws-efs-csi-driver.serviceAccountName" . }}
+      securityContext:
+        {{- toYaml .Values.podSecurityContext | nindent 8 }}
+      {{- with .Values.nodeSelector }}
       nodeSelector:
-        beta.kubernetes.io/os: linux
```
krmichel, line 38: Is it intentional to lose this selector?

helm/templates/daemonset.yaml
```diff
@@ -28,36 +30,40 @@ spec:
...
     {{- end }}
-    {{- if .Values.imagePullSecrets }}
+    {{- with .Values.imagePullSecrets }}
       imagePullSecrets:
-      {{- range .Values.imagePullSecrets }}
-        - name: {{ . }}
-      {{- end }}
+        {{- toYaml . | nindent 8 }}
     {{- end }}
+      serviceAccountName: {{ template "aws-efs-csi-driver.serviceAccountName" . }}
```
krmichel, line 37: There isn't anything technically wrong with using template here; however, template can't be used in pipelines so I tend to use include instead.

helm/Chart.yaml
quizy101: I agree on changing to 0.4.0, it is a lot of changes, I don't think I can make the call to 1.0.0.

helm/templates/_helpers.tpl
quizy101: Looks like ternary is a helm3 supported function but not in helm2. Since the chart api version is still set to v1, it would cause some issues to change the chart to be only helm3 supported. As it is right now, the chart is compatible with helm2 and helm3.

helm/templates/daemonset.yaml
quizy101: I had trouble with this, originally I had removed the hardcoded namespace because I felt that it would be up to the individual who installed the driver where they would like it to do. The problem is that once I was OK to test, the tests assume the chart is installed in kube-system, so without a default namespace set, it installs in the default namespace and the tests fail. Wasn't sure what to do so I went with a middle approach.

helm/templates/daemonset.yaml
quizy101: Great catch! Will remove. TY

helm/templates/daemonset.yaml
quizy101: No, but I would like to move it into values.yaml to give the user more choice. Will fix.

helm/templates/daemonset.yaml
quizy101: Whoa, yes big mistake here, fixing as well. Definitely a merge error.

helm/templates/_helpers.tpl
krmichel: Helm 2 is no longer supported, even for security patches.  Helm is strongly recommending migrating charts to helm 3 might be worth considering.  If the owners weigh in on the namespace part maybe they would have an opinion about switching to only supporting helm 3

helm/templates/daemonset.yaml
krmichel: I haven't looked at these tests much but I think at least this helm install would need to change to have `--namespace kube-system` added to it if the hard coded namespace were removed.
https://github.com/kubernetes-sigs/aws-efs-csi-driver/blob/eed29e4c149e50452e8f030f31ee05736b1fa160/tester/e2e-test-config.yaml#L30

helm/templates/daemonset.yaml
```diff
@@ -28,36 +30,40 @@ spec:
...
-      {{- end }}
-      priorityClassName: system-node-critical
+    {{- end }}
+    {{- with .Values.affinity }}
+      affinity:
+        {{- toYaml . | nindent 8 }}
+    {{- end }}
+    {{- with .Values.tolerations }}
       tolerations:
-        - operator: Exists
```
krmichel, line 52: Sorry, I didn't notice this before but with this being a daemonset it should probably have a default set of tolerations that will work in many cases.  This toleration is one that tolerates any taint that exists.  In the ebs project they kept this as a default with a boolean value to turn it on an off.

quizy101: I see the relevant PRs for that. I give way too much credit to the admin for how they want to configure their tolerations. I can move `- operator: Exists` into values.yaml as the default for the chart, then the admin can configure it how they want later if they need to. This should keep the default behavior without the need of a bool.

helm/templates/daemonset.yaml
quizy101: Thanks for pointing that out, I changed the install to include the namespace and went back to my original commit with using the namespace from the release. e2e works perfectly now and this is the behavior that I would expect.

helm/templates/_helpers.tpl
quizy101: I'm conflicted on how to proceed between helm2 and helm3. There are several parts that I am trying to address from the original chart that definitely help those that are still on helm2. I would propose that with this PR we get it to a stable helm2 supported version and maybe as a v1 the switch to helm3 happens going forward?

I'm open to whichever the maintainers would like to go with this but my original need was to get PSP and RBAC enabled on the chart. I'm also ok with doing the helm3 cutover PR later.

helm/values.yaml
```diff
@@ -13,15 +11,89 @@ sidecars:
...
+    pullPolicy: IfNotPresent
   nodeDriverRegistrarImage:
     repository: quay.io/k8scsi/csi-node-driver-registrar
     tag: "v1.3.0"
+    pullPolicy: IfNotPresent
 
 imagePullSecrets: []
 nameOverride: ""
 fullnameOverride: ""
+priorityClassName: system-node-critical
```
wongma7, line 23: now that there is a controller component (the deployment in controller.yaml), probably need to prefix this with `nodePriorityClassName` and also have `controllerPrijorityClassName?

helm/templates/_helpers.tpl
wongma7: I do not think it is important to continue supporting helm 2. I agree with you @quizy101 that it can be left as a follow-up PR/discussion though: if you don't need to take advantage of helm 3 features to achieve the goals of this PR then let's leave it for later.

If I understand correctly, we'll simply need to flip chart version to v2 to signal to users that our chart will only support helm v3 from then on.

To be totally clear, since https://github.com/kubernetes-sigs/aws-efs-csi-driver/pull/292 just merged, there is only one chart version in existence: 0.1.0. I want to next release a 0.2.0 with various fixes over the last few monnthsp, including this one if we can get it merged soon. Then 0.3.0 or 1.0.0 could become the helm3 cutoff.

helm/Chart.yaml
```diff
@@ -1,8 +1,8 @@
 apiVersion: v1
 appVersion: "1.0.0"
-name: aws-efs-csi-driver
 description: A Helm chart for AWS EFS CSI Driver
-version: 0.3.0
+name: aws-efs-csi-driver
+version: 0.4.0
```
wongma7, line 5: leave as 0.1.0 for now, since https://github.com/kubernetes-sigs/aws-efs-csi-driver/pull/292 merged and  the only chart in the index is 0.1.0. After https://github.com/kubernetes-sigs/aws-efs-csi-driver/pull/291 merges, I will try to bump the version and include this and everything else between 0.1.0 and now.

helm/values.yaml
```diff
@@ -35,16 +107,15 @@ resources: {}
...
+podAnnotations: {}
+
+nodeSelector:
+  beta.kubernetes.io/os: linux
 
 tolerations: []
 
 affinity: {}
 
-node:
```
wongma7, line 44: Let's keep this, continuing my comment above, since there is a controller component now we are going to need to separate node values from controller values.

ref:
https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/charts/aws-ebs-csi-driver/values.yaml#L86

helm/values.yaml
```diff
@@ -13,15 +11,89 @@ sidecars:
...
 
-podAnnotations: {}
+serviceAccount:
+  # Specifies whether a service account should be created
+  create: false
+  # The name of the service account to use.
+  # If not set and create is true, a name is generated using the fullname template
+  name: ""
+
+rbac:
```
wongma7, line 32: this should be named less ambiguously, it means the psp-related rbac, not the other rbac resopurces  that are necessary for the driver to run

wongma7:
The merging of https://github.com/kubernetes-sigs/aws-efs-csi-driver/pull/274 has caused some conflict for this PR.

I DON'T expect you to add all the same values you did for the node daemonset to the new controller deployment, but PR needs rebase and we now have to disambiguate values based on whehter they will touch the node daemonset or the controller deployment

thanks, I'll try to get this in for the next helm release

helm/templates/daemonset.yaml
wongma7: it seems subject to a lot of debate. https://github.com/helm/helm/issues/5465 I don't mind whether we hardcode the namespace or do this Release.namespace thing, as long as the `helm install` in the test script AND the README work.

ccucui:
Hey folks, any news solving these conflicts?

quizy101:
There are a few conflicting files at this point so it will take a moment to sort out but I would love to get this completed. Thanks @aquam8 for the poke.

PR #247, opened 2020-09-09
deploy/kubernetes/base/node.yaml: Fixing deployment on Bottlerocket.

tatodorov:
Updated the path where are written efs-utils.conf, efs-utils.crt and
privateKey.pem files to /var/amazon/efs which is one of the few
writeable file systems in Bottlerocket.

**Is this a bug fix or adding new feature?**
This resolves issue #246.

**What is this PR about? / Why do we need it?**
This allows aws-efs-csi-driver to be deployed also on Bottlerocket OS where /etc is read-only.

**What testing is done?** 
aws-efs-csi-driver was successfully deployed on Amazon Linux 2 and Bottlerocket OS.

on Bottlerocket OS
```kubectl get nodes -o wide
NAME                              STATUS   ROLES    AGE   VERSION   INTERNAL-IP       EXTERNAL-IP   OS-IMAGE                KERNEL-VERSION   CONTAINER-RUNTIME
ip-192-168-102-136.ec2.internal   Ready    <none>   14m   v1.17.9   192.168.102.136   <none>        Bottlerocket OS 1.0.1   5.4.50           containerd://1.3.7+unknown
ip-192-168-102-151.ec2.internal   Ready    <none>   14m   v1.17.9   192.168.102.151   <none>        Bottlerocket OS 1.0.1   5.4.50           containerd://1.3.7+unknown
ip-192-168-102-72.ec2.internal    Ready    <none>   14m   v1.17.9   192.168.102.72    <none>        Bottlerocket OS 1.0.1   5.4.50           containerd://1.3.7+unknown

kubectl -n kube-system get pods -l app=efs-csi-node -o wide
NAME                 READY   STATUS    RESTARTS   AGE   IP                NODE                              NOMINATED NODE   READINESS GATES
efs-csi-node-bpxd9   3/3     Running   0          20m   192.168.102.136   ip-192-168-102-136.ec2.internal   <none>           <none>
efs-csi-node-mgzpk   3/3     Running   0          19m   192.168.102.151   ip-192-168-102-151.ec2.internal   <none>           <none>
efs-csi-node-mwfm7   3/3     Running   0          20m   192.168.102.72    ip-192-168-102-72.ec2.internal    <none>           <none>
```

Logs from efs-plugin:
```
kubectl -n kube-system logs efs-csi-node-bpxd9 efs-plugin
0909 20:33:43.690761       1 mount_linux.go:163] Cannot run systemd-run, assuming non-systemd OS
[...]

PR #286, opened 2020-12-07
change config dir location

### LINKED ISSUE #246: Unable to deploy aws-efs-csi-driver on bottlerocket
/kind bug

**What happened?**
I deployed a sample EKS with bottlerocket cluster and tried to deploy the aws-csi-driver using the documentation on the github page
```sh
kubectl get nodes -o wide
NAME                                          STATUS   ROLES    AGE     VERSION   INTERNAL-IP     EXTERNAL-IP    OS-IMAGE                KERNEL-VERSION   CONTAINER-RUNTIME
ip-192-168-20-34.us-east-2.compute.internal   Ready    <none>   2d23h   v1.17.9   192.168.20.34   3.131.94.184   Bottlerocket OS 1.0.0   5.4.50           containerd://1.3.7+unknown
ip-192-168-36-82.us-east-2.compute.internal   Ready    <none>   2d23h   v1.17.9   192.168.36.82   3.22.217.215   Bottlerocket OS 1.0.0   5.4.50           containerd://1.3.7+unknown
ip-192-168-75-77.us-east-2.compute.internal   Ready    <none>   2d23h   v1.17.9   192.168.75.77   52.14.56.84    Bottlerocket OS 1.0.0   5.4.50           containerd://1.3.7+unknown
```

error:
```sh
k get pod -n kube-system -l app=efs-csi-node
NAME                 READY   STATUS             RESTARTS   AGE
efs-csi-node-7qkdw   2/3     CrashLoopBackOff   27         116m
efs-csi-node-fbhbq   2/3     CrashLoopBackOff   27         116m
efs-csi-node-sfvdw   2/3     CrashLoopBackOff   27         116m
```

log:
```
k -n kube-system logs -f efs-csi-node-7qkdw -c efs-plugin
I0904 15:44:54.508728       1 mount_linux.go:163] Cannot run systemd-run, assuming non-systemd OS
I0904 15:44:54.508789       1 mount_linux.go:164] systemd-run failed with: exit status 1
I0904 15:44:54.508798       1 mount_linux.go:165] systemd-run output:
I0904 15:44:54.508969       1 driver.go:87] Starting watchdog
I0904 15:44:54.509054       1 efs_watch_dog.go:174] Copying /etc/amazon/efs/efs-utils.conf since it doesn't exist
F0904 15:44:54.509907       1 main.go:50] open /etc/amazon/efs/efs-utils.conf: permission denied
```
aws-csi-driver up and running

* deploy a eks cluster with bottlerocket
then:
```sh
kubectl apply -k "github.com/kubernetes-sigs/aws-efs-csi-driver/deploy/kubernetes/overlays/stable/?ref=release-1.0"
```
I opened a issue with the bottlerocket repo:
https://github.com/bottlerocket-os/bottlerocket/issues/1111

The issue seems to have appeared in kubernetes-sigs/aws-efs-csi-driver@b3baff8, which added an attempt to persist a key to /etc/amazon on the host.

**Environment**
- Kubernetes version (use `kubectl version`): 1.17.x
- Driver version: 1.0

webern:

Fixes #246
Replaces #247

Fixes https://github.com/bottlerocket-os/bottlerocket/issues/1111

``` text
change config dir location

Before this change, the driver fails to work on Bottlerocket because it
tries to write files into the host directory /etc/amazon/efs. On
Bottlerocket that directory is not writable by the container.

We want to move the host mount point to /var/amazon/efs, but we need to
support backward compatibility with hosts that may already have configs
written to /etc/amazon/efs.

To solve this, we mount both host paths and check for the presence of
config files at container runtime. If we find files in the old location,
we symlink to that location, otherwise we symlink to the new location.
```

**What testing is done?**

I built my changes to a container image, pushed to a private ECR repo, and changed yaml files accordingly to use my changes. Throughout, I used a canary application from these instructions https://docs.aws.amazon.com/eks/latest/userguide/efs-csi.html which mounts an EFS volume and logs to it every five seconds.

- First, I deployed the current production version to a cluster that had one AL2 node and one Bottlerocket node.
  - I observed that the deployment failed on Bottlerocket and succeeded on AL2.
  - I ssh'ed to the AL2 box and confirmed that files had been written to `/etc/amazon/efs`.
  - I confirmed that the canary application was writing logs to the EFS mount, but only on AL2, Bottlerocket was broken.
- Next I deployed my updated version of the efs driver.
  - I observed that pods on both hosts were now healthy and writing to their EFS mounts.
  - I ssh'ed to AL2 and found that `/etc/amazon/efs` was still being used and `/var/amazon/efs` was empty.
  - I ssh'ed to Bottlerocket and found that `/etc/amazon/efs` was empty and `/var/amazon/efs/ was being used.
- I rebooted both nodes and found that when the pods resumed, they were still able to write to their EFS mounts.
- I also deployed my changed driver into a clean cluster and observed that both AL2 and Bottlerocket used `/var/amazon/efs` and that mounts were working.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -39,7 +39,7 @@ RUN yum install amazon-efs-utils-1.26-3.amzn2.noarch -y
 # client certificate, in the same config directory can be persisted to host with a host path volume.
 # Otherwise creating a host path volume for that directory will clean up everything inside at the first time.
 # Those static files need to be copied back to the config directory when the driver starts up.
-RUN cp -r /etc/amazon/efs /etc/amazon/efs-static-files
+RUN mv /etc/amazon/efs /etc/amazon/efs-static-files
 
 COPY --from=builder /go/src/github.com/kubernetes-sigs/aws-efs-csi-driver/bin/aws-efs-csi-driver /bin/aws-efs-csi-driver
 COPY THIRD-PARTY /
--- a/deploy/kubernetes/base/node.yaml
+++ b/deploy/kubernetes/base/node.yaml
@@ -45,7 +45,9 @@ spec:
             - name: efs-state-dir
               mountPath: /var/run/efs
             - name: efs-utils-config
-              mountPath: /etc/amazon/efs
+              mountPath: /var/amazon/efs
+            - name: efs-utils-config-legacy
+              mountPath: /etc/amazon/efs-legacy
           ports:
             - containerPort: 9809
               name: healthz
@@ -104,6 +106,10 @@ spec:
             path: /var/run/efs
             type: DirectoryOrCreate
         - name: efs-utils-config
+          hostPath:
+            path: /var/amazon/efs
+            type: DirectoryOrCreate
+        - name: efs-utils-config-legacy
           hostPath:
             path: /etc/amazon/efs
             type: DirectoryOrCreate
--- a/helm/templates/daemonset.yaml
+++ b/helm/templates/daemonset.yaml
@@ -76,7 +76,9 @@ spec:
             - name: efs-state-dir
               mountPath: /var/run/efs
             - name: efs-utils-config
-              mountPath: /etc/amazon/efs
+              mountPath: /var/amazon/efs
+            - name: efs-utils-config-legacy
+              mountPath: /etc/amazon/efs-legacy
           ports:
             - name: healthz
               containerPort: 9809
@@ -136,6 +138,10 @@ spec:
             path: /var/run/efs
             type: DirectoryOrCreate
         - name: efs-utils-config
+          hostPath:
+            path: /var/amazon/efs
+            type: DirectoryOrCreate
+        - name: efs-utils-config-legacy
           hostPath:
             path: /etc/amazon/efs
             type: DirectoryOrCreate
--- a/pkg/driver/config_dir.go
+++ b/pkg/driver/config_dir.go
@@ -1,0 +1,99 @@
+/*
+Copyright 2019 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package driver
+
+import (
+	"fmt"
+	"k8s.io/klog"
+	"os"
+	"path"
+)
+
+// InitConfigDir decides which of two mounted directories will be used to store driver config files. It creates a
+// symlink at etcAmazonEfs to the chosen location (legacyDir or preferredDir). If neither of these locations is present
+// (i.e. when the user does not need to durably store configs and thus does not mount host directories), an empty
+// directory will be created at etcAmazonEfs.
+//
+// - legacyDir    is the path to a config directory where previous versions of this driver may have written config
+//                files. In previous versions of this driver, a host path that was not writeable on Bottlerocket hosts
+//                was being used, so we introduce preferredDir.
+//
+// - preferredDir is the path to config directory that we will use so long as we do not find files in legacyDir.
+//
+// - etcAmazonEfs is the path where the symlink will be written. In practice, this will always be /etc/amazon/efs, but
+//                we take it as an input so the function can be tested.
+// Examples:
+// On a host that has EFS mounts created by an earlier version of this driver, InitConfigDir will detect a conf file in
+// legacyDir and write a symlink at etcAmazonEfs pointing to legacyDir.
+//
+// On a host that does not have pre-existing legacy EFS mount configs, InitConfigDir will detect no files in  legacyDir
+// and will create a symlink at etcAmazonEfs pointing to preferredDir.
+//
+// For a use case in which durable storage of the configs is not required, and no host volumes are mounted, the function
+// creates an empty directory at etcAmazonEfs.
+//
+// If a symlink already existing at etcAmazonEfs, InitConfigDir does nothing. If something other than a symlink exists
+// at etcAmazonEfs, InitConfigDir returns an error.
+func InitConfigDir(legacyDir, preferredDir, etcAmazonEfs string) error {
+
+	// if there is already a symlink or directory in place, we have nothing to do
+	if _, err := os.Stat(etcAmazonEfs); err == nil {
+		klog.Infof("Symlink or directory exists at '%s', no need to create one", etcAmazonEfs)
+		return nil
+	}
+
+	// if neither legacyDir nor preferredDir exist, create an empty directory.
+	if _, err := os.Stat(preferredDir); err != nil {
+		if _, err := os.Stat(legacyDir); err != nil {
+			klog.Infof("Mounted directories do not exist, creating directory at '%s'", etcAmazonEfs)
+			if err := os.MkdirAll(etcAmazonEfs, 0755); err != nil {
+				return fmt.Errorf(
+					"unable to create directory at '%s': %s",
+					etcAmazonEfs,
+					err.Error())
+			}
+			return nil
+		}
+	}
+
+	// check if a conf file exists in the legacy directory and symlink to the directory if so
+	existingConfFile := path.Join(legacyDir, "efs-utils.conf")
+	if _, err := os.Stat(existingConfFile); err == nil {
+		if err = os.Symlink(legacyDir, etcAmazonEfs); err != nil {
+			return fmt.Errorf(
+				"unable to create symlink from '%s' to '%s': %s",
+				etcAmazonEfs,
+				legacyDir,
+				err.Error())
+		}
+		klog.Infof("Pre-existing config files are being used from '%s'", legacyDir)
+		return nil
+	}
+
+	klog.Infof("Creating symlink from '%s' to '%s'", etcAmazonEfs, preferredDir)
+
+	// create a symlink to the config directory
+	if err := os.Symlink(preferredDir, etcAmazonEfs); err != nil {
+		return fmt.Errorf(
+			"unable to create symlink from '%s' to '%s': %s",
+			etcAmazonEfs,
+			preferredDir,
+			err.Error())
+	}
+
+	return nil
+}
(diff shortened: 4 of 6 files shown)


webern:
>  webern force-pushed the webern:bottlerocket branch from 78d281c to d31f20d

This push fixes the author and a couple of comments. I'm still figuring out the paperwork for the CLA.

webern:
>  webern force-pushed the webern:bottlerocket branch from d31f20d to dcee1e2

Hopefully this push will get me passed the CLA?

webern:
I signed it

webern:
I signed it

webern:
Hmmm, OK, I may need some help understanding why the e2e tests are failing. Or maybe a walkthrough of what's actually happening. e.g. does a container with my Go changes get built and used in the test?

webern:
> webern force-pushed the webern:bottlerocket branch from dcee1e2 to 7fc9f51

This push brings the helm chart inline with the other podspec.

> webern force-pushed the webern:bottlerocket branch from 7fc9f51 to e7cc126

This push raises code coverage to 100% for the new function.

helm/templates/daemonset.yaml
```diff
@@ -67,7 +67,9 @@ spec:
             - name: efs-state-dir
               mountPath: /var/run/efs
             - name: efs-utils-config
-              mountPath: /etc/amazon/efs
+              mountPath: /var/amazon/efs
```
wongma7, line 70: I am not 100% sure but I think changes to the helm chart have to be made at the same time as or after the release containing these code changes, i.e. 1.1.0/1.0.1. Otherwise the current helmchart 0.3.0 with driver version 1.0.0 won't work anymore.

LEaving this comment so i can check and be sure...

pkg/driver/config_dir_test.go
```diff
@@ -0,0 +1,304 @@
...
+		t.Fatalf("Unable to create directory: %v", err)
+	}
+
+	// function under test
+	if err := InitConfigDir("whatever1", "whatever2", etcAmazonEfs); err == nil {
+		t.Errorf("InitConfigDir was expected to return an error but did not.")
+	}
+}
+
+// TestInitConfigDirPreferredError asserts that a symlink an error is returned when a symlink cannot be created to the
```
wongma7, line 200: asserts that ~~a symlink~~ an error is returned...

pkg/driver/config_dir_test.go
```diff
@@ -0,0 +1,304 @@
...
+	}
+}
+
+// TestInitConfigDirDisappeared asserts that an error is returned if the preferred config dir is a file.
+func TestInitConfigDirIsFile(t *testing.T) {
+	dir := tempDir(t)
+	defer os.RemoveAll(dir)
+
+	// create an empty legacy dir
+	legacyDir := filepath.Join(dir, "legacy")
```
wongma7, line 285: since most test cases need to create both a legacyDir and preferredDir (mimicking that kubernetes/kubelet would have created them before the Pod started since they are DirectoryOrCreate HostPaths), could be useful; to refactor it into a function

pkg/driver/config_dir_test.go
```diff
@@ -0,0 +1,304 @@
...
+
+	etcAmazonEfs := filepath.Join(dir, "symlink")
+
+	// create a symlink, as if this container has been run previously
+	if err := os.Symlink(preferredDir, etcAmazonEfs); err != nil {
+		t.Fatalf("unable to create symlink %v", err)
+	}
+
+	// function under test
+	if err := InitConfigDir(legacyDir, preferredDir, etcAmazonEfs); err != nil {
```
wongma7, line 158: What about a test case vEery similar to this where InitConfigDir is called twice? Arguably redundant, but in practice that is what will happen

wongma7:
/approve
lgtm

thank you very much I really appreciate the numerous clear test cases , I will let @jqmichael  review also before dropping lgtm

helm/templates/daemonset.yaml
webern: This change was needed to get CI to pass. I don't understand all the packaging implications, but if we back it out we should expect to CI fail. It was this push that got CI to pass https://github.com/kubernetes-sigs/aws-efs-csi-driver/compare/dcee1e279142b1cdcdc452ab2d688f7077788c19..7fc9f513ab23e9a94470f0311575d9f8011e3195

wongma7: O yeah, makes sense, thanks let's keep it

wongma7: You are right, with the new install instruction only after the index gets updated would chart be consumable

pkg/driver/config_dir.go
```diff
@@ -0,0 +1,91 @@
...
+// On a host that does not have pre-existing legacy EFS mount configs, InitConfigDir will detect no files in  legacyDir
+// and will create a symlink at etcAmazonEfs pointing to preferredDir.
+//
+// If a symlink already existing at etcAmazonEfs, InitConfigDir does nothing. If something other than a symlink exists
+// at etcAmazonEfs, InitConfigDir returns an error.
+func InitConfigDir(legacyDir, preferredDir, etcAmazonEfs string) error {
+
+	// if there is already a symlink in place, we have nothing to do
+	if info, err := os.Lstat(etcAmazonEfs); err == nil {
+		if info.Mode()&os.ModeSymlink != 0 {
```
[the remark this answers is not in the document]
wongma7: the verify test passed so seems so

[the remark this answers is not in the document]
webern: Surprisingly `go fmt` removes the spaces that I so purposefully put there in the first place!

Dockerfile
```diff
@@ -39,7 +39,7 @@ RUN yum install amazon-efs-utils-1.26-3.amzn2.noarch -y
 # client certificate, in the same config directory can be persisted to host with a host path volume.
 # Otherwise creating a host path volume for that directory will clean up everything inside at the first time.
 # Those static files need to be copied back to the config directory when the driver starts up.
-RUN cp -r /etc/amazon/efs /etc/amazon/efs-static-files
+RUN mv /etc/amazon/efs /etc/amazon/efs-static-files
```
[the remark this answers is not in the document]
webern: Right, previously you were mounting a host path 'over' this. Now I am writing a symlink in this location, which is simpler if nothing is there already.

deploy/kubernetes/base/node.yaml
```diff
@@ -45,7 +45,9 @@ spec:
             - name: efs-state-dir
               mountPath: /var/run/efs
             - name: efs-utils-config
-              mountPath: /etc/amazon/efs
+              mountPath: /var/amazon/efs
+            - name: efs-utils-config-legacy
+              mountPath: /etc/amazon/efs-legacy
```
wongma7, line 50: the tests that are failing are all related to the dynamic PR merged just now https://github.com/kubernetes-sigs/aws-efs-csi-driver/pull/274.

Could you make the same changes in node.yaml to controller.yaml? shoudl fix it.

webern: Thanks I didn't understand what was going on.

wongma7: Sorry, what I suggested isn't right either, the controller yaml makes no reference to any config dir

Let me take a closer look

wongma7:
Ok, got it, since the controller makes no reference to any config directories, the new function fails like this:
```
I0106 05:34:47.978762       1 config_dir.go:72] Creating symlink from '/etc/amazon/efs' to '/var/amazon/efs'
F0106 05:34:47.978839       1 main.go:60] config directory '/var/amazon/efs' does not exist: 'stat /var/amazon/efs: no such file or directory'
```
I guess it should be oky to remove the check and just create preferredDir /var/amazon/efs if it doesn't exist? In the node setting we can assume that it will be created by hostpath DirectoryOrCreate, in controller setting we don't care about persisting configs at all.

@kbasv since the controller should only briefly mount efs volumes to delete them is it necessary for its config and such to be persistent? I am saying here no, but could you please confirm.

kbasv:
@wongma7 Controller doesn't require persistent configs

webern:
> I guess it should be oky to remove the check and just create preferredDir /var/amazon/efs if it doesn't exist? In the node setting we can assume that it will be created by hostpath DirectoryOrCreate, in controller setting we don't care about persisting configs at all.

OK, If I understand correctly, we also need to run in an environment where the mounts that I was expecting are not there. In that case we probably want the contents of `/etc/amazon/efs-static-files` to be available at `/etc/amazon/efs`. I think the simplest thing to do would be to create a symlink `/etc/amazon/efs` -> `/etc/amazon/efs-static-files` when neither of the mounted directories are present.

wongma7:
@webern yes that makes sense to me

webern:
This has most of my changes. A simplification of the config dir function to account for runtimes where legacy and preferred host paths are not mounted:

> webern force-pushed the webern:bottlerocket branch from 2e6f31a to 216e3f8 9 minutes ago

I missed a couple of refactor opportunities in the tests and added them in this push:

> webern force-pushed the webern:bottlerocket branch from 216e3f8 to ec2d8df 2 minutes ago

I rebased and fixed conflicts in main.go:

> webern force-pushed the webern:bottlerocket branch from ec2d8df to 5ccb581 2 minutes ago

webern:
> webern force-pushed the webern:bottlerocket branch from 5ccb581 to 45acb17 20 seconds ago

I can't make os.CreateDirAll fail reliably in any environment, so I deleted that test. Unfortunately I can't give you 100% coverage as that error return cannot reasonably be exercised.

webern:
> @webern yes that makes sense to me

Actually the right solution was to create an empty dir since the configs will be copied into it.

pkg/driver/config_dir_test.go
```diff
@@ -0,0 +1,243 @@
...
+}
+
+// TestInitConfigDirPreExistingConfigError asserts that an error is returned if a symlink cannot be created to the
+// legacyDir.
+func TestInitConfigDirPreExistingConfigError(t *testing.T) {
+	dir := tempDir(t)
+	defer cleanup(t, dir)
+
+	// create legacy dir and a fake pre-existing conf file
+	// create an empty legacy dir
```
wongma7, line 231: nit extra comment here

wongma7:
/lgtm

beautiful, thank you

Next steps for me:
- Test the upgrade scenario end-to-end w/ both helm and kustomize. (cc https://github.com/kubernetes-sigs/aws-efs-csi-driver/pull/213)
- Make a helm chart and release of the driver. Probably will do minor version bump.

webern:
I fixed the incorrect comment. We needed a retest anyway. The 12 failed tests looked like timeouts, which is what I remember failed tests looking like in the past when a retest passed. Unfortunately fixing that comment removed the lgtm flag.

wongma7:
/lgtm

Yes, just checked logs too and this time the containers all became ready so I don't think it has anything to do with this PR. will repro locally tomorrow if happens again

PR #288, opened 2020-12-17
Added dnsPolicy and dnsConfig to helm chart

nxf5025:
New feature

Allowing the ability to set dnsPolicy & dnsConfig on the helm chart. This will allow a user more control over their dns settings such as forcing the use of AWS nameservers (As shown in the comments of the values.yaml)

**What testing is done?**
helm template as well as an install on an EKS cluster

### DIFF
--- a/helm/Chart.yaml
+++ b/helm/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: "1.0.0"
 name: aws-efs-csi-driver
 description: A Helm chart for AWS EFS CSI Driver
-version: 0.3.1
+version: 0.3.0
 kubeVersion: ">=1.14.0-0"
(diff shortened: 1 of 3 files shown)


wongma7:
/lgtm
/approve
thank you

helm/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: "1.0.0"
 name: aws-efs-csi-driver
 description: A Helm chart for AWS EFS CSI Driver
-version: 0.2.1
+version: 0.3.1
```
wongma7, line 5: should it be 0.3.0

nxf5025: Good catch. Just updated

wongma7:
/hold cancel
/lgtm

thanks again

PR #297, opened 2021-01-08
Add documentation and examples for dynamic provisioning

kbasv:
Documentation update for #274 

Add documentation and provide an example for dynamic provisioning 

**What testing is done?** 
N/A

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -19,8 +19,23 @@ The [Amazon Elastic File System](https://aws.amazon.com/efs/) Container Storage 
 Currently only static provisioning is supported. This means an AWS EFS file system needs to be created manually on AWS first. After that it can be mounted inside a container as a volume using the driver.
 
 The following CSI interfaces are implemented:
-* Node Service: NodePublishVolume, NodeUnpublishVolume, NodeGetCapabilities, NodeGetInfo, NodeGetId
+* Controller Service: CreateVolume, DeleteVolume, ControllerGetCapabilities, ValidateVolumeCapabilities
+* Node Service: NodePublishVolume, NodeUnpublishVolume, NodeGetCapabilities, NodeGetInfo, NodeGetId, NodeGetVolumeStats
 * Identity Service: GetPluginInfo, GetPluginCapabilities, Probe
+
+### CreateVolume Parameters
+| Parameters          | Values | Default | Optional  | Description |
+|---------------------|--------|---------|-----------|-------------|
+| provisioningMode    | efs-ap |         | false     | Type of volume provisioned by efs. Currently, Access Points are supported. |
+| fileSystemId        |        |         | false     | File System under which access points are created. | 
+| directoryPerms      |        |         | false     | Directory permissions for [Access Point root directory](https://docs.aws.amazon.com/efs/latest/ug/efs-access-points.html#enforce-root-directory-access-point) creation. |
+| gidRangeStart       |        | 50000   | true      | Start range of the POSIX group Id to be applied for [Access Point root directory](https://docs.aws.amazon.com/efs/latest/ug/efs-access-points.html#enforce-root-directory-access-point) creation. |
+| gidRangeEnd         |        | 7000000 | true      | End range of the POSIX group Id. |
+| basePath            |        |         | true      | Path under which access points for dynamic provisioning is created. If this parameter is not specified, access points are created under the root directory of the file system |
+
+**Notes**:
+* Custom Posix group Id range for Access Point root directory must include both `gidRangeStart` and `gidRangeEnd` parameters.
+* When using a custom Posix group ID range, there is a possibility for the driver to run out of available POSIX group Ids. We suggest ensuring custom group ID range is large enough or create a new storage class with a new file system to provision additional volumes. 
 
 ### Encryption In Transit
 One of the advantages of using EFS is that it provides [encryption in transit](https://aws.amazon.com/blogs/aws/new-encryption-of-data-in-transit-for-amazon-efs/) support using TLS. Using encryption in transit, data will be encrypted during its transition over the network to the EFS service. This provides an extra layer of defence-in-depth for applications that requires strict security compliance.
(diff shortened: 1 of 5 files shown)


kbasv:
@wongma7 Updated documentation and provided example for #274, please have a look. Thanks!

docs/README.md
```diff
@@ -19,9 +19,24 @@ The [Amazon Elastic File System](https://aws.amazon.com/efs/) Container Storage
 Currently only static provisioning is supported. This means an AWS EFS file system needs to be created manually on AWS first. After that it can be mounted inside a container as a volume using the driver.
 
 The following CSI interfaces are implemented:
+* Controller Service: CreateVolume, DeleteVolume, ControllerGetCapabilities, ValidateVolumeCapabilities
```
wongma7, line 22: I think there should be a (temporary) note that you need a prerelase version of the driver, because I'm planning to release 1.1.0 with the bottlerocket fix and various version bumps, and then do the dynamic provisioning release in 1.2.0 or something.

docs/iam-policy-example.json
```diff
@@ -0,0 +1,15 @@
+{
+  "Version": "2012-10-17",
+  "Statement": [
+    {
+      "Action": [
+        "elasticfilesystem:CreateAccessPoint",
+        "elasticfilesystem:DeleteAccessPoint",
+        "elasticfilesystem:DescribeFileSystems",
```
wongma7, line 8: what about permission to mount access points to delete their contents? Since the controller is the one creating the access point i guess it doesn't need it?

docs/README.md
kbasv: What will be the pre-release version?

wongma7: Hmm since it doesn't exist yet, I think we can hold this PR until it does, then we will be able to provide clearer instructions for how to try it

docs/iam-policy-example.json
kbasv: Not required for the controller as these permissions are used to talk to EFS APIs. IIRC, the mount permissions are defined on the file systems and exist as file system policies to limit access to file system.

wongma7: got it.

the default for a given fs is to allow all mountsso if user has some kind of custom restriction on filesystem they need to edit that policy themself https://docs.aws.amazon.com/efs/latest/ug/iam-access-control-nfs-efs.html

docs/README.md
```diff
@@ -52,14 +67,20 @@ The following sections are Kubernetes specific. If you are a Kubernetes user, us
...
 * Encryption of data in transit - EFS file systems are mounted with encryption in transit enabled by default in the master branch version of the driver.
 
 **Notes**:
 * Since EFS is an elastic file system it doesn't really enforce any file system capacity. The actual storage capacity value in persistent volume and persistent volume claim is not used when creating the file system. However, since the storage capacity is a required field by Kubernetes, you must specify the value and you can use any valid value for the capacity.
 
 ### Installation
-Deploy the driver:
+####Set up driver permission:
+The driver requires IAM permission to talk to Amazon EFS to manage the volume on user's behalf. There are several methods to grant driver IAM permission:
+* Using IAM Role for Service Account: create an [IAM Role for service accrounts](https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html) with the [required permissions](./iam-policy-example.json). Uncomment annotations and put the IAM role ARN in [service-account manifest](../deploy/kubernetes/base/serviceaccount-csi-controller.yaml)
```
wongma7, line 80: this is only available for EKS so we should mark it "recommended if you're using EKS"

kbasv: Sure, will update it

benjamin-tucker: line 80 - typo 'accrounts'.   And Thanks - this literally is making our site smile :-)

examples/kubernetes/dynamic_provisioning/specs/storageclass.yaml
```diff
@@ -0,0 +1,12 @@
...
+metadata:
+  name: efs-sc
+provisioner: efs.csi.aws.com
+parameters:
+  provisioningMode: efs-ap
+  fileSystemId: fs-92107410
+  directoryPerms: "700"
+  gidRangeStart: "1000"
+  gidRangeEnd: "2000"
+  basePath: "/dynamic_provisioning"
```
jqmichael, line 12: Can we add comments in the example that these are optional? e.g.

```
basePath: "/dynamic_provisioning" #optional 
```

kbasv: Updated

examples/kubernetes/dynamic_provisioning/README.md
```diff
@@ -0,0 +1,49 @@
+## Dynamic Provisioning
+This example shows how to create a dynamically provisioned volume created through [EFS access points](https://docs.aws.amazon.com/efs/latest/ug/efs-access-points.html) and Persistent Volume Claim (PVC) and consume it from a pod.
+
+**Note**: this example requires Kubernetes v1.13+ and driver version >= 1.1.0.
```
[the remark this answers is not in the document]
wongma7: Also the kubernetes v ersion is not accurate anymore, I think we need 1.17+ because of the sidecars like the version of external-provisioner is only compatible with 1.17+ . Can you double check this and edit.

examples/kubernetes/dynamic_provisioning/specs/storageclass.yaml
```diff
@@ -0,0 +1,12 @@
+kind: StorageClass
```
wongma7, line 1: Do you mind also removing the storageclass and all references to it from the `static_provisioning` example? Now that the driver supports both, having a SC there will only cause confusion. https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/784

wongma7: actually I can do this in a follow up

examples/kubernetes/dynamic_provisioning/README.md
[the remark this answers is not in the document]
kbasv: Agreed. Updated

PR #305, opened 2021-01-14
WIP: DO NOT MERGE! Revert "Introduce Controller with Create and Delete Volume capability"

wongma7:
This is for testing a hypothetical 1.1 release that has bottlerocket (and other small) fixes but not dynamic provisioning.

I won't merge this into master. I plan to creatge a release-1.1 branch thta looks like this. I could also rebase and drop the controller commit but I think having an explicit revert is cleaner.

Ref, check commits since Jul 22:

https://github.com/kubernetes-sigs/aws-efs-csi-driver/commits/master/cmd
https://github.com/kubernetes-sigs/aws-efs-csi-driver/commits/master/pkg

This reverts commit f2fc7cee45cf87ad8bd73ccc3280b99e1b75dc8b.

**What testing is done?**


gazal-k:
we do like the dynamic provisioning functionality. Is the plan to add it back soon?

PR #309, opened 2021-01-19
Add tags to EFS resources. Update EFS Utils

### LINKED ISSUE #303: Dynamic PV tests flaky
/kind bug
https://testgrid.k8s.io/provider-aws-efs-csi-driver#e2e-test&width=20

for example, README fix failing tests: https://github.com/kubernetes-sigs/aws-efs-csi-driver/pull/301

kbasv:

* Allow adding tags to EFS resources to configure IAM policy condition based on tags.
* Add default tags and provide a command line option to provide more tags.
* Update RBAC permissions and efs-utils version.
*  Fixes #303 

**What testing is done?** 

Tested against my EKS cluster. 

Ran `make test-e2e` on a fresh ec2 instance

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -41,6 +41,7 @@ func main() {
 		volMetricsFsRateLimit    = flag.Int("vol-metrics-fs-rate-limit", 5, "Volume metrics routines rate limiter per file system")
 		deleteAccessPointRootDir = flag.Bool("delete-access-point-root-dir", false,
 			"Opt in to delete access point root directory by DeleteVolume. By default, DeleteVolume will delete the access point behind Persistent Volume and deleting access point will not delete the access point root directory or its contents.")
+		tags = flag.String("tags", "", "Space separated key:value pairs which will be added as tags for EFS resources. For example, 'environment:prod region:us-east-1'")
 	)
 	klog.InitFlags(nil)
 	flag.Parse()
@@ -59,7 +60,7 @@ func main() {
 	if err != nil {
 		klog.Fatalln(err)
 	}
-	drv := driver.NewDriver(*endpoint, etcAmazonEfs, *efsUtilsStaticFilesPath, *volMetricsOptIn, *volMetricsRefreshPeriod, *volMetricsFsRateLimit, *deleteAccessPointRootDir)
+	drv := driver.NewDriver(*endpoint, etcAmazonEfs, *efsUtilsStaticFilesPath, *tags, *volMetricsOptIn, *volMetricsRefreshPeriod, *volMetricsFsRateLimit, *deleteAccessPointRootDir)
 	if err := drv.Run(); err != nil {
 		klog.Fatalln(err)
 	}
--- a/pkg/driver/driver.go
+++ b/pkg/driver/driver.go
@@ -19,6 +19,7 @@ package driver
 import (
 	"context"
 	"net"
+	"strings"
 
 	"github.com/container-storage-interface/spec/lib/go/csi"
 	"google.golang.org/grpc"
@@ -46,9 +47,10 @@ type Driver struct {
 	volStatter               VolStatter
 	gidAllocator             GidAllocator
 	deleteAccessPointRootDir bool
+	tags                     map[string]string
 }
 
-func NewDriver(endpoint, efsUtilsCfgPath, efsUtilsStaticFilesPath string, volMetricsOptIn bool, volMetricsRefreshPeriod float64, volMetricsFsRateLimit int, deleteAccessPointRootDir bool) *Driver {
+func NewDriver(endpoint, efsUtilsCfgPath, efsUtilsStaticFilesPath, tags string, volMetricsOptIn bool, volMetricsRefreshPeriod float64, volMetricsFsRateLimit int, deleteAccessPointRootDir bool) *Driver {
 	cloud, err := cloud.NewCloud()
 	if err != nil {
 		klog.Fatalln(err)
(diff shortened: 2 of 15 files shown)


cmd/main.go
```diff
@@ -41,6 +41,7 @@ func main() {
 		volMetricsFsRateLimit    = flag.Int("vol-metrics-fs-rate-limit", 5, "Volume metrics routines rate limiter per file system")
 		deleteAccessPointRootDir = flag.Bool("delete-access-point-root-dir", false,
 			"Opt in to delete access point root directory by DeleteVolume. By default, DeleteVolume will delete the access point behind Persistent Volume and deleting access point will not delete the access point root directory or its contents.")
+		tags = flag.String("tags", "", "Space separated key:value pairs which will be added as tags for EFS resources. For example, 'environment:prod region:us-east-1'")
```
wongma7, line 44: this looks like it needs gofmt.

I am not sure I like the accepted format,  but I don't have a better one to suggest. I checked the awscli convention and it's more verbose and harder to parse...:
`Key=string,Value=string ...`

pkg/driver/efs_watch_dog.go
```diff
@@ -78,6 +78,14 @@ tls_cert_renewal_interval_min = 60
 
 [client-info] 
 source={{.EfsClientSource}}
+
+[cloudwatch-log]
+# enabled = true
```
wongma7, line 83: wait so is it enabled? Or just placeholder?

test/e2e/e2e.go
```diff
@@ -157,6 +157,21 @@ var _ = ginkgo.SynchronizedBeforeSuite(func() []byte {
 
 		FileSystemId = id
 		ginkgo.By(fmt.Sprintf("Created EFS filesystem %q in region %q for cluster %q", FileSystemId, Region, ClusterName))
+		//Sleep 5 minutes to allow file system bootstrapping
+		ginkgo.By("Sleeping 5 minutes to allow file system bootstrapping")
+		time.Sleep(5 * time.Minute)
+		ginkgo.By("Done sleeping. Now proceeding to test access point creation")
+		accessPointId, err := c.CreateAccessPoint(FileSystemId, ClusterName)
+		if err != nil {
+			framework.ExpectNoError(err, "creating access point")
```
wongma7, line 166: not sure if this check will work. from the e2e logs it seems like creation silently succeeds. But it's worth a try.

cmd/main.go
kbasv: This is how gofmt like it. I tried padding with enough spaces to match above variables, but `make verify` failed.

I feel `key:value` is easier to parse than `key=string,value=string` for multiple tags

pkg/driver/efs_watch_dog.go
kbasv: By default, it is not enabled. However if you set `enabled = true` and have the right cloud watch permissions, efs-utils will start logging to cloud watch.

test/e2e/e2e.go
kbasv: Looks like creating is not enough, I'll need to mount file system using access points too for all the boot strapping to complete.

kbasv:
@wongma7 dynamic provisioning e2e tests passes now

daisuke-yoshimoto:
@wongma7 @nckturner @kbasv 
Is this PR not yet merged?

daisuke-yoshimoto:
@wongma7 @nckturner @kbasv
Is there any update on this?

wongma7:
coveralls is blocking. I guess it has a point this time, it wouldn't hurt for parsing functions to have some tests

pkg/driver/driver.go
```diff
@@ -124,3 +127,23 @@ func (d *Driver) Run() error {
 	klog.Infof("Listening for connections on address: %#v", listener.Addr())
 	return d.srv.Serve(listener)
 }
+
+func parseTagsFromStr(tagStr string) map[string]string {
+	defer func() {
+		if r := recover(); r != nil {
+			klog.Errorf("Failed to parse input tag string: %v", tagStr)
```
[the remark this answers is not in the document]
kbasv: I can add a test here.

The panicking and recovering is when there is a typo in the input or the key value pair is not of the form `key`:`value`. For example: `key-value key;value`

test/e2e/e2e.go
```diff
@@ -157,6 +159,47 @@ var _ = ginkgo.SynchronizedBeforeSuite(func() []byte {
...
+		if err = makeDir(target); err != nil {
+			framework.ExpectNoError(err, fmt.Sprintf("Failed to create temp directory %q", target))
+		}
+		defer os.RemoveAll(target)
+
+		// mount access point
+		command := exec.Command("/bin/sh", "-c", "mount", "-t", "efs", "-o", "tls,accesspoint="+accessPointId, FileSystemId, target)
+		if err = command.Run(); err != nil {
+			framework.ExpectNoError(err, "Failed to mount using access point")
+		}
```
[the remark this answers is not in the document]
kbasv: This is a setup step to bootstrap access points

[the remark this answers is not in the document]
wongma7: do we expect users to do this? Do we need to document the need to bootstrap access points anywhere?

pkg/driver/driver.go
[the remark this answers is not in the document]
wongma7: Also WDYT about parsing at startup and fast failing instead of at createvolume time?

With current implementation, driver will startup and mount/unmount will still work but then all createvolumes will fail and the user will have to do debugging. Whereas if we fast fail they will presumably see the error message printed by the plugin

wongma7:
One last question: what do yo uthink about making tags a storageclass parameter instead of a command line flag?

It is more flexible storageclass a could have its volumes tagged a certain way, b another way, and as a user I can more easijly audit  both.

kbasv:
I like the idea of tags in the storage class parameters. We can eventually move it as a storage class parameter or we can do it now. I'm okay with either way

wongma7:
/lgtm

I am really wary of the e2e test changes, I think we need to document, but in interest of gathering data let's merge it and see if it helps with the test flakiness

Issue #320, opened 2021-02-02
Reuse EBS E2E testing scripts (git subtree) for creating  kops clusters


wongma7:
/feature

Depending on aws-k8s-tester for testing is more complicated than calling scripts like https://github.com/kubernetes-sigs/aws-ebs-csi-driver/tree/master/hack/e2e straight up. Namely we have to install this binary every CI run and it turns out building go modules in CI environment is not very reliable : https://github.com/kubernetes-sigs/aws-efs-csi-driver/pull/315

The EBS scripts are ALMOST usable out of the box. I will copy the same approach as csi-release-tools .

wongma7:
resolved

PR #322, opened 2021-02-03
WIP: subtree aws-ebs-csi-driver/hack/e2e

wongma7:
**Is this a bug fix or adding new feature?** feature

**What is this PR about? / Why do we need it?** To reduce code duplication and avoid dependency on go binary/go mod, let's copy the basic kops and helm testing bash scripts from ebs. fixes https://github.com/kubernetes-sigs/aws-efs-csi-driver/issues/320

**What testing is done?** Works locally:

TEST_ID=18512 CLEAN=false KOPS_STATE_FILE=s3://mattwon-e2e GINKGO_NODES=1 make test-e2e

Wil lcheck if works in CI.

If so, I'll push the changes to hack/e2e back up to aws-ebs-csi-driver. Either before merging this then rebasing or after, doesn't much matter.


wongma7:
pull-aws-efs-csi-driver-e2e passed.

The test took half the time because previously we were calling go test and now we are calling ginkgo -p -nodes=4 to run them in parallel.

Issue #328, opened 2021-02-10
efs-csi-node DaemonSet schedules to Fargate nodes with included example


benmccown-amz:
/kind bug

**What happened?**

When following the EFS CSI documentation at [1] the efs-csi-node DaemonSet attempts to schedule to fargate based nodes if you are running any fargate workloads within your EKS cluster. This results in pods failing to deploy.

[1] https://docs.aws.amazon.com/eks/latest/userguide/efs-csi.html

**What you expected to happen?**

I would expect if I were to follow the documentation provided that the DaemonSet would not scheduled pods to fargate nodes unless it is required and built to function and schedule correctly.

**How to reproduce it (as minimally and precisely as possible)?**

1. Deploy EKS Fargate workloads within your cluster
2. Install the EFS CSI as per the docs at [1].
3. Get pods in kube-system to see the pods in the Pending status

[1] https://docs.aws.amazon.com/eks/latest/userguide/efs-csi.html

**Anything else we need to know?**:

This is easily resolvable with a node anti affinity rule on the DaemonSet. I have a branch that I have tested which resolves this issue. I will further detail my changes in a PR.

**Environment**
- Kubernetes version (use `kubectl version`):

```
Client Version: version.Info{Major:"1", Minor:"18+", GitVersion:"v1.18.9-eks-d1db3c", GitCommit:"d1db3c46e55f95d6a7d3e5578689371318f95ff9", GitTreeState:"clean", BuildDate:"2020-10-20T22:21:03Z", GoVersion:"go1.13.15", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"18+", GitVersion:"v1.18.9-eks-d1db3c", GitCommit:"d1db3c46e55f95d6a7d3e5578689371318f95ff9", GitTreeState:"clean", BuildDate:"2020-10-20T22:18:07Z", GoVersion:"go1.13.15", Compiler:"gc", Platform:"linux/amd64"}
```

- Driver version: 1.0

benmccown-amz:
Modified spec.template.spec for node.yaml

```
    spec:
      nodeSelector:
        kubernetes.io/os: linux
      affinity:
        nodeAffinity:
          requiredDuringSchedulingIgnoredDuringExecution:
            nodeSelectorTerms:
            - matchExpressions:
              - key: eks.amazonaws.com/compute-type
                operator: NotIn
                values:
                - fargate
```

[...]

PR #329, opened 2021-02-10
Daemonset Antiaffinity for Fargate Nodes

benmccown-amz:

Bug fix. Relating to [Issue 328](https://github.com/kubernetes-sigs/aws-efs-csi-driver/issues/328)

By default when following the docs at [1] with a mixed ec2/fargate EKS cluster the efs-csi-node DaemonSet will try to schedule to fargate nodes which results in pods becoming stuck in the "Pending" status.

This is easily resolvable with a node anti affinity rule.

[1] https://docs.aws.amazon.com/eks/latest/userguide/efs-csi.html

**What testing is done?** 

I have tested the changes on my branch within a test cluster using the command below which is the exact command from the documentation but with a modified endpoint pointing to my test branch.

```
kubectl apply -k https://github.com/benmccown-amz/aws-efs-csi-driver/deploy/kubernetes/overlays/stable/ecr?ref=daemonset-antiaffinity-fargate
```

### DIFF
--- a/deploy/kubernetes/base/node.yaml
+++ b/deploy/kubernetes/base/node.yaml
@@ -20,6 +20,15 @@ spec:
     spec:
       nodeSelector:
         kubernetes.io/os: linux
+      affinity:
+        nodeAffinity:
+          requiredDuringSchedulingIgnoredDuringExecution:
+            nodeSelectorTerms:
+            - matchExpressions:
+              - key: eks.amazonaws.com/compute-type
+                operator: NotIn
+                values:
+                - fargate
       hostNetwork: true
       priorityClassName: system-node-critical
       tolerations:


benmccown-amz:
Just got CLA sorted.

wongma7:
/approve

thank you, could you squash the commits into 1. e.g. git rebase -i upstream/master ``

benmccown-amz:
@wongma7 commits have been squashed.

wongma7:
/lgtm

thanks again

nicogl:
@wongma7 @ben-mccown
The DaemonSet antiaffinity for Fargate nodes introduced here doesn't solve the issue. I tested it and still get PodNotSupported errors. I think the problem is that the way the "**_matchExpressions_**" element is defined doesn't have an infuence on the NodeGroup nodes:

```
    - matchExpressions:
        - key: eks.amazonaws.com/compute-type
          operator: NotIn
          values:
          - fargate
```

As per my understanding, above condition requires that the label "_**eks.amazonaws.com/compute-type**_" exists on the node for it to be evaluated. Now note that this label is only present on Fargate nodes, not on EC2 nodes beloging to a node group. This can be checked by running this command:  `kubectl get nodes --show-labels`

The way I fixed it was changing the affinity to:

```
      affinity:
        nodeAffinity:
          requiredDuringSchedulingIgnoredDuringExecution:
            nodeSelectorTerms:
            - matchExpressions:
              - key: eks.amazonaws.com/nodegroup
                operator: Exists
                values: []
```
This way I ensure that the related pods are run on a node group and not Fargate.

Last but not least, this affinity also has to be added to the "efs-csi-controller" Deployment element. Otherwise, the controller pods also throw errors.

Hope this makes sense, kind regards.

PR #349, opened 2021-03-10
Add PVC and PV metadata as tags to EFS access points

edanc:
new feature

Gets the extra metadata and adds it as tags to the access point. 

**What testing is done?** 
Tested against my eks cluster.  

Most of this work was based from https://github.com/kubernetes-sigs/aws-efs-csi-driver/pull/309.

### DIFF
--- a/charts/aws-efs-csi-driver/templates/controller-deployment.yaml
+++ b/charts/aws-efs-csi-driver/templates/controller-deployment.yaml
@@ -80,6 +80,7 @@ spec:
             - --v=5
             - --feature-gates=Topology=true
             - --leader-election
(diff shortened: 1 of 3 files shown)


charts/aws-efs-csi-driver/templates/controller.yaml
```diff
@@ -81,6 +81,7 @@ spec:
             - --v=5
             - --feature-gates=Topology=true
             - --leader-election
+            - --extra-create-metadata
```
edanc, line 84: https://github.com/kubernetes-csi/external-provisioner/pull/399

Issue #360, opened 2021-03-16
helm chart hardcoded namespace to kube-system


debu99:
are you only allowing to install in kube-system?

wongma7:
Let's copy https://github.com/kubernetes-sigs/aws-ebs-csi-driver/pull/768

PR #364, opened 2021-03-16
helm chart clean up: Remove hardcoded kube-system, set controller default true, generate kustomize from helm, bump sidecars

wongma7:

- Remove hardcoded kube-system: fixes https://github.com/kubernetes-sigs/aws-efs-csi-driver/issues/360
- set controller default true: This is in preparation for release 1.2.0.  if folks try to install the helm chart from master branch then the controller pod won't work or might crashloop since 1.1.1 doesn't have any controller capabilities
- generate kustomize from helm: our kustomize and helm have gotten out of sync, this make rule is to ensure they stay similar.
- bump sidecars: bumped livenessprobe and registrar to latest and pull from gcr instead of quay. No breaking changes, but the healthz-port flagi sdeprecated so we should move off it later.
**What testing is done?**


wongma7:
@kbasv 

we will need to set controller true in preparation for 1.2 release (see PR description).

Alongside this change I have bundled various clean up fixes, PTAL

charts/aws-efs-csi-driver/templates/controller-deployment.yaml
```diff
@@ -4,7 +4,6 @@ kind: Deployment
 apiVersion: apps/v1
 metadata:
   name: efs-csi-controller
-  namespace: kube-system
```
kbasv, line 7: Should this be replaced with {{ .Release.Namespace}}?

charts/aws-efs-csi-driver/templates/node-daemonset.yaml
```diff
@@ -3,7 +3,6 @@ kind: DaemonSet
 apiVersion: apps/v1
 metadata:
   name: efs-csi-node
-  namespace: kube-system
```
kbasv, line 6: Same as above

deploy/kubernetes/base/controller-deployment.yaml
```diff
@@ -29,7 +31,7 @@ spec:
         - name: efs-plugin
           securityContext:
             privileged: true
-          image: amazon/aws-efs-csi-driver:latest
+          image: "amazon/aws-efs-csi-driver:v1.1.1"
```
kbasv, line 34: Do we bump this 1.2?

charts/aws-efs-csi-driver/templates/controller-deployment.yaml
wongma7: following the ebs version of the PR, it's fine to leave it blank and it will automagically get created in the namesapce you pass via helm install -n namespace

deploy/kubernetes/base/controller-deployment.yaml
wongma7: Yeah what I'm planning is

1. create release-1.2 branch
2. push changelog, helm chart changes, etc. to release-1.2 branch
3. tag the HEAD of release-1.2 branch with v1.2.0 / v1.2.0-rc.1
4. after verifying image has pushed etc., cherry-pick the commit from step 2 into master branch.

wongma7:
@kbasv I added one commit to bump sidecars. They  don't have any breaking changes (still say they are compatible with 1.13 and above). However, csi-provisioner is only compatible with 1.17 and above so that will be the constraint for 1.2 release.

kbasv (approved):

charts/aws-efs-csi-driver/templates/controller-deployment.yaml
jqmichael: What if they don't pass in the namespace flag and continue with what they've been used to?

wongma7: i tested, it will go to default namespace. or whatever namespace their current context is

wongma7: updated helm upgrade command in README to include namespace flag. same as ebs repo https://github.com/kubernetes-sigs/aws-ebs-csi-driver#deploy-driver

wongma7:
coveralls bugged again :angry:

Skaronator:
I can see that there is now a release-1.2 docker tag but no Github Relase and no Helm Chart release. Will that follow? Thanks.

Issue #294, opened 2021-01-06
Push images to public ECR and private ECR


wongma7:
Similar to https://github.com/kubernetes-sigs/aws-ebs-csi-driver, community gcr `k8s.gcr.io/provider-aws/aws-efs-csi-driver` should be the canonical source of images. Docker Hub has throttling now.

Current push destinations:
1. Docker Hub: amazon/aws-efs-csi-driver
2. GitHub registry
3. EKS private ECR: 602401143452.dkr.ecr.us-west-2.amazonaws.com/eks/aws-efs-csi-driver

Desired push destinations:
1. ~~Community GCR: k8s.gcr.io/provider-aws/aws-efs-csi-driver~~
2. ~~Docker Hub: amazon/aws-efs-csi-driver~~ deprecated
- ~~GitHub registry~~ (remove because nobody uses it, plus a v2 GitHub registry is in beta that might see more use.)
3. EKS private ECR: 602401143452.dkr.ecr.us-west-2.amazonaws.com/eks/aws-efs-csi-driver
    - also need the sidecars here otherwise users with air gapped clusters can't pull from eks-d
4. EKS public ECR
    -  currently rely on eks-d sidecar images but we also need to provide our own public ecr csi driver image

tsarna:
Given that this container is for use on AWS, I would say public ECR should be a definite yes!

universam1:
when can we expect that AWS publishes their images in their registry?

gowthams316:
Can we also have the other side car images as part of EFS CSI driver per https://github.com/kubernetes-sigs/aws-efs-csi-driver/blob/master/deploy/kubernetes/overlays/stable/kustomization.yaml are also updated to point to ECR repo `602401143452.dkr.ecr.<region>.amazonaws.com/eks/aws-efs-csi-driver`

wongma7:
@gowthams316 the ecr kustomize repo should already point to private 602401143452 ECR for the driver image and public eks-distro ECR for the sidecars. Do you mean that it should all point to either private ECr or all to public ECR?

Cleaning up the image and yaml distribution is on the roadmap but no ETA right now. Here is the current state, it is rather messy

Installation method | Source of driver image | Source of sidecar images
-- | -- | --
helm chart | Docker Hub (official amazon org) | Public ECR (owned by EKS-D)
kustomize stable overlay | Docker Hub (official amazon org) | Community GCR
[...]

PR #368, opened 2021-03-18
Release v1.2.0

kbasv:
Release 1.2.0

**What testing is done?**

### DIFF
--- a/charts/aws-efs-csi-driver/values.yaml
+++ b/charts/aws-efs-csi-driver/values.yaml
@@ -6,7 +6,7 @@ replicaCount: 2
 
 image:
   repository: amazon/aws-efs-csi-driver
-  tag: "v1.1.1"
+  tag: "v1.2.0"
   pullPolicy: IfNotPresent
 
 sidecars:
@@ -18,7 +18,7 @@ sidecars:
     tag: "v2.1.0"
   csiProvisionerImage:
     repository: k8s.gcr.io/sig-storage/csi-provisioner
-    tag: "v2.1.1"
+    tag: "v2.0.3"
 
 imagePullSecrets: []
 nameOverride: ""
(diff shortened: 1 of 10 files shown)


charts/aws-efs-csi-driver/values.yaml
```diff
@@ -5,8 +5,8 @@
 replicaCount: 2
 
 image:
-  repository: amazon/aws-efs-csi-driver
-  tag: "v1.1.1"
+  repository: k8s.gcr.io/provider-aws/aws-efs-csi-driver
```
wongma7, line 8: my b, i forgot we haven't set up GCR yet for this repo unlike EBS. we are still going to need to point  to docker hub by default (amazon/aws-efs-csi-driver) please regenerate the templates : (

wongma7: https://github.com/kubernetes-sigs/aws-efs-csi-driver/issues/294

kbasv: Oops, sure thing

deploy/kubernetes/base/controller-deployment.yaml
```diff
@@ -31,7 +31,7 @@ spec:
         - name: efs-plugin
           securityContext:
             privileged: true
-          image: "amazon/aws-efs-csi-driver:v1.1.1"
+          image: "amazon/aws-efs-csi-driver:latest"
```
wongma7, line 34: we don't push latest tag anymore, so latest is probably ponting to some super old version.

I don't think it matters since our README instructs to use the stable overlay...installing the base templates as-is, naked, is probably not a good idea in any case.

kbasv: So I'll update it to v1.2.0 here?

PR #372, opened 2021-03-19
Update ReadMe file

kbasv:
ReadMe file update

**What testing is done?**


wongma7:
/retest

I think ever since we started using the scripts, the tests are running in parallel, and the test case where we restart the driver could interfere with parallel test cases. I think it coudl be a legitimate bug, because I doubt the interference explains parallel test cases timing out after 5 minutes, butg we need https://github.com/kubernetes-sigs/aws-efs-csi-driver/pull/373 to help debug

PR #377, opened 2021-03-21
Release 1.2

kbasv:
Helm chart update for release 1.2

**What testing is done?**


.github/workflows/release.yaml
```diff
@@ -11,9 +11,6 @@ jobs:
     steps:
       - name: Checkout code
         uses: actions/checkout@v1
-      - name: Create Helm chart
-        run: |
-          tar cvzf helm-chart.tgz helm
```
jqmichael, line 16: Did we delete this because it was never being used?

kbasv: I ported this from master branch. I think @wongma7 has more context on this

jqmichael: >Matthew Wong
> we have a separate github action just for releasing the helm chart, and we moved the location of the helm chart, so that one was broken. and yes, the tgz is kind of useless since we recommend using helm repo approach

charts/aws-efs-csi-driver/templates/controller-deployment.yaml
```diff
@@ -34,7 +34,7 @@ spec:
         {{- with .Values.nodeSelector }}
         {{- . | toYaml | nindent 8 }}
         {{- end }}
-      {{- if .Values.serviceAccount.controller.create }}
+      {{- if .Values.controller.create }}
```
wongma7, line 37: this is redundant now since the whole file is wrapped in if .Values.controller.create

charts/aws-efs-csi-driver/templates/controller-serviceaccount.yaml
```diff
@@ -9,6 +9,7 @@ metadata:
   annotations:
     {{- toYaml . | nindent 4 }}
   {{- end }}
+  {{- end }}
```
wongma7, line 12: nit indenting is off, should have on ident

charts/aws-efs-csi-driver/templates/controller-deployment.yaml
kbasv: okay - I'll get rid of this

wongma7:
@kbasv can you also include https://github.com/kubernetes-sigs/aws-efs-csi-driver/pull/383 .

I also reproduced the error locally.

Issue #313, opened 2021-01-27
efs-csi-controller won't start if IMDS access is blocked


korbin:
/kind bug

**What happened?**

With IMDS disabled per best practices (https://docs.aws.amazon.com/eks/latest/userguide/best-practices-security.html) on Bottlerocket hosts, pods from the `efs-csi-controller` deployment will not start.

We need something similar for the controller, or for it to just not need IMDS access to begin with.

```F0127 18:13:01.145009       1 driver.go:54] could not get metadata from AWS: EC2 instance metadata is not available```
is emitted to the log and a crash occurs.

**What you expected to happen?**

I expected  `efs-csi-controller` to start. Passing the region/instance ID/other IMDS-sourced information would be acceptable.

**How to reproduce it (as minimally and precisely as possible)?**

- Block IMDS access
- Deploy `efs-csi-controller`

**Anything else we need to know?**:

The DaemonSet uses `hostNetwork: true` to regain access to the IMDS (https://github.com/kubernetes-sigs/aws-efs-csi-driver/pull/188)

**Environment**
- Kubernetes version (use `kubectl version`):
EKS 1.18
- Driver version:
`master`

korbin:
I was able to get the `efs-csi-controller` running by removing the liveness check and switching to `hostNetwork: true` until the IMDS dependency can be resolved.

wongma7:
Yes, seems we'll have to add hostNetwork true, thank you for trying and verifying it fixes it.

(I am looking into ways to avoid talking to instance metadata altogther since we only use it for super basic stuff like instance id but not even sure if it's feasible yet)

davidshtian:
Tried at on premises physical servers environment (not AWS env), and it still throws errors below. Any extra configuration need be be done for this scenario? Thanks.

`could not get metadata from AWS: EC2 instance metadata is not available `

davidshtian:
@korbin Hi Korbin~ Got the same issue for on premises physical servers Kubernetes environment. For this workaround, tried to remove livenessProbe part for container efs-plugin in controller deployment (https://github.com/kubernetes-sigs/aws-efs-csi-driver/blob/master/deploy/kubernetes/base/controller-deployment.yaml), and the error still exist. Also need to remove the container liveness-probe? Thanks.

> I was able to get the `efs-csi-controller` running by removing the liveness check and switching to `hostNetwork: true` until the IMDS dependency can be resolved.

groodt:
[...]

PR #380, opened 2021-03-22
 Add `hostNetwork: true` on efs-csi-controller deployement.

jihed:
This will add the fix for #313 on helm chart.

As #313 still open, the helm chart needs to enable hostNetwork on `efs-csi-driver` deployment. 

**What testing is done?** 
``` kgpo
+ kubectl get pods
NAME                                           READY   STATUS    RESTARTS   AGE
efs-csi-controller-54967db5bd-nq4qv            3/3     Running   0          9m2s
efs-csi-node-865bp                             3/3     Running   0          3d1h
efs-csi-node-g728g                             3/3     Running   0          3d1h
efs-csi-node-jbplr                             3/3     Running   0          3d1h
efs-csi-node-jdxb7                             3/3     Running   0          3d1h
efs-csi-node-mhmm5                             3/3     Running   0          3d1h
efs-csi-node-npgsc                             3/3     Running   0          3d1h
efs-csi-node-xp4tm                             3/3     Running   0          3d1h```

### DIFF
--- a/deploy/kubernetes/base/controller-serviceaccount.yaml
+++ b/deploy/kubernetes/base/controller-serviceaccount.yaml
@@ -47,7 +47,7 @@ metadata:
 subjects:
   - kind: ServiceAccount
     name: efs-csi-controller-sa
-    namespace: default
+    namespace: sample
 roleRef:
   kind: ClusterRole
   name: efs-csi-external-provisioner-role
--- a/charts/aws-efs-csi-driver/templates/controller-deployment.yaml
+++ b/charts/aws-efs-csi-driver/templates/controller-deployment.yaml
@@ -23,6 +23,7 @@ spec:
       annotations: {{ toYaml .Values.node.podAnnotations | nindent 8 }}
       {{- end }}
     spec:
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

**Comment by wongma7** [comment]
/ok-to-test

I thought we already had it? Might be confusing it with ebs 🤦

**Comment by wongma7** [comment]
Could you also run `make generate-kustomize` to update the  kustomize tempalte asw well. thanks.

File: deploy/kubernetes/base/controller-serviceaccount.yaml [hunk 1]
```diff
@@ -47,7 +47,7 @@ metadata:
 subjects:
   - kind: ServiceAccount
     name: efs-csi-controller-sa
-    namespace: default
+    namespace: sample
```
wongma7, line 50: this change seems unintentional, i think it will break the kustomize install command we put in the README

wongma7: I will fix it in a future pr, might be that make generate-kustomize has an issue

PR #384, opened 2021-03-22
Revert efs-utils to 1.28.2

wongma7:

**What is this PR about? / Why do we need it?** I am reasonably sure that this is contributing to the high failure rate https://testgrid.k8s.io/provider-aws-efs-csi-driver#e2e-test&width=20. Locally I ran the test and reproduced the write pod hanginga fter the driver pod restarts (it cannot read or write to the efs mount point). I then replaced the driver pod with a 1.1 version and the mount stopped hanging.

**What testing is done?**


nckturner (approved):

wongma7:
@Cappuccinuo I was seeing CI flakiness between driver with this efs-utils version and the 1.1 driver. As there was not much time to debug the root cause for 1.2 release, I just reverted. Will try bumping it back again.

Issue #283, opened 2020-11-19
resource limits in helm values file don'


dmcnaught:
/kind bug

**What happened?**

**What you expected to happen?**

**How to reproduce it (as minimally and precisely as possible)?**

**Anything else we need to know?**:

**Environment**
- Kubernetes version (use `kubectl version`):
- Driver version:

wongma7:
/assign

related to "Very dirty Helm chart " https://github.com/kubernetes-sigs/aws-efs-csi-driver/issues/243

julianguinard:
Is there any update on this issue? It is very important that resources values are taken into account into daemonset and deployment.

The DaemonSet launched without limits from the Chart ended up eating all of our memory on every node on our cluster because of leaky livenessprobe:2.1.0 sidecar (see https://github.com/kubernetes-sigs/aws-efs-csi-driver/issues/341)

diranged:
This is a pretty serious bug... it should be trivial to fix, but given the issues that are all marked stale, I am not inclined to write up a patch unless there is some indication it will actually be accepted and released...

wongma7:
/remove-lifecycle rotten

This should be fixed in the next release, there will be node.resources specifically for the daemonset. https://github.com/kubernetes-sigs/aws-efs-csi-driver/blob/master/charts/aws-efs-csi-driver/values.yaml#L96.

I'll comment + close when it's actually released

iketiunn:
/remove-lifecycle rotten

As the issue https://github.com/kubernetes-sigs/aws-efs-csi-driver/issues/1160 remains unresolved, I'm wondering if it's possible to set resource limits for the `node` as a workaround?

PR #398, opened 2021-03-30
Add resource requests and limits to both DaemonSet and Deployment deployed by the Helm Chart

fkrestan:
bugfix
Fix for #283 
**What testing is done?** 
I've rendered the the Helm with all combinations and ran those against a K8s cluster.


fkrestan:
I apologize, but the [linked CLA signing](https://github.com/kubernetes/community/blob/master/CLA.md#the-contributor-license-agreement) docs seems outdated. [Looking at what seems to bu up-to-date docs](https://docs.linuxfoundation.org/lfx/easycla/contributors/individual-contributor#github) it seems that the CLA process is broken for this repo as the "EasyCLA" button is missing.

fkrestan:
I yield in favor of #406.

PR #410, opened 2021-04-12
Added http_proxy to the helm chart

### LINKED ISSUE #409: Ability to set http proxy in helm chart
The EFS Controller cannot communicate with AWS API when deployed by the helm chart. This is because my organization does not currently deploy VPC endpoints for EFS and requires internet traffic to be routed through a proxy. 

The [EBS CSI Driver Helm Chart](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/charts/aws-ebs-csi-driver/templates/controller.yaml#L97) allows the http proxy for the controller to be set. I would like to be able to do the same thing with the EFS CSI Driver's Helm Chart.

clarkdoan:
This PR adds the ability of setting proxy variables. Mimics the EBS CSI Driver's [implementation](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/charts/aws-ebs-csi-driver/templates/controller.yaml#L97). 

Closes #409 

**What testing is done?** 
This was tested by upgrading a release and by creating a new release.

### DIFF
--- a/charts/aws-efs-csi-driver/values.yaml
+++ b/charts/aws-efs-csi-driver/values.yaml
@@ -52,6 +52,7 @@ controller:
   # path on efs when deleteing an access point
   deleteAccessPointRootDir: false
   podAnnotations: {}
+  proxy: {}
   resources:
     {}
     # We usually recommend not to specify default resources and to leave this as a conscious
--- a/charts/aws-efs-csi-driver/templates/controller-deployment.yaml
+++ b/charts/aws-efs-csi-driver/templates/controller-deployment.yaml
@@ -57,6 +57,14 @@ spec:
           env:
             - name: CSI_ENDPOINT
               value: unix:///var/lib/csi/sockets/pluginproxy/csi.sock
+            {{- if .Values.controller.proxy.http_proxy }}
+            - name: HTTP_PROXY
+              value: {{ .Values.controller.proxy.http_proxy | quote }}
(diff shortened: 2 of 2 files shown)


charts/aws-efs-csi-driver/values.yaml
```diff
@@ -34,6 +34,22 @@ sidecars:
 
 imagePullSecrets: []
 
+proxy: {}
+#  http_proxy:
+#  no_proxy:
+
+resources:
```
wongma7, line 41: think this showed up becasue you had to rebase ,sorry about that, could you remove (the resources got moved to controller.resources so this field isn't used anymore)

danielwegener:
Unfortunately @clarkdoan  does not seem to be available. Would it be (process-wise) possible if I handed in his commit+fix in a different PR?

charts/aws-efs-csi-driver/values.yaml
clarkdoan: @wongma7 I just pushed an update.

danielwegener: @clarkdoan Hate to ask for it, but would you rebase it once more please? :)

danielwegener:
@wongma7 sounds good for me :)

MageshSrinivasulu:
@danielwegener  Is the issue fixed and do we have the proxy added to the helm chart?

Issue #411, opened 2021-04-13
Files not getting deleted on the EFS when PVCs & Access Points are Deleted


MarkSpencerTan:
/kind bug

**What happened?**
I'm using Dynamic Provisioning and I was wondering if it is expected behavior or a bug for my files to still be in the EFS file system even after we have cleaned up our pods and PVC/PVs. This is resulting in our meter usage keep going up even though we are redeploying new pods and the old files from already deleted pods and PVs are still in the file system.

I expected the Reclaim Policy: Delete option of the storage class to be able to remove those files as well, not just the Access Points.

AWS Support suggests that we attach a cleanup sidecar pod to remove the files before it gets deleted, but it would be really nice if that could be handled by the storage class itself to avoid us having to create a separate pod and handle the cleaning part separately.

**What you expected to happen?**
After we delete the PVC/PV that was being used by the pods, it should have deleted the files in the attached Filesystem

**How to reproduce it (as minimally and precisely as possible)?**

1. Use dynamic provisioning option of the storage class with the reclaim policy set to Delete:
```yaml
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  name: efs-sc
provisioner: efs.csi.aws.com
reclaimPolicy: Delete
parameters:
  provisioningMode: efs-ap
  fileSystemId: $EFS_FILE_SYSTEM_ID
  directoryPerms: "700"
```
2. Create a Pod with a PVC that uses this storage class. Make sure the pod creates a file in the volume
3. Delete the Pod and the PVC that was created
4. Mount the EFS Filesystem to an instance that can mount it
5. You'll find the file still exists in the file system under the folder of its previously existing Access Point

**Anything else we need to know?**:

**Environment**
- Kubernetes version (use `kubectl version`): 1.18*
- Driver version: 1.2.0

MarkSpencerTan:
Looking through the code, I found the solution using an argument that's passed into the EFS driver container. I wish this was documented a bit more since it can be misleading for users when it comes to the Delete option of reclaimPolicy.

To enable this functionality here's what I did:

go to the deployment spec of the EFS CSI Driver and add this into the arguments section: `--delete-access-point-root-dir=true`

```yaml
    spec:
      containers:
      - args:
        - --delete-access-point-root-dir=true
        - --endpoint=\$(CSI_ENDPOINT)
        - --logtostderr
        - --v=5
```

chrishenzie:
[...]

PR #412, opened 2021-04-14
feat: add helm config to enable deleteAccessPointRootDir

KarstenSiemer:
This is a new feature. (Also includes very minor fix for the log level in the controller deployment for the log level)
It adds the ability to enable the --delete-access-point-root-dir flag on the controller deployments efs-plugin sidecar.
**What testing is done?** 
Talked about the flag in #411 
I'd suggest to use it like this in helm. If the value is set to false or is omitted, then the flag will be set to false. If the value is set to true, then it will be set to true. This way we don't have to change the chart again, if the default behavior of the flag will be changed, like if it would default to true.
Merged this pr in my fork and rolled it out in a cluster.

### DIFF
--- a/charts/aws-efs-csi-driver/Chart.yaml
+++ b/charts/aws-efs-csi-driver/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: "1.2.0"
 name: aws-efs-csi-driver
 description: A Helm chart for AWS EFS CSI Driver
-version: 1.2.2
+version: 1.2.3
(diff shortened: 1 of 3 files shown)


KarstenSiemer:
lessened coverage makes sense in this case. But do we want to write tests for the helm chart?
How do we want to proceed with this pr?

wongma7:
@KarstenSiemer don't worry about ocveralls, it's bugged sometimes, I will merge once the other Required tests pass.

PR #415, opened 2021-04-19
AWS authentication via credentials file

mrproper:
Feature
Currently the only options for giving the efs driver aws authentication is via using eks with an iam service account or using an instance profile policy.
For those not running EKS an instance profile isnt acceptable without a lot of other plumbing.
So to go around this, i first attempted to use environment variables for credentials which the driver code does not look up but it does look up /root/.aws/credentials

**What testing is done?** 
Tested that it works for creating dynamic efs pvc's on instances without an instance profile


wongma7:
the driver code is using the go sdk which *should* look up environment variables. what environment variables did you try? AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY should be respected

PR #417, opened 2021-04-20
Resolve #416 add reclaim policy to efs storage classes in helm chart

### LINKED ISSUE #416: Please allow me to set the "reclaimPolicy" in custom efs storage classes in the helm chart

I have a use case where custom efs-sc volumes will be created by helm charts and then deleted later on. I need those volumes to be available for reuse at a later time.
The k8 documentation that I've read indicates that the correct way to do this is to set the `ReclaimPolicy` on the storage class.
In the helm chart.

Allow me to set the `reclaimPolicy` in custom storage classes created by the helm chart.

I could patch the deployment manually but this would require me to add another deployment script or external tool into the mix. I am trying to manage everything with helm, and this appears to be a very easy fix.

**Additional context**
https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.21/#storageclass-v1-storage-k8s-io

lukeplausin:

Introducing a new feature - specify the reclaim policy

So that I can set EFS PVs to be retained on pvc deletion instead of cascade deleted

**What testing is done?** 

Compiled the helm chart locally

### DIFF
--- a/charts/aws-efs-csi-driver/templates/storageclass.yaml
+++ b/charts/aws-efs-csi-driver/templates/storageclass.yaml
@@ -8,6 +8,9 @@ provisioner: efs.csi.aws.com
 mountOptions: 
 {{ toYaml . }}
 {{- end }}
+{{- if .reclaimPolicy }}
+reclaimPolicy: {{ .reclaimPolicy }}
+{{- end }}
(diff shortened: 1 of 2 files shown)


wongma7:
/ok-to-test
/check-cla

thanks

PR #418, opened 2021-04-20
Add fargate support in the EFS CSI driver

anonymint:
For driver deployment in Fargate tasks environment

New Feature

Current driver is aiming toward running on EC2 using IMDS which is not available in Fargate, so this PR is to add Fargate support and check if in Fargate environment we can use Task Metadata Endpoint as a way to pull data similar to IMDS.

**What testing is done?** 
* Unit tests added
* Run E2E testing  with success result 
* Manual testing in Fargate environment


pkg/driver/efs_watch_dog.go
```diff
@@ -214,7 +219,9 @@ func (w *execWatchdog) updateConfig(efsClientSource string) error {
 		return fmt.Errorf("cannot create config file %s for efs-utils. Error: %v", w.efsUtilsCfgPath, err)
 	}
 	defer f.Close()
-	efsCfg := efsUtilsConfig{EfsClientSource: efsClientSource}
+	// used on Fargate, IMDS queries suffice otherwise
+	region := os.Getenv("AWS_REGION")
```
wongma7, line 223: @kbasv i wonder if we could add support for querying fargate task metadata to efs-utils? https://github.com/aws/efs-utils/blob/c5abeed1a90a20c0ac0ad1a592296f795cd33e7c/src/mount_efs/__init__.py#L293 will create issue later

pkg/cloud/task_metadata_test.go
```diff
@@ -0,0 +1,97 @@
...
+				if err == nil {
+					t.Fatalf("getTaskMetadata() failed: expected error")
+				}
+			}
+		})
+	}
+}
+
+func TestGetTMDSV4ResponseNonECSEnvironment(t *testing.T) {
+	os.Remove(taskMetadataV4EnvName)
```
wongma7, line 91: my reading is that his will remove the file named 'ECS_CONTAINER_METADATA_URI_V4' , is that correct/ intentional?

anonymint: it's supposed to be `Unsetenv` or set this env value to empty string, i'll fix this in commit

pkg/driver/efs_watch_dog.go
kbasv: Should be possible since efs-utils queries ECS metadata for credentials - https://github.com/aws/efs-utils/blob/c5abeed1a90a20c0ac0ad1a592296f795cd33e7c/src/mount_efs/__init__.py#L442

wongma7: @anonymint the comment "// used on Fargate, IMDS queries suffice otherwise" is not 100% clear, are you saying it is guaranteed that AWS_REGION will be set if this application is run as a fargate task/pod?

anonymint: @wongma7 ythere is another [PR not merged](https://github.com/kubernetes-sigs/aws-efs-csi-driver/pull/227) regarding this and I would change this to be `$AWS_DEFAULT_REGION` as current setup in Fargate today, will update accordingly and send you package that Fargate uses and guarantee to have it. 

on the side note, IMHOM It would be  cleaner to add in attached efs-utils as you mention, there should be similar way/location to pull this region.

wongma7:
/lgtm
/approve

thank you

PR #422, opened 2021-04-22
Add `iam` mount option while deleting Access Point root directory

### LINKED ISSUE #419: Controller doesn't use the iam mount option when deleting the AP root directory
/kind bug

**What happened?**

The controller is unable to mount the EFS volume to delete the AP root directory (when we have passed in `--delete-access-point-root-dir=true`) when the EFS FS has a non-default policy and doesn't allow anonymous mounts. The controller logs this:

```
I0420 20:36:02.323046       1 connection.go:186] GRPC error: rpc error: code = Internal desc = Could not mount "fs-12345678" at "/var/lib/csi/pv/fsap-12345678901234567": mount failed: exit status 32
Mounting command: mount
Mounting arguments: -t efs -o tls fs-12345678 /var/lib/csi/pv/fsap-12345678901234567
Output: Could not start amazon-efs-mount-watchdog, unrecognized init system "aws-efs-csi-dri"
b'mount.nfs4: access denied by server while mounting 127.0.0.1:/'
```

The controller should use the same IAM credentials it uses to create/delete APs to also mount the FS to delete the data.

1. Create an EFS FS with a non-default policy (e.g., set a principal to `{"AWS": "<account_id>"}`
2. Create an IAM role for the controller that gives it full EFS permissions
3. Create a new PVC, let controller create the AP and root directory
4. Delete the PVC, observe the controller fail to mount the directory to clean it up

The mountOptions are hardcoded to just `tls` and don't specify `iam` in https://github.com/kubernetes-sigs/aws-efs-csi-driver/blob/d4f2e0c22bc2af2c910e479d2979c193b6e35ed4/pkg/driver/controller.go#L245-L253

**Environment**
- Kubernetes version (use `kubectl version`): Server Version: version.Info{Major:"1", Minor:"19+", GitVersion:"v1.19.8-eks-96780e", GitCommit:"96780e1b30acbf0a52c38b6030d7853e575bcdf3", GitTreeState:"clean", BuildDate:"2021-03-10T21:32:29Z", GoVersion:"go1.15.8", Compiler:"gc", Platform:"linux/amd64"}
- Driver version: 1.2

kbasv:
Fixes #419 

Adds `iam` to controller mount option. `iam` ensures mounts for file systems with non `anonymous` principles to succeed.

**What testing is done?** 
Tested on my cluster

### DIFF
--- a/pkg/driver/controller.go
+++ b/pkg/driver/controller.go
@@ -242,7 +242,7 @@ func (d *Driver) DeleteVolume(ctx context.Context, req *csi.DeleteVolumeRequest)
 			}
 
 			//Mount File System at it root and delete access point root directory
-			mountOptions := []string{"tls"}
+			mountOptions := []string{"tls", "iam"}
 			target := TempMountPathPrefix + "/" + accessPointId
 			if err := d.mounter.MakeDir(target); err != nil {
 				return nil, status.Errorf(codes.Internal, "Could not create dir %q: %v", target, err)


wongma7:
/ok-to-test

I dont fully understand the issue. If we add iam, and mount.efs fails to find creds, the mount will fail. Why does it need creds?

kbasv:
@wongma7 EFS allows user to define file system policy to prevent `anonymous access`. In such cases, client has to authenticate, otherwise file system access will be denied.

For the node mount, user can supply `iam` option under volume mount options. Additionally, the user has to ensure efs-plugin running on the node has credentials required to authenticate either through IRSA or other mechanisms.

wongma7:
OK. I think i understand now. if anonymous access is allowed, but we supply the iam mount option anyway, the mount wil lstill succeed because the controller must have iam cresd somewhre (even if it doesn't actually need them to do the mount)

PR #425, opened 2021-04-27
Add x-AZ mount support for efs-csi-driver

kbasv:
New feature.

Bump efs-utils to 1.30.1. This adds support for x-AZ mounts. Users can pass desired AZ under `mountOptions` either in PV or Storage class. For example, to mount file system with a mount target in the AZ us-east-1a, add `az=us-east-1a` under mount options.

**What testing is done?** 
Tested on my EKS cluster

### DIFF
--- a/pkg/driver/efs_watch_dog.go
+++ b/pkg/driver/efs_watch_dog.go
@@ -26,9 +26,17 @@ import (
 	"k8s.io/klog"
 )
 
-// https://github.com/aws/efs-utils/blob/v1.28.2/dist/efs-utils.conf
+// https://github.com/aws/efs-utils/blob/v1.30.2/dist/efs-utils.conf
 const (
 	efsUtilsConfigTemplate = `
+#
+# Copyright 2017-2018 Amazon.com, Inc. and its affiliates. All Rights Reserved.
+#
+# Licensed under the MIT License. See the LICENSE accompanying this file
+# for the specific language governing permissions and limitations under
+# the License.
+#
+
 [DEFAULT]
 logging_level = INFO
 logging_max_bytes = 1048576
@@ -37,7 +45,7 @@ logging_file_count = 10
 state_file_dir_mode = 750
 
 [mount]
-dns_name_format = {fs_id}.efs.{region}.{dns_name_suffix}
+dns_name_format = {az}.{fs_id}.efs.{region}.{dns_name_suffix}
 dns_name_suffix = amazonaws.com
 #The region of the file system when mounting from on-premises or cross region.
 {{if .Region -}}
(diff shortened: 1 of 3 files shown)


pkg/driver/efs_watch_dog.go
```diff
@@ -37,14 +45,10 @@ logging_file_count = 10
 state_file_dir_mode = 750
 
 [mount]
-dns_name_format = {fs_id}.efs.{region}.{dns_name_suffix}
+dns_name_format = {az}.{fs_id}.efs.{region}.{dns_name_suffix}
 dns_name_suffix = amazonaws.com
 #The region of the file system when mounting from on-premises or cross region.
-{{if .Region -}}
-region = {{.Region -}}
```
wongma7, line 44: @anonymint will fargate work without this?

kbasv: My bad, I should probably put that back for fargate support

wongma7: yea I assume taht if both region and az are provided then az mount option takes precedence?

kbasv: If both region and az are provided, then efs-utils will use both to resolve dns (see line 48)

wongma7: I see, so someone could provide a nonsense combination . anyway, I think as long as we keep this here then the fargate situation will still work . IIRC the fargate issue is that there is no instance metadata for efs-utils to figure out what region it's in to construct the efs dns name, so we have to explicitly provide it, that's all...

kbasv: There is always a chance that someone can provide a bad combination of az and region. If someone provided a bad combination, efs-utils will fail to resolve dns and throw a mount failure exception.

anonymint: @wongma7 and @kbasv yes Fargate needs that line, probably too late it's back already, all good then.

wongma7:
BTW I want to do a stress test of some kind before we make another release https://github.com/kubernetes-sigs/aws-efs-csi-driver/issues/428.

pkg/driver/efs_watch_dog.go
```diff
@@ -26,9 +26,17 @@ import (
 	"k8s.io/klog"
 )
 
-// https://github.com/aws/efs-utils/blob/v1.28.2/dist/efs-utils.conf
+// https://github.com/aws/efs-utils/blob/v1.30.2/dist/efs-utils.conf
```
Cappuccinuo, line 29: This is the v1.30.2 config file, any reason we don't keep the consistency with the version in DockerFile?

pkg/driver/efs_watch_dog.go
```diff
@@ -248,7 +264,7 @@ func (w *execWatchdog) runLoop(stopCh <-chan struct{}) {
 	for {
 		select {
 		case <-stopCh:
-			klog.Info("stopping...")
+			//klog.Info("stopping...")
```
Cappuccinuo, line 267: Why comment this out?

Cappuccinuo (commented):
Just bump the version? The csi-driver will take care of the option parsing, right? Also why the config file is separate in the efs_watch_dog.go file?

pkg/driver/efs_watch_dog.go
kbasv: Planning to bump efs-utils to 1.30.2 when it is available in linux repo. There isn't too much difference between conf files for 1.30.1 & 1.30.2 so decided to copy over the content of 1.30.2

pkg/driver/efs_watch_dog.go
kbasv: Ah this sneaked in when I was running tests locally, will revert in next commit.

kbasv:
> Just bump the version? The csi-driver will take care of the option parsing, right? Also why the config file is separate in the efs_watch_dog.go file?

Yup bumping efs-utils version is enough for x-AZ. CSI driver will pass the option to efs-utils. 

Regarding separate config file, I believe it was a side effect of trying the preserve the `static` config files between driver restarts to ensure mounts continue to work on driver restarts.

spanktar:
@kbasv How would one utilize this in the Helm chart?

Issue #252, opened 2020-09-13
NodeGetCapabilities logged to stderr instead of stdout


cbighea:
/kind bug

**What happened?**
The aws-efs-csi-driver pods have the following item showing up in stderr in Datadog monitoring, which doesn't seem to be an error:
`I0913 12:09:11.322490       1 node.go:226] NodeGetCapabilities: called with args`

**What you expected to happen?**
I expected this to be logged to stdout

**How to reproduce it (as minimally and precisely as possible)?**
Just followed the normal setup instructions, nothing special.

**Anything else we need to know?**:
I think this output should instead go to `/proc/1/fd/1` though I may be wrong.

**Environment**
- Kubernetes version (use `kubectl version`): 1.17
- Driver version: 1.0.0

keifgwinn:
Does it even need to log this at all? It's the single largest message count on my system, hundreds of thousands of lines of useless logging.

edvsuperbrave:
We get these entries thousands of times in our logs too:

main.go:68] Health check succeeded
connection.go:184] GRPC error: <nil>
connection.go:183] GRPC response: {}
connection.go:181] GRPC request: {}
connection.go:180] GRPC call: /csi.v1.Identity/Probe
main.go:50] Sending probe request to CSI driver "efs.csi.aws.com"

We sent out k8s logs to DataDog which marks it as errors. Probably because they're also sent to stderr. It's useless data and I'd rather see this gone completely.

The process should only output something when the probe request / health check fails. Apart from that, it should keep quiet, IMHO.

It is not only annoying as the logs are littered with incorrect errors, it also adds to the costs of transmitting, parsing. storing and monitoring the logs of our k8s system.

Info on what produces the errors:

kube_daemon_set: efs-csi-node
kube_app_instance: aws-efs-csi-driver
kube_app_name: aws-efs-csi-driver
kube_namespace: kube-system
pod_name: efs-csi-node-qf25c
kube_container_name: liveness-probe

wongma7:
klog https://github.com/kubernetes/klog logs to stderr by default.

I think the issue is the helm chart defaults to loglevel 5 https://github.com/kubernetes-sigs/aws-efs-csi-driver/blob/master/charts/aws-efs-csi-driver/values.yaml#L52 which is too high. These messages are set at level 4/5 which is meant for debugging, the chart doesn't need to default that high

PR #426, opened 2021-04-27
Reduce default log level to 2

wongma7:
**Is this a bug fix or adding new feature?** fix https://github.com/kubernetes-sigs/aws-efs-csi-driver/issues/252

**What is this PR about? / Why do we need it?** 5 is meant for debugging, https://github.com/kubernetes/community/blob/master/contributors/devel/sig-instrumentation/logging.md 

I ran make generate-kustomize.

**What testing is done?**

### DIFF
--- a/deploy/kubernetes/base/controller-deployment.yaml
+++ b/deploy/kubernetes/base/controller-deployment.yaml
@@ -37,9 +37,8 @@ spec:
           args:
             - --endpoint=$(CSI_ENDPOINT)
             - --logtostderr
-            - --v=5
-            # Uncomment below line to allow access point root directory to be deleted by controller.
-            #- --delete-access-point-root-dir
+            - --v=2
+            - --delete-access-point-root-dir=false
           env:
             - name: CSI_ENDPOINT
               value: unix:///var/lib/csi/sockets/pluginproxy/csi.sock
@@ -62,7 +61,7 @@ spec:
           image: public.ecr.aws/eks-distro/kubernetes-csi/external-provisioner:v2.1.1-eks-1-18-2
           args:
             - --csi-address=$(ADDRESS)
-            - --v=5
+            - --v=2
             - --feature-gates=Topology=true
             - --leader-election
           env:
--- a/deploy/kubernetes/base/controller-serviceaccount.yaml
+++ b/deploy/kubernetes/base/controller-serviceaccount.yaml
@@ -47,7 +47,7 @@ metadata:
 subjects:
   - kind: ServiceAccount
     name: efs-csi-controller-sa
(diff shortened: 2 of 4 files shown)


ayberk:
/lgtm

We need this for other CSI drivers as well.

deploy/kubernetes/base/controller-serviceaccount.yaml
```diff
@@ -47,7 +47,7 @@ metadata:
 subjects:
   - kind: ServiceAccount
     name: efs-csi-controller-sa
-    namespace: sample
+    namespace: default
```
ayberk, line 50: Do we want this?

deploy/kubernetes/base/controller-deployment.yaml
```diff
@@ -37,9 +37,8 @@ spec:
           args:
             - --endpoint=$(CSI_ENDPOINT)
             - --logtostderr
-            - --v=5
-            # Uncomment below line to allow access point root directory to be deleted by controller.
-            #- --delete-access-point-root-dir
+            - --v=2
+            - --delete-access-point-root-dir=false
```
ayberk, line 41: Do we want this?

deploy/kubernetes/base/controller-serviceaccount.yaml
wongma7: IDK where it came from. it's make generate-kustomize at work.

I think default makes more sense than sample though?

deploy/kubernetes/base/controller-deployment.yaml
wongma7: make generate-kustomize cleans out comments. It's hard to preserve them. TBH, I think it's better if we try to copy what @AndyXiangLi  did for ebs and document these flags in README than keep comments that won't be noticed https://github.com/kubernetes-sigs/aws-ebs-csi-driver#driver-options

wongma7: I misread, thought you meant just the comment being removed.

about it being false by default:

yes, that's intended, it comes from the helm chart making it false by default.

wongma7: one more comment:

about it being false by default in the helm chart in the first place:

discussion is here: https://github.com/kubernetes-sigs/aws-efs-csi-driver/issues/411

deploy/kubernetes/base/controller-serviceaccount.yaml
ayberk: Yeah agreed. Just wanted to make sure we don't include something we don't want.

PR #435, opened 2021-04-30
Create a new AWS session with the region

### LINKED ISSUE #433: EFS driver doesn't use regional STS endpoint
/kind bug

**What happened?**

I have EKS deployed to a private VPC that does not have direct internet access but does have VPC endpoints created for all necessary services, including STS. For all other controllers I have deployed to this cluster, I have patched the Deployments to include the `AWS_STS_REGIONAL_ENDPOINTS=regional` environment variable which means that instead of using `sts.amazonaws.com`, it uses `sts.<region>.amazonaws.com` which resolves to the local VPC endpoint. 

Deploying the EFS driver along with one of the examples I found it was timing out trying to describe the EFS filesystem. Looking at the DNS traffic from the controller Pod I could see it was still resolving `sts.amazonaws.com` and then failing to connect to it despite passing in the `AWS_STS_REGIONAL_ENDPOINTS` variable. This is before it is even trying to access the EFS API.

It should be resolving and using `sts.<region>.amazonaws.com` which is directly accessible as a VPC endpoint.

Patch the aws-efs-csi-driver container to include the `AWS_STS_REGIONAL_ENDPOINTS=regional` environment variable and watch the DNS traffic. Something like `tcpdump port 53 and host <controller Pod IP>` on the Node running the controller.

I have a PR to fix this.

**Environment**
- Kubernetes version (use `kubectl version`): 1.17
- Driver version: 1.2.0

bodgit:

Fixes #433

Rather than use a region-less session and configure the EFS client with the region from the metadata service, create a new session with the region. This way when `AWS_STS_REGIONAL_ENDPOINTS=regional` is set the session will return the correct STS endpoint.

**What testing is done?** 

With this change, when I deploy the EFS driver to my EKS cluster with the aws-efs-csi-driver container patched to include the `AWS_STS_REGIONAL_ENDPOINTS=regional` environment variable it now works correctly and uses the `sts.<region>.amazonaws.com` endpoint. If I remove the environment variable or set it to `legacy` then it reverts back to using the `sts.amazonaws.com` endpoint which is the original behaviour.

### DIFF
--- a/pkg/cloud/cloud.go
+++ b/pkg/cloud/cloud.go
@@ -95,7 +95,7 @@ func NewCloud() (Cloud, error) {
 		return nil, fmt.Errorf("could not get metadata from AWS: %v", err)
 	}
 
-	efsClient := efs.New(sess, aws.NewConfig().WithRegion(metadata.GetRegion()))
+	efsClient := efs.New(session.Must(session.NewSession(aws.NewConfig().WithRegion(metadata.GetRegion()))))
 	return &cloud{
 		metadata: metadata,
 		efs:      efsClient,


wongma7:
I'm not that well-versed in the quirks of sdk client/session config but this looks right to me and i'll take your word for it. : D

/ok-to-test
/lgtm
/approve

bodgit:
Not sure if those test failures are related to my change or not?

vxst:
Encountered the same problem, the changes lgtm.

I believed perhaps the testsuit should update the env set. 

Should I try to do a commit on this or the testsuit is limited to AWS official?

vxst:
It seems that the default env works.

Issue #436, opened 2021-05-03
Add helm value for toggling hostNetwork true on controller/node


wongma7:
/kind bug

**What happened?** hostNetwork true is not always needed, but we decided to default to it for sake of users whose containers can't access instance metadata without hostNetwork true https://github.com/kubernetes-sigs/aws-efs-csi-driver/issues/313

This unintentionally breaks users whose Nodes have the port 9809 occupied . The point of hostnetwork is not to expose this port on the node but it's what it does and the pod fails if the port is occupied.

**What you expected to happen?**

Add a helm value to toggle it and consider defaulting it false.

**How to reproduce it (as minimally and precisely as possible)?**

**Anything else we need to know?**:

**Environment**
- Kubernetes version (use `kubectl version`):
- Driver version:

wongma7:
Also could make the port configurable as well

PR #437, opened 2021-05-03
Change controller port 9808->9909 to avoid node/ebs conflict

wongma7:

**What is this PR about? / Why do we need it?** 9808 is a poor choice because ebs asks for the same port. Related regarding hostnetwork true: https://github.com/kubernetes-sigs/aws-efs-csi-driver/issues/436

**What testing is done?**


kbasv:
Downgrading efs-utils to v1.28.2 might fix e2e for the time being.

PR #448, opened 2021-05-13
Update access point root directory name to use PV name

kbasv:

 Access point root directory name will be created with PV name. This makes it easy to track access point associated with a PV .

**What testing is done?** 

Tested on EKS cluster.

Before this change, AP root directory looks like:

```bash
efs-csi-ap-1000-72f94a19-4f91-46c0-bd7b-d7acde1d969c
```

After applying this change, AP root directory looks like:

```bash
pvc-0af3bcf3-6f60-4164-8055-88466a239624
```

### DIFF
--- a/pkg/driver/controller.go
+++ b/pkg/driver/controller.go
@@ -23,7 +23,6 @@ import (
 	"strings"
 
 	"github.com/container-storage-interface/spec/lib/go/csi"
-	"github.com/google/uuid"
 	"github.com/kubernetes-sigs/aws-efs-csi-driver/pkg/cloud"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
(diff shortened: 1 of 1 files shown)


pkg/driver/controller.go
```diff
@@ -179,8 +177,7 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
 		return nil, err
 	}
 
-	gidStr := strconv.Itoa(gid)
-	rootDirName := RootDirPrefix + "-" + gidStr + "-" + uuid.New().String()
+	rootDirName := volName
```
wongma7, line 180: this is the PVC name, not PV name I think. PVC name is not unique, at a given time u can have 2 PVC with same name in different namespace. should a unique string still be appended to be safe?

wongma7: nvm, i double checked it is PV name.

wongma7: since we use it as the clienttoken and handle AlreadyExists correctly, this LGTM.

Issue #468, opened 2021-06-02
Decouple EC2 metadata service (IMDS) for on-premise Kubernetes environment


davidshtian:
**Is your feature request related to a problem?/Why is this needed**
Yes, as shown in [issue 449](https://github.com/kubernetes-sigs/aws-efs-csi-driver/issues/449), _efs-csi-controller_ won't start in on-premise Kubernetes environment as the IMDS service access is not available. The root cause is that EFS CSI Controller will [retrieve metadata information](https://github.com/kubernetes-sigs/aws-efs-csi-driver/blob/master/pkg/cloud/metadata.go) here.

The deployment in on-premise environment will throw the errors below:
```
I0519 11:17:33.469123       1 config_dir.go:62] Mounted directories do not exist, creating directory at '/etc/amazon/efs'
F0519 11:17:39.776125       1 driver.go:56] could not get metadata from AWS: EC2 instance metadata is not available
```

**Describe the solution you'd like in detail**
To decouple EC2 metadata service (IMDS) for on-premise Kubernetes environment, so that this will allow users could start _efs-csi-controller_ normally. And when decouple EC2 metadata service (IMDS), the solution could be leveraging Kubernetes [environment variables](https://kubernetes.io/docs/tasks/inject-data-application/define-environment-variable-container/), 
 feature where users could provide the region and availability zone information that the controller needs.

**Describe alternatives you've considered**
No alternative solution yet.

**Additional context**
I've already tried remove the dependency of EC2 metadata service (IMDS) and re-built the project, then use the Dockerfile to build a new image, and I tested adding environment variables (instanceID, region, availabilityZone) , _efs-csi-controller_ will start normally.

And after successful deployment, there are also errors thrown when mounting EFS file system. The binary entrypoint _aws-efs-csi-driver_ would dynamically generate configuration file, but the region information need to specify in the conf file which will be override by _aws-efs-csi-driver_. I've tried [PostStart container lifecycle hooks](https://kubernetes.io/docs/concepts/containers/container-lifecycle-hooks/) to override the configuration file and fixed the issue.

Hope this feature request could be added so that other on-premise Kubernetes users will take the advantage of EFS service. Thanks~

PR #469, opened 2021-06-02
Add support for on-premise Kubernetes environment

davidshtian:
Yes, this is add feature for supporting on-premise Kubernetes environment to deploy Amazon EFS CSI Driver, and this is related to [issue 468](https://github.com/kubernetes-sigs/aws-efs-csi-driver/issues/468).

As on-premise Kubernetes users also would like to leverage Amazon EFS service for shared file system, while currently there are several issues when deploying to on-premise environment, like EC2 metadata service dependency.

**What testing is done?** 
I've tested using actual on-premise Kubernetes environment.

### DIFF
--- a/deploy/kubernetes/base/controller-deployment.yaml
+++ b/deploy/kubernetes/base/controller-deployment.yaml
@@ -42,6 +42,12 @@ spec:
           env:
             - name: CSI_ENDPOINT
               value: unix:///var/lib/csi/sockets/pluginproxy/csi.sock
+            - name: instanceID
+              value: <change to your mock instanceID>
+            - name: region
+              value: <change to your mock region>
+            - name: availabilityZone
+              value: <change to your mock availabilityZone>
           volumeMounts:
             - name: socket-dir
               mountPath: /var/lib/csi/sockets/pluginproxy/
--- a/deploy/kubernetes/base/controller-deployment.yaml
+++ b/deploy/kubernetes/base/controller-deployment.yaml
@@ -42,12 +42,6 @@ spec:
           env:
             - name: CSI_ENDPOINT
               value: unix:///var/lib/csi/sockets/pluginproxy/csi.sock
-            - name: instanceID
-              value: <change to your mock instanceID>
-            - name: region
-              value: <change to your mock region>
-            - name: availabilityZone
-              value: <change to your mock availabilityZone>
           volumeMounts:
             - name: socket-dir
               mountPath: /var/lib/csi/sockets/pluginproxy/
--- a/docs/ON-PREMISE.md
+++ b/docs/ON-PREMISE.md
@@ -1,0 +1,58 @@
+## Notes for on-premise Kubernetes environment
+
+### Decouple EC2 metadata service (IMDS)
+Since on-premise Kubernetes environment cannot access Amazon EC2 metadata service and cannot get information about instanceID, region and availabilityZone, additional environment variables need to be set, otherwise it will throw "could not get metadata from AWS: EC2 instance metadata is not available" described in [issue 468](https://github.com/kubernetes-sigs/aws-efs-csi-driver/issues/468).
+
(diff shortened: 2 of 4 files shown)


wongma7:
/ok-to-test

Over at https://github.com/kubernetes-sigs/aws-ebs-csi-driver we have been working on decoupling driver from instance metadata as well. And I was planning to port the solution here. The solution over there is to check, in order:
1. EC2 instance metadata
2. Kubernetes API (fall back to this because the node object has instance id, region, zone in spec.providerID)

Do your Nodes have spec.providerID set? If so, the solution could work here too.

Still, there is value in adding environment variable overrides in case spec.providerID is not set or kubernetes API access is blocked from CSI node pods.

The order could be something like:

1. Environment variable override (this PR). 
2. ECS/fargate instance metadata
3. EC2 instance metadata
4. Kubernetes API (the node object has instance id, region, zone in spec.providerID)

Let me know what you think.

davidshtian:
> /ok-to-test
> 
> Over at https://github.com/kubernetes-sigs/aws-ebs-csi-driver we have been working on decoupling driver from instance metadata as well. And I was planning to port the solution here. The solution over there is to check, in order:
> 
> 1. EC2 instance metadata
> 2. Kubernetes API (fall back to this because the node object has instance id, region, zone in spec.providerID)
> 
> Do your Nodes have spec.providerID set? If so, the solution could work here too.
> 
> Still, there is value in adding environment variable overrides in case spec.providerID is not set or kubernetes API access is blocked from CSI node pods.
> 
> The order could be something like:
> 
> 1. Environment variable override (this PR).
> 2. ECS/fargate instance metadata
> 3. EC2 instance metadata
> 4. Kubernetes API (the node object has instance id, region, zone in spec.providerID)
> 
> Let me know what you think.

Hi~ The check order looks great! I've checked that there is no "spec.providerID" in on premise environment, it seems to be set by Cloud Controller Manager. Thanks~

CNLHC:
> /ok-to-test
> 
> Over at https://github.com/kubernetes-sigs/aws-ebs-csi-driver we have been working on decoupling driver from instance metadata as well. And I was planning to port the solution here. The solution over there is to check, in order:
> 
> 1. EC2 instance metadata
> 2. Kubernetes API (fall back to this because the node object has instance id, region, zone in spec.providerID)
> 
> Do your Nodes have spec.providerID set? If so, the solution could work here too.
> 
> Still, there is value in adding environment variable overrides in case spec.providerID is not set or kubernetes API access is blocked from CSI node pods.
> 
> The order could be something like:
> 
> 1. Environment variable override (this PR).
> 2. ECS/fargate instance metadata
> 3. EC2 instance metadata
> 4. Kubernetes API (the node object has instance id, region, zone in spec.providerID)
> 
> Let me know what you think.

Hello @wongma7 
It seems that this feature does not gain some progress yet.  Are you still working on this feature? If you are too busy to work on this I am glad to sign the CLA and develop this.

jonathanrainer:
> > /ok-to-test
> 
> > 
> 
> > Over at https://github.com/kubernetes-sigs/aws-ebs-csi-driver we have been working on decoupling driver from instance metadata as well. And I was planning to port the solution here. The solution over there is to check, in order:
> 
> > 
> 
> > 1. EC2 instance metadata
> 
> > 2. Kubernetes API (fall back to this because the node object has instance id, region, zone in spec.providerID)
> 
> > 
> 
> > Do your Nodes have spec.providerID set? If so, the solution could work here too.
> 
> > 
> 
> > Still, there is value in adding environment variable overrides in case spec.providerID is not set or kubernetes API access is blocked from CSI node pods.
> 
> > 
> 
> > The order could be something like:
> 
> > 
> 
> > 1. Environment variable override (this PR).
> 
> > 2. ECS/fargate instance metadata
> 
> > 3. EC2 instance metadata
> 
> > 4. Kubernetes API (the node object has instance id, region, zone in spec.providerID)
> 
> > 
> 
> > Let me know what you think.
> 
> 
> 
> Hello @wongma7 
> 
> It seems that this feature does not gain some progress yet.  Are you still working on this feature? If you are too busy to work on this I am glad to sign the CLA and develop this.

Hi, I have a PR open for this in #681. Would very much appreciate a review :)

PR #470, opened 2021-06-03
Add support for cross account mount

kbasv:

New Feature

### Support for cross-account mount.

A new mount option was introduced to [mount-helper](https://github.com/aws/efs-utils) `mounttargetip`. This option allows users to pass the ip address of the desired mount target of the efs file system. This option also facilitates cross account mount support provided [vpc-peering](https://docs.aws.amazon.com/vpc/latest/peering/working-with-vpc-peering.html) has been established.

This PR adds cross account support using above mount option.

Pre-requisite steps to perform cross account mounts with EKS cluster in Account `A` & EFS in Account `B`:

1.  Perform [vpc-peering](https://docs.aws.amazon.com/vpc/latest/peering/working-with-vpc-peering.html) between EKS cluster `vpc` and EFS `vpc`.
2.  Create an IAM role in Account `B` which has a trust relationship with Account `A` and add an EFS policy with permissions to call `DescribeMountTargets`. This role will be used by CSI-Driver to determine the mount targets for file system in account `B`
3. Create kubenetes secret with `awsRoleArn` as the key and the role from step 2 as the value. For example, `kubectl create secret generic x-account --namespace=default --from-literal=awsRoleArn='arn:aws:iam::1234567890:role/EFSCrossAccountAccessRole'`
4. Create a service account with IAM role for the EKS cluster and attach it node-deamonset. Attach this IAM role with EFS client mount permission policy.
5. Create a [file system policy](https://docs.aws.amazon.com/efs/latest/ug/iam-access-control-nfs-efs.html#file-sys-policy-examples) for file system in account `B` which allows account `A` to perform mount on it.

Additionally, a new storage class parameter `az` allows users to specify preferred mount target for mount. If preferred AZ is unavailable, a random AZ will chosen.

Sample Storage class:

```bash
kind: StorageClass
apiVersion: storage.k8s.io/v1
metadata:
  name: efs-sc
provisioner: efs.csi.aws.com
mountOptions:
  - iam
  - az=us-east-1f # optional, allows cross AZ mount
parameters:
  provisioningMode: efs-ap
  fileSystemId: fs-01234abcd
  directoryPerms: "700"
  az: us-east-1f # optional, used by controller to provide a preferred mount-target AZ
  csi.storage.k8s.io/provisioner-secret-name: x-account # optional, required for cross-account mount. Should be unique per storage class
  csi.storage.k8s.io/provisioner-secret-namespace: kube-system # option, required for cross-account mount
```

**What testing is done?** 

Tested by spinning up EKS cluster in account `A` and performing cross account mount on file systems in different accounts.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -31,18 +31,26 @@ ENV EFS_CLIENT_SOURCE=$client_source
 
 RUN GOOS=${TARGETOS} GOARCH=${TARGETARCH} make aws-efs-csi-driver
 
-FROM amazonlinux:2.0.20210219.0
+FROM amazonlinux:2
 # Install efs-utils from github by default. It can be overriden to `yum` with --build-arg when building the Docker image.
 # If value of `EFSUTILSSOURCE` build arg is overriden with `yum`, docker will install efs-utils from Amazon Linux 2's yum repo.
 ARG EFSUTILSSOURCE=github
 RUN if [ "$EFSUTILSSOURCE" = "yum" ]; \
     then echo "Installing efs-utils from Amazon Linux 2 yum repo" && \
-         yum -y install amazon-efs-utils-1.30.1-1.amzn2.noarch; \
-    else echo "Installing efs-utils from github" && \
+         yum -y install amazon-efs-utils-1.31.1-1.amzn2.noarch; \
+    else echo "Installing efs-utils from github using the latest git tag" && \
          yum -y install git rpm-build make && \
          git clone https://github.com/aws/efs-utils && \
-         cd efs-utils && make rpm && yum -y install build/amazon-efs-utils*rpm; \
+         cd efs-utils && \
+         git checkout $(git describe --tags $(git rev-list --tags --max-count=1)) && \
+         make rpm && yum -y install build/amazon-efs-utils*rpm; \
     fi
+
+# Install botocore required by efs-utils for cross account mount
+RUN yum -y install wget
+RUN wget https://bootstrap.pypa.io/get-pip.py -O /tmp/get-pip.py
+RUN python3 /tmp/get-pip.py
+RUN pip3 install botocore || /usr/local/bin/pip3 install botocore
 
 # At image build time, static files installed by efs-utils in the config directory, i.e. CAs file, need
 # to be saved in another place so that the other stateful files created at runtime, i.e. private key for
--- a/charts/aws-efs-csi-driver/templates/controller-serviceaccount.yaml
+++ b/charts/aws-efs-csi-driver/templates/controller-serviceaccount.yaml
@@ -40,6 +40,9 @@ rules:
   - apiGroups: ["coordination.k8s.io"]
     resources: ["leases"]
     verbs: ["get", "watch", "list", "delete", "update", "create"]
+  - apiGroups: [ "" ]
+    resources: [ "secrets" ]
+    verbs: [ "get", "watch", "list" ]
 
 ---
 
--- a/pkg/cloud/cloud.go
+++ b/pkg/cloud/cloud.go
@@ -20,13 +20,15 @@ import (
 	"context"
 	"errors"
 	"fmt"
-
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/awserr"
+	"github.com/aws/aws-sdk-go/aws/credentials/stscreds"
 	"github.com/aws/aws-sdk-go/aws/request"
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/efs"
 	"k8s.io/klog"
+	"math/rand"
+	"time"
 )
 
 const (
@@ -65,12 +67,20 @@ type AccessPointOptions struct {
 	Tags           map[string]string
 }
 
+type MountTarget struct {
+	AZName        string
+	AZId          string
+	MountTargetId string
+	IPAddress     string
+}
+
 // Efs abstracts efs client(https://docs.aws.amazon.com/sdk-for-go/api/service/efs/)
 type Efs interface {
 	CreateAccessPointWithContext(aws.Context, *efs.CreateAccessPointInput, ...request.Option) (*efs.CreateAccessPointOutput, error)
 	DeleteAccessPointWithContext(aws.Context, *efs.DeleteAccessPointInput, ...request.Option) (*efs.DeleteAccessPointOutput, error)
 	DescribeAccessPointsWithContext(aws.Context, *efs.DescribeAccessPointsInput, ...request.Option) (*efs.DescribeAccessPointsOutput, error)
 	DescribeFileSystemsWithContext(aws.Context, *efs.DescribeFileSystemsInput, ...request.Option) (*efs.DescribeFileSystemsOutput, error)
+	DescribeMountTargetsWithContext(aws.Context, *efs.DescribeMountTargetsInput, ...request.Option) (*efs.DescribeMountTargetsOutput, error)
 }
 
 type Cloud interface {
@@ -79,6 +89,7 @@ type Cloud interface {
 	DeleteAccessPoint(ctx context.Context, accessPointId string) (err error)
 	DescribeAccessPoint(ctx context.Context, accessPointId string) (accessPoint *AccessPoint, err error)
 	DescribeFileSystem(ctx context.Context, fileSystemId string) (fs *FileSystem, err error)
+	DescribeMountTargets(ctx context.Context, fileSystemId, az string) (fs *MountTarget, err error)
 }
 
 type cloud struct {
@@ -96,6 +107,23 @@ func NewCloud() (Cloud, error) {
 	}
 
 	efsClient := efs.New(session.Must(session.NewSession(aws.NewConfig().WithRegion(metadata.GetRegion()))))
+	return &cloud{
+		metadata: metadata,
+		efs:      efsClient,
+	}, nil
+}
+
+// NewCloudWithRole returns a new instance of AWS cloud after assuming an aws role
+// It panics if driver does not have permissions to assume role.
+func NewCloudWithRole(awsRoleArn string) (Cloud, error) {
+	sess := session.Must(session.NewSession(&aws.Config{}))
+	metadata, err := NewMetadataService(sess)
+	if err != nil {
+		return nil, fmt.Errorf("Could not get metadata from AWS: %v", err)
+	}
+
+	creds := stscreds.NewCredentials(sess, awsRoleArn)
+	efsClient := efs.New(session.Must(session.NewSession(aws.NewConfig().WithCredentials(creds).WithRegion(metadata.GetRegion()))))
 	return &cloud{
 		metadata: metadata,
 		efs:      efsClient,
@@ -187,7 +215,7 @@ func (c *cloud) DescribeAccessPoint(ctx context.Context, accessPointId string) (
 
 func (c *cloud) DescribeFileSystem(ctx context.Context, fileSystemId string) (fs *FileSystem, err error) {
 	describeFsInput := &efs.DescribeFileSystemsInput{FileSystemId: &fileSystemId}
-	klog.V(5).Infof("Calling DescribeFS with input: %+v", *describeFsInput)
+	klog.V(5).Infof("Calling DescribeFileSystems with input: %+v", *describeFsInput)
 	res, err := c.efs.DescribeFileSystemsWithContext(ctx, describeFsInput)
 	if err != nil {
 		if isAccessDenied(err) {
@@ -196,7 +224,7 @@ func (c *cloud) DescribeFileSystem(ctx context.Context, fileSystemId string) (fs
 		if isFileSystemNotFound(err) {
 			return nil, ErrNotFound
 		}
-		return nil, fmt.Errorf("describe File System failed: %v", err)
+		return nil, fmt.Errorf("Describe File System failed: %v", err)
 	}
 
 	fileSystems := res.FileSystems
@@ -205,6 +233,52 @@ func (c *cloud) DescribeFileSystem(ctx context.Context, fileSystemId string) (fs
 	}
 	return &FileSystem{
 		FileSystemId: *res.FileSystems[0].FileSystemId,
+	}, nil
+}
+
+func (c *cloud) DescribeMountTargets(ctx context.Context, fileSystemId, azName string) (fs *MountTarget, err error) {
+	describeMtInput := &efs.DescribeMountTargetsInput{FileSystemId: &fileSystemId}
+	klog.V(5).Infof("Calling DescribeMountTargets with input: %+v", *describeMtInput)
+	res, err := c.efs.DescribeMountTargetsWithContext(ctx, describeMtInput)
+	if err != nil {
+		if isAccessDenied(err) {
+			return nil, ErrAccessDenied
+		}
+		if isFileSystemNotFound(err) {
+			return nil, ErrNotFound
+		}
+		return nil, fmt.Errorf("Describe Mount Targets failed: %v", err)
+	}
+
+	mountTargets := res.MountTargets
+	if len(mountTargets) == 0 {
+		return nil, fmt.Errorf("Cannot find mount targets for file system %v. Please create mount targets for file system.", fileSystemId)
+	}
+
+	availableMountTargets := getAvailableMountTargets(mountTargets)
+
+	if len(availableMountTargets) == 0 {
+		return nil, fmt.Errorf("No mount target for file system %v is in available state. Please retry in 5 minutes.", fileSystemId)
+	}
+
+	var mountTarget *efs.MountTargetDescription
+	if azName != "" {
+		mountTarget = getMountTargetForAz(availableMountTargets, azName)
+	}
+
+	// Pick random Mount target from available mount target if azName is not provided.
+	// Or if there is no mount target matching azName
+	if mountTarget == nil {
+		klog.Infof("Picking a random mount target from available mount target")
+		rand.Seed(time.Now().Unix())
+		mountTarget = availableMountTargets[rand.Intn(len(availableMountTargets))]
+	}
+
+	return &MountTarget{
+		AZName:        *mountTarget.AvailabilityZoneName,
+		AZId:          *mountTarget.AvailabilityZoneId,
+		MountTargetId: *mountTarget.MountTargetId,
+		IPAddress:     *mountTarget.IpAddress,
 	}, nil
 }
 
@@ -246,4 +320,25 @@ func parseEfsTags(tagMap map[string]string) []*efs.Tag {
 		})
 	}
 	return efsTags
+}
+
+func getAvailableMountTargets(mountTargets []*efs.MountTargetDescription) []*efs.MountTargetDescription {
+	availableMountTargets := []*efs.MountTargetDescription{}
+	for _, mt := range mountTargets {
+		if *mt.LifeCycleState == "available" {
+			availableMountTargets = append(availableMountTargets, mt)
+		}
+	}
+
+	return availableMountTargets
+}
+
+func getMountTargetForAz(mountTargets []*efs.MountTargetDescription, azName string) *efs.MountTargetDescription {
+	for _, mt := range mountTargets {
+		if *mt.AvailabilityZoneName == azName {
+			return mt
+		}
+	}
+	klog.Infof("There is no mount target match %v", azName)
+	return nil
 }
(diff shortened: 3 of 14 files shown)


pkg/cloud/cloud_test.go
```diff
@@ -596,3 +596,213 @@ func TestDescribeFileSystem(t *testing.T) {
 		t.Run(tc.name, tc.testFunc)
 	}
 }
+
+func TestDescribeMountTargets(t *testing.T) {
```
wongma7, line 600: wdyt about refactoring to be table driven, I want to stop the "testFunc" pattern, it is verbose and harder to read in my opinion compared to providing input, output, expected error

pkg/cloud/cloud.go
```diff
@@ -102,6 +113,23 @@ func NewCloud() (Cloud, error) {
 	}, nil
 }
 
+// NewCloudWithRole returns a new instance of AWS cloud after assuming an aws role
+// It panics if session is invalid
```
wongma7, line 117: panic sounds wrong? if we don't have permission to assume the role it panics?

Dockerfile
```diff
@@ -31,19 +31,25 @@ ENV EFS_CLIENT_SOURCE=$client_source
 
 RUN GOOS=${TARGETOS} GOARCH=${TARGETARCH} make aws-efs-csi-driver
 
-FROM amazonlinux:2.0.20210219.0
+FROM amazonlinux:2.0.20210421.0
 # Install efs-utils from github by default. It can be overriden to `yum` with --build-arg when building the Docker image.
 # If value of `EFSUTILSSOURCE` build arg is overriden with `yum`, docker will install efs-utils from Amazon Linux 2's yum repo.
 ARG EFSUTILSSOURCE=github
```
Cappuccinuo, line 37: Should change it back to yum since 1.31.1 is already present in the yum repo of amazon linux?

Did we consider the logic that if the version is not present in yum repo, then we fallback to use the installation of Github?

charts/aws-efs-csi-driver/templates/controller-serviceaccount.yaml
```diff
@@ -40,6 +40,9 @@ rules:
   - apiGroups: ["coordination.k8s.io"]
     resources: ["leases"]
     verbs: ["get", "watch", "list", "delete", "update", "create"]
+  - apiGroups: [ "" ]
+    resources: [ "secrets" ]
+    verbs: [ "get", "watch", "list" ]
```
Cappuccinuo, line 45: What's the effect of these configurations? Why the apiGroups are empty?

pkg/cloud/cloud.go
```diff
@@ -79,6 +89,7 @@ type Cloud interface {
 	DeleteAccessPoint(ctx context.Context, accessPointId string) (err error)
 	DescribeAccessPoint(ctx context.Context, accessPointId string) (accessPoint *AccessPoint, err error)
 	DescribeFileSystem(ctx context.Context, fileSystemId string) (fs *FileSystem, err error)
+	DescribeMountTargets(ctx context.Context, fileSystemId, az string) (fs *MountTarget, err error)
```
Cappuccinuo, line 92: What's the difference between this DescribeMountTargets and the DescribeMountTargets in efs-utils itself? Is this one used to retrieve mounttargetip address and use the mounttargetip option to mount?

pkg/cloud/cloud.go
```diff
@@ -102,6 +113,23 @@ func NewCloud() (Cloud, error) {
 	}, nil
 }
 
+// NewCloudWithRole returns a new instance of AWS cloud after assuming an aws role
+// It panics if session is invalid
+func NewCloudWithRole(awsRoleArn string) (Cloud, error) {
+	sess := session.Must(session.NewSession(&aws.Config{}))
+	metadata, err := NewMetadataService(sess)
+	if err != nil {
+		return nil, fmt.Errorf("could not get metadata from AWS: %v", err)
```
Cappuccinuo, line 122: nit: s/could/Could

pkg/cloud/cloud.go
```diff
@@ -102,6 +113,23 @@ func NewCloud() (Cloud, error) {
 	}, nil
 }
 
+// NewCloudWithRole returns a new instance of AWS cloud after assuming an aws role
+// It panics if session is invalid
+func NewCloudWithRole(awsRoleArn string) (Cloud, error) {
+	sess := session.Must(session.NewSession(&aws.Config{}))
+	metadata, err := NewMetadataService(sess)
```
Cappuccinuo, line 120: So in csi-driver world, we need the metadata service to determine the region? Is there any other usage for this metadata?

pkg/cloud/cloud.go
```diff
@@ -208,6 +236,51 @@ func (c *cloud) DescribeFileSystem(ctx context.Context, fileSystemId string) (fs
 	}, nil
 }
 
+func (c *cloud) DescribeMountTargets(ctx context.Context, fileSystemId, azName string) (fs *MountTarget, err error) {
+	describeMtInput := &efs.DescribeMountTargetsInput{FileSystemId: &fileSystemId}
+	klog.V(5).Infof("Calling DescribeMountTargets with input: %+v", *describeMtInput)
```
Cappuccinuo, line 241: This is info level log? Should it be a debug level?

pkg/cloud/cloud.go
```diff
@@ -247,3 +320,24 @@ func parseEfsTags(tagMap map[string]string) []*efs.Tag {
...
+
+func getMountTargetForAz(mountTargets []*efs.MountTargetDescription, azName string) *efs.MountTargetDescription {
+	for _, mt := range mountTargets {
+		if *mt.AvailabilityZoneName == azName {
+			return mt
+		}
+	}
+	klog.Infof("There is no mount target match %v. Will pick a random mount target.", azName)
+	return nil
+}
```
Cappuccinuo, line 343: Any idea how we can avoid to implement the same logic on both efs-utils side and csi-driver side?

pkg/driver/controller.go
```diff
@@ -76,7 +79,10 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
 		gidMax           int
 		basePath         string
 		provisioningMode string
+		roleArn          string
+		azName           string
 		err              error
+		localCloud       cloud.Cloud
```
Cappuccinuo, line 85: nit: make all the var in alphabetical order

pkg/driver/controller.go
```diff
@@ -162,8 +167,28 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
 		basePath = value
 	}
 
+	if value, ok := volumeParams[AzName]; ok {
+		azName = value
+	}
+
+	// Fetch aws role ARN for cross account mount from secrets
```
Cappuccinuo, line 174: I feel like this part needs more commend, how/where the cross-account mount credential is retrieved and what is secrets?

pkg/driver/controller.go
```diff
@@ -162,8 +167,28 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
...
+	secrets := req.GetSecrets()
+
+	if value, ok := secrets[RoleArn]; ok {
+		roleArn = value
+	}
+
+	if roleArn != "" {
+		localCloud, err = cloud.NewCloudWithRole(roleArn)
+		if err != nil {
+			return nil, status.Errorf(codes.Unauthenticated, "Unable to initialize aws cloud: %v. Please verify role has the correct AWS permissions for cross account mount", err)
```
Cappuccinuo, line 184: Where should the roleArn be configured? in .config file?

pkg/driver/controller.go
```diff
@@ -196,16 +221,50 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
...
 
+	volContext := map[string]string{}
+
+	// Fetch mount target Ip for cross-account mount
+	if roleArn != "" {
+		mountTarget, err := localCloud.DescribeMountTargets(ctx, accessPointsOptions.FileSystemId, azName)
+		if err != nil {
+			klog.Warningf("Failed to describe mount targets for file system %v. Skip using `mounttargetip` mount option: %v", accessPointsOptions.FileSystemId, err)
+		} else {
+			volContext[MountTargetIp] = mountTarget.IPAddress
```
Cappuccinuo, line 232: What if we don't set this mounttarget ip address and just rely on efs-utils to find the mounttarget? Will that fail?

Dockerfile
```diff
@@ -31,19 +31,25 @@ ENV EFS_CLIENT_SOURCE=$client_source
 
 RUN GOOS=${TARGETOS} GOARCH=${TARGETARCH} make aws-efs-csi-driver
 
-FROM amazonlinux:2.0.20210219.0
+FROM amazonlinux:2.0.20210421.0
```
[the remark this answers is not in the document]
kbasv: I remember it used to be `FROM amazonlinux:2`. I don't remember why we changed to release versions. @wongma7  do you know why?

Dockerfile
kbasv: Since it takes a while for new `efs-utils` release to show up on yum, I think we should install it directly from github. 

Also installing from github always ensures latest efs-utils version is installed. We could avoid bumping up the `efs-utils` release version everytime a new version is released.

charts/aws-efs-csi-driver/templates/controller-serviceaccount.yaml
kbasv: Empty apiGroups indicates it applies to core k8s APIs. 

This RBAC rule authorizes a service account associated with controller to read k8s [secret](https://kubernetes.io/docs/concepts/configuration/secret/) objects

pkg/cloud/cloud.go
kbasv: There is no difference, they are the same APIs.

pkg/cloud/cloud.go
kbasv: Agreed, I'll update the comment accordingly

pkg/cloud/cloud.go
kbasv: AFAIK, metadata service is being used to get region and instance Id (node Id).

pkg/cloud/cloud.go
kbasv: There isn't a specific "debug level". You can increase/decrease verbosity of your logs using klogs's log level.

pkg/cloud/cloud.go
kbasv: One possible way is for efs-utils to accept `role` as a mount option.

pkg/cloud/cloud_test.go
kbasv: Yeah I think table driven approach reduces the verbose nature of the these tests. I'll refactor these tests.

pkg/driver/controller.go
kbasv: I added information on how to setup the secrets in the description of the PR.

You'd need to pass the secret into storage class parameter and [external-provisioner](https://github.com/kubernetes-csi/external-provisioner) will pull the information and adds it to `CreateVolume` request.

kbasv: I plan to add some documentation on readMe on how it works.

pkg/driver/controller.go
kbasv: The roleArn should be provided via secret

pkg/driver/controller.go
kbasv: Yeah that would fail since mount happens on `node-deamonset` which is a separate process. 

Since you cannot setup config/credential files on the driver pod, driver pod will not know how to switch credentials for different AWS accounts and will utilize the credentials from default instance role or service account.

Dockerfile
wongma7: as a middleground, we could checkout latest github tag/version instead of building master branch. something like https://stackoverflow.com/questions/17414104/git-checkout-latest-tag/17414170.

Cappuccinuo: +1

pkg/driver/controller.go
Cappuccinuo: What is secret? Is that some specific configuration for csi-driver?

pkg/cloud/cloud.go
Cappuccinuo: I guess comparing with this approach, adding a role option should be preferred, that will reduce the dup logic in the driver side, also on efs-utils side we provide the output & log message like the ip cannot be connected yet, no available mount target in the AZ, we can reuse it here.

So is this a temp workaround? Should we deprecate the duplicates once the role arn option can be used?

kbasv: While `role` mount option is preferred to reduce the dup logic, I'm not sure if csi-driver can capture all the messages from efs-utils. CSI driver only captures the message(s) shown on the console after mount attempt and not messages logged in `mount.log`. I feel it is an overhead to reach mount.log and would rather throw the messages in the driver logs.

What do you think?

pkg/driver/controller.go
kbasv: Secret is an object which lets you store information, in this case we rely on users to configure roleArn through secrets.

K8s allows secrets to be accessed by a pod in multiple ways. In our context, if a user provides inputs secret names & namespace into the storage class, [external-provisioner](https://kubernetes-csi.github.io/docs/external-provisioner.html) will extract it and pass it on to csi-driver's Controller service `rpc`, provided it has right RBAC rule to access the secret

More info on how secrets work for CSI operations: https://kubernetes-csi.github.io/docs/secrets-and-credentials.html#csi-driver-secrets

Dockerfile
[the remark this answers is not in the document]
Cappuccinuo: @wongma7 Can we just use `FROM amazonlinux:2`? Or there is some specific reason to use the specific tag.

pkg/cloud/cloud.go
Cappuccinuo: The mount attempt will output the message to the console, e.g. https://github.com/aws/efs-utils/blob/master/src/mount_efs/__init__.py#L1789-L1790 here we use fatal_error to throw back and output the error message.

It will be hard if we change the common logic in two place in the future, also if we add a logic, we need to implement the same logic in the driver, that will be not easy to maintain, and csi-driver should be mainly responsible for managing. 

So IMO for this feature we can use the logic here, once we add the role feature, we should explore to optimize the code path and reduce the common logic as much as possible. Let efs-utils handle most of the mount work and logic.

Cappuccinuo (approved):

Dockerfile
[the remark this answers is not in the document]
wongma7: @Cappuccinuo yes we can, let's use amazonlinux:2. I think originally we were having trouble tracking versions so we pinned, but, pinning has the downside that we have to bump every time we build, and if we neglect to bump we become outdated and get hit by CVEs etc.

pkg/cloud/cloud.go
wongma7: it may also be used for credentials since it's in the default chain. 

we do eventually want to make it optional though, some efforts from ebs we can copy soon. i.e., region and instance id can be determined by reading the Node object in Kubernetes API. creds can be acquired from secret or, in EKS, IAM for ServiceAccounts

pkg/cloud/cloud.go
```diff
@@ -247,3 +320,24 @@ func parseEfsTags(tagMap map[string]string) []*efs.Tag {
...
+	return availableMountTargets
+}
+
+func getMountTargetForAz(mountTargets []*efs.MountTargetDescription, azName string) *efs.MountTargetDescription {
+	for _, mt := range mountTargets {
+		if *mt.AvailabilityZoneName == azName {
+			return mt
+		}
+	}
+	klog.Infof("There is no mount target match %v. Will pick a random mount target.", azName)
```
wongma7, line 341: this message could be moved to where the random mount target is picked (the caller)

pkg/driver/controller.go
wongma7: if roleArn is empty, AzName is ignored. should AzName be respected for cross-AZ mount? Trying to make sense of all scenarios' behavior and make sure we don't break the current behavior. If neither roleArn nor az are provided, then efs-utils should find the mount target.

cross-AZ + cross-Account (both az and roleArn have been specified) 
cross-AZ + same-Account (only az has been specified)
same-AZ + cross-Account (only roleArn has been specified)
same-AZ + same-Account (neither have been specified)

pkg/driver/controller.go
```diff
@@ -196,16 +222,50 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
...
 
 func (d *Driver) DeleteVolume(ctx context.Context, req *csi.DeleteVolumeRequest) (*csi.DeleteVolumeResponse, error) {
+	var (
+		localCloud cloud.Cloud
+		roleArn    string
+		err        error
+	)
+
+	// Fetch aws role ARN for cross account mount from secrets
+	secrets := req.GetSecrets()
```
wongma7, line 254: this can be refactored into function (getCloud or something) since both DeleteVol and CreateVol rely on it

pkg/driver/node_test.go
```diff
@@ -303,6 +303,19 @@ func TestNodePublishVolume(t *testing.T) {
 			mountArgs:     []interface{}{volumeId + ":/", targetPath, "efs", []string{}},
 			mountSuccess:  true,
 		},
+		{
+			name: "success: normal with volume context populated from dynamic provisioning",
```
wongma7, line 307: I would like to see an e2e test for cross Account mount but it's going to be hard since we need 2 accounts and to store creds somewhere. Any ideas how to achieve this? we could store the creds in secrets manager in the existing CI account and construct the secret by retrieving fro msecrets manager as part of the test.

pkg/cloud/cloud.go
kbasv: `azName` is optional and this func is only called if azName is provided. I think we should keep this message here to avoid if-else in the caller to print this message

pkg/driver/controller.go
kbasv: If a user wants to use cross-az/same-az for same account, they can make use of `az` mount option, for example they can pass `az=<desired az>` under storage class mount options and efs-utils will figure out the az/mount-target-ip to mount.

However, for cross-account mount, due to the inability to pass credentials to node service / efs-utils, we have to pre-determine the az/mount-target-ip and pass it to efs-utils.

pkg/cloud/cloud.go
wongma7: i mean here: https://github.com/kubernetes-sigs/aws-efs-csi-driver/pull/470/files#diff-d7ef58b21290a74508c38ca5fdcd104ca5d098820cfac0681ba5415c9820c91dR271

pkg/driver/node_test.go
kbasv: Need to give this some thought. Apart from having to manage credentials for two accounts, Cross-account mount requires `vpc peering`. I'm not sure how to achieve this on the fly for the integ tests.

pkg/driver/controller.go
wongma7: hm I didn't even notice az in both mountOptions and parameters tbh. Do you mind putting your comment in a code comment?

I double checked and if both mounttargetip and az set then efs-utils will just ignore az instead of trying to determine.

regarding the inability to pass credentials, there might be a way (in the future) to use the token feature https://github.com/kubernetes/enhancements/tree/master/keps/sig-storage/1855-csi-driver-service-account-token so that the node service pod can use the client pod's token to get AWS API creds. but that's another big issue.

wongma7:
lgtm pending doc/README/example update for the new parameter/secret

Dockerfile
```diff
@@ -31,19 +31,27 @@ ENV EFS_CLIENT_SOURCE=$client_source
...
+         cd efs-utils && \
+         git checkout $(git describe --tags $(git rev-list --tags --max-count=1)) && \
+         make rpm && yum -y install build/amazon-efs-utils*rpm; \
     fi
 
+# Install botocore required by efs-utils for cross account mount
+RUN yum -y install wget
+RUN wget https://bootstrap.pypa.io/get-pip.py -O /tmp/get-pip.py
+RUN python3 /tmp/get-pip.py
+RUN pip3 install botocore || /usr/local/bin/pip3 install botocore
```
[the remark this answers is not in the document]
wongma7: good idea, yeah we should follow the `RUN x && \ y` pattern and clean stuff up like yum cache, the size went up a lot. issue here: https://github.com/kubernetes-sigs/aws-efs-csi-driver/issues/481

PR #478, opened 2021-06-08
Added DNS Config for Controller to support EFS Access Point deletion.

shyamg28:
Bug fix according to me. Currently, only EFS Node supports DNS config allowing for file system DNS resolution. But the deletion of AccessPoints is controlled by EFS Controller which cannot resolve the file system domain name. 

This PR updates the EFS controller deployment template by adding DNS config support.

**What testing is done?** 
Deployed on EKS cluster and tested dynamic provisioning as well as auto-deletion of AccessPoints upon deletion of PersistentVolumeClaim.


shyamg28:
do-not-merge/work-in-progress

shyamg28:
CLA signed.

shyamg28:
Is there anything pending from my end?

PR #483, opened 2021-06-08
Clean up unnecessary resources after installation in docker file

kbasv:

Clean up unnecessary resources after installation in docker file

**What testing is done?**


wongma7:
/lgtm

thanks!!

Issue #486, opened 2021-06-10
Error message wrong with only gidRangeStart without gidRangeEnd


pierluigilenoci:
/kind bug

**What happened?**

If you specify only `gidRangeStart` without specifying `gidRangeEnd` the error message is wrong:

```shell
rpc error: code = InvalidArgument desc = Missing gidRangeStart parameter
```

**What you expected to happen?**

I expect to receive this message: 
```shell
rpc error: code = InvalidArgument desc = Missing gidRangeEnd parameter
```

In addition, it should be more clearly specified in the documentation whether these two parameters are really optional and that if one is specified, the other must also be specified. 

**How to reproduce it (as minimally and precisely as possible)?**

Don't specify `gidRangeEnd`

**Anything else we need to know?**:

I found the error in the code:

https://github.com/kubernetes-sigs/aws-efs-csi-driver/blob/863e445739d34cda44ca55e5dc17045650a40b51/pkg/driver/controller.go#L149-L153

This should be:

```go
return nil, status.Errorf(codes.InvalidArgument, "Missing %v parameter", GidMax)
```

**Environment**
- Kubernetes version (use `kubectl version`): not relevant
- Driver version: v1.3.0

kbasv:
Fixed in #487, closing this issue.

pierluigilenoci:
Closed by #487

PR #487, opened 2021-06-10
Fixed the error message

pierluigilenoci:

Bugfix for #486 

Because the error message now is misleading

**What testing is done?** 

None, just changed the error message string


pierluigilenoci:
@kbasv could you please take a look?

PR #494, opened 2021-06-16
Added extraCreateMetadata to the chart

### LINKED ISSUE #446: Dynamic Provisioning: Add PV/PVC identifying tags to the Access Points created by the csi
Hello,

its currently hard to identify which EFS Access Point is related to which pvc when looking at the EFS conosole.
I think that by adding a few tags to the Access Point resources created by the csi this issue can be easily resolved.
For example you can see the tags that are added to EBS volumes created by cloud provider:

- kubernetes.io/created-for/pv/name
- kubernetes.io/created-for/pvc/name
- kubernetes.io/created-for/pvc/namespace

Thank you.

pierluigilenoci:

New feature: added the chance to configure extraCreateMetadata in the CSI Provisioner container

Fixes #446 

**What testing is done?** 

Deployed on EKS cluster and tested

### DIFF
--- a/charts/aws-efs-csi-driver/values.yaml
+++ b/charts/aws-efs-csi-driver/values.yaml
@@ -41,6 +41,8 @@ controller:
   create: true
   # Number for the log level verbosity
   logLevel: 2
+  # If set, add pv/pvc metadata to plugin create requests as parameters.
+  extraCreateMetadata: true
   # Add additional tags to access points
   tags: {}
     # environment: prod
--- a/charts/aws-efs-csi-driver/Chart.yaml
+++ b/charts/aws-efs-csi-driver/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: aws-efs-csi-driver
-version: 2.1.1
+version: 2.1.2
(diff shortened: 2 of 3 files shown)


pierluigilenoci:
@wongma7 could you please take a look?

charts/aws-efs-csi-driver/values.yaml
```diff
@@ -41,6 +41,8 @@ controller:
   create: true
   # Number for the log level verbosity
   logLevel: 2
+  # If set, add pv/pvc metadata to plugin create requests as parameters.
+  extraCreateMetadata: false
```
wongma7, line 45: we can default it true, i ended up doing so in ebs https://github.com/kubernetes-sigs/aws-ebs-csi-driver/pull/937

pierluigilenoci: I was not sure about that. I'll do it.

pierluigilenoci: Done

wongma7:
/lgtm

thanks.

helm chart release will come soon.

PR #495, opened 2021-06-16
Add additional env vars to containers in Helm chart

dev-in-the-clouds:
Feature

Add support in Helm chart to add custom env vars to efs-plugin and csi-provisioner containers

**What testing is done?** 
Deployed the Helm chart with custom env vars

### DIFF
--- a/charts/aws-efs-csi-driver/values.yaml
+++ b/charts/aws-efs-csi-driver/values.yaml
@@ -45,6 +45,10 @@ controller:
   tags: {}
     # environment: prod
     # region: us-east-1
+  # Add additional env to containers
+  env:
+    efsPlugin: []
+    csiProvisioner: []
   # Enable if you want the controller to also delete the
   # path on efs when deleteing an access point
   deleteAccessPointRootDir: false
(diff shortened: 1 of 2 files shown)


dev-in-the-clouds:
CLA Signed

charts/aws-efs-csi-driver/values.yaml
```diff
@@ -45,6 +45,10 @@ controller:
   tags: {}
     # environment: prod
     # region: us-east-1
+  # Add additional env to containers
+  env:
+    efsPlugin: []
```
wongma7, line 50: for consistency with EBS, this should just be `controller.env: []`

similarly, `controller.env.csiProvisioner: []` should be moved to `sidecars.csiProvisioner.env`

wongma7: ref:

https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/charts/aws-ebs-csi-driver/values.yaml#L65

https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/charts/aws-ebs-csi-driver/values.yaml#L13

wongma7 (commented):
please shuffle the values a little for consistency with EBS chart. there is some overlap in users of both charts so ideally they should be as similar as possible. thank you

Issue #396, opened 2021-03-28
storage.k8s.io/v1beta1 CSIDriver is deprecated in v1.19+ unavailable in v1.22+


cippaciong:
/kind bug

**What happened?**
When you apply the manifests on Kubernetes >= `1.19`, the following warning message pops up: `Warning: storage.k8s.io/v1beta1 CSIDriver is deprecated in v1.19+, unavailable in v1.22+; use storage.k8s.io/v1 CSIDriver`

I'm not sure if this is intended for backward compatibility, but I guess the `apiVersion` can be updated since CSI reached GA in Kubernetes `1.14` and, as of version `v.1.2.0` of this driver, only Kubernetes versions >= `1.17` are supported.

**What you expected to happen?**
Don't see the warning

**How to reproduce it (as minimally and precisely as possible)?**
Apply the kustomize manifests on Kubernetes >= 1.19 (should be the same with helm)

**Anything else we need to know?**:
As I said, I'm not sure if this is the intended behaviour for backward compatibility. If that's the case feel free to close the issue and sorry for the fuss 🙂 

**Environment**
- Kubernetes version (use `kubectl version`): `v1.19.6-eks-49a6c0`
- Driver version: `v1.2.0`

stijnbrouwers:
+1

PR #506, opened 2021-07-08
Avoid deprecation warnings. CSI is GA as of 1.13

sdwire:
According to https://kubernetes.io/blog/2019/01/15/container-storage-interface-ga/, CSI became GA in 1.13. As of 1.20, it's throwing deprecation warnings, triggering issue #396 . I'll admit to not knowing how the build process works for this project, so I haven't tested this. But this appears to be a reasonable place to reflect where GA actually happened so that the latest versions of the driver can stop throwing deprecation warnings.


sdwire:
/close

Looks like the goal here was already accomplished through some other means. This PR is unnecessary.

PR #507, opened 2021-07-09
Added proxy to the chart

### LINKED ISSUE #409: Ability to set http proxy in helm chart

danielwegener:
This PR adds the ability of setting proxy environment variables for the controller.

Closes #409
This is the another attempt for PR #410 which seem to be stuck in feedback process (with minor changes).

What testing is done?
We have used the changes from #410  for ~2 month.

### DIFF
--- a/charts/aws-efs-csi-driver/templates/controller-deployment.yaml
+++ b/charts/aws-efs-csi-driver/templates/controller-deployment.yaml
@@ -57,6 +57,16 @@ spec:
           env:
             - name: CSI_ENDPOINT
               value: unix:///var/lib/csi/sockets/pluginproxy/csi.sock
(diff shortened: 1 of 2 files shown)


danielwegener:
closed in favor of #410

PR #509, opened 2021-07-13
WIP: test removing reaper

wongma7:

**What is this PR about? / Why do we need it?** It's not clear if the reaper is needed , the watchdog should be responsible for terminating/cleaning up stunnel processes. https://github.com/aws/efs-utils/blob/f3bb4a228a40bcf2e82fc28ab898d2b44ed8b05e/src/watchdog/__init__.py#L608-L642 . ~I guess if the watchdog restarts then zombie stunnel processes might need to be reaped?~ edit: i dont think zombies can happen actually, the stunnel processes should still be running and their PID is stored in state file, so watchdog should figure out again that they are still running and not spawn another stunnel.

But we are hearing about issues where mount fails with 'wait: no child processes' and I suspect this reaper is to blame, it might be racing with os.Command in mount_linux.go https://github.com/kubernetes/mount-utils/blob/19c6fae41ad49f3e4026f28b043f41343aa852fd/mount_linux.go#L176, I can't think of any other explanation for this weird error.

original PR adding reaper https://github.com/kubernetes-sigs/aws-efs-csi-driver/pull/113

**What testing is done?**

### DIFF
--- a/pkg/driver/driver.go
+++ b/pkg/driver/driver.go
@@ -120,10 +120,6 @@ func (d *Driver) Run() error {
 		return err
 	}
 
-	reaper := newReaper()
-	klog.Info("Staring subreaper")
-	reaper.start()
-
 	klog.Infof("Listening for connections on address: %#v", listener.Addr())
 	return d.srv.Serve(listener)
 }


wongma7:
/close

we do need a reaper because stunnel process have parent driver->mount.efs->stunnel but get terminated by driver->watchdog, and since driver is the parent and not watchdog it's driver's responsibility to reap driver->mount.efs->stunnel processes.

However as mentioend I suspect it is possible the reaper is waiting on non-stunnel processes and racing with os.exec in other parts of the driver, causing those os.exec waits to error out because the process disappears/gets replaced from under them?.

PR #510, opened 2021-07-13
Fast-forward to latest ebs hack/e2e scripts with eksctl support, k8s 1.20, etc.

wongma7:

I applied the latest version of scripts from https://github.com/kubernetes-sigs/aws-ebs-csi-driver/tree/master/hack/e2e
```
git fetch ebs                                                                  
git diff HEAD:hack/e2e ebs/master:hack/e2e > /tmp/hack_e2e.diff                
git apply --reject --directory hack/e2e /tmp/hack_e2e.diff
git commit
```

**What testing is done?** 
scripts work in ebs repo, will let CI test if they continue to work in this one.

### DIFF
--- a/hack/e2e/README.md
+++ b/hack/e2e/README.md
@@ -41,7 +41,16 @@ To commit changes and submit them as a PR back to the ebs repo:
 
 ```
 git diff ebs/master:hack/e2e HEAD:hack/e2e > /tmp/hack_e2e.diff
-cd $GOPATH/src/github.com/kubernetes-sigs/aws-ebs-csi-driver
+pushd $GOPATH/src/github.com/kubernetes-sigs/aws-ebs-csi-driver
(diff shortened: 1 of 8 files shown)


nckturner:
So are the copied from the EBS repository with slight modifications?  Don't we use submodules to avoid copying scripts?  (Should we start moving some shared scripts, like the eksctl, to https://github.com/kubernetes-sigs/provider-aws-test-infra?)

wongma7:
@nckturner there are no modifications, see the commands i  did in the pr test. submodules doesn't work well for copying a subdirectory, that's why i use read-tree and the commands i did.

nckturner:
Ok, Not sure exactly what you mean, I think we can probably solve this by moving anything that is the same to the test-infra respository but I'm not going to block on that.

Issue #503, opened 2021-07-06
PodSecurityPolicy and Daemonset Service Account for Helm Chart


Reinkaos:
Hey,

Not sure if this is a bug or a feature, have put it here as it's preventing me from using the chart but it could be considered a feature. 

**What happened?**

I am unable to get the Node CSI Driver deployed using the Helm Chart because it does not come with it's own Pod Security Policy, does not comply with my clusters default pod security policy and there is no service account available with the daemonset that I can bind my own policy too. 

**What you expected to happen?**

The chart to deploy an appropriate service account and pod security policy for the controller and the daemonset or it comes with a service account that can be used for binding the policy. 

**How to reproduce it (as minimally and precisely as possible)?**

You can reproduce it by implementing a PSP like so:

```
apiVersion: policy/v1beta1
kind: PodSecurityPolicy
metadata:
  name: restrictive
spec:
  privileged: false
  hostNetwork: false
  allowPrivilegeEscalation: false
  defaultAllowPrivilegeEscalation: false
  hostPID: false
  hostIPC: false
  runAsUser:
    rule: RunAsAny
  fsGroup:
    rule: RunAsAny
  seLinux:
    rule: RunAsAny
  supplementalGroups:
    rule: RunAsAny
  volumes:
  - '*'
  allowedCapabilities:
  - '*'
  ```

and binding it to your authenticated group:

```
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  name: restrictive-psp
rules:
- apiGroups:
  - policy
  resourceNames:
  - restrictive
  resources:
  - podsecuritypolicies
  verbs:
  - use
---
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRoleBinding
metadata:
  name: restrictive-psp
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: ClusterRole
  name: restrictive-psp
subjects:
- apiGroup: rbac.authorization.k8s.io
  kind: Group
  name: system:authenticated
```

and then trying to deploy the chart, which will result in this error on the controller and the daemonset:

```
[...]

Issue #397, opened 2021-03-29
Daemonset is using the default service account 


pre:
The DaemonSet defined `privileged:true` and uses `hostPath` which require these permissions granted by a Pod Security Policy.

However, the DaemonSet does not use an explicit `serviceAccount` which means that the `default` service account of its namespace is implicitly used.

In order use this Helm chart with PSPs, as a workaround, the aws-efs-csi-driver needs to deployed into a k8s namespace of its own and the grant the `default` service account these permissions. 

It would be neat if an explicit `serviceAccount` could be configured for both the Deployment and the DaemonSet. This would allow using the same PodSecurityPolicy or different (depending on the level of folio hat). 

/bug

Related: https://github.com/kubernetes-sigs/aws-efs-csi-driver/issues/243

joelthompson:
This looks to be a duplicate of #316 / #335

Issue #316, opened 2021-01-29
Node.yaml should allow specifying serviceaccount (release 1.1)?


esalberg:
/kind bug

**What happened?**
I was testing service account settings with release 1.1 (dev overlay).  After updating serviceaccount-csi-controller.yaml, the serviceAccountName setting in controller.yaml worked appropriately.  However, the pods from node.yaml were not using the serviceAccountName (until I added that line to the template spec).

**What you expected to happen?**
Could you please confirm whether node.yaml also needs to assume a role for IRSA to work successfully?  Our goal is to specify a specific role for the EFS CSI driver (e.g. role.efs) and then add that to the EFS filesystem policies for access.  That way, we won't have to update the EKS worker node role in the filesystem policies every time we update an EKS node group.

I wanted to check prior to release 1.1 just to make sure this wasn't overlooked.  Please close this ticket if node pods don't need to assume the role, just the controller pods.

Note: My IRSA testing ultimately failed anyway, I'm assuming due to issue #290 (Upgrade to efs-utils v1.28.2).  Or is the service account supposed to work / be assumed successfully for EFS even without the efs-utils update?
 
**How to reproduce it (as minimally and precisely as possible)?**
 - Update serviceaccount-csi-controller.yaml (or use Kustomize to override, I'm assuming) to set  the role-arn annotation
 - Deploy with kubectl apply -k dev/ (from the overlays directory)
 - Check the pods and see that the SA is in use for the controller pods but not the node pods (set to default role)

 (Even after I added the service account by hand to the node.yaml template spec, NewClientConnections from an app pod were rejected due to use the default role - which as I noted could be due to the efs-utils version.  I have not tried manually updating efs-utils in the pod to confirm.)

**Anything else we need to know?**:

**Environment**
- Kubernetes version (use `kubectl version`): v1.18.9-eks-d1db3c
- Driver version: 1.1 dev (release-1.1 branch)

wongma7:
We have included an example controller serviceaccount because it's ALWAYS needed but we are lacking an example node serviceaccount because it's only sometimes needed. We don't intend for both controller and node to use the same serviceaccount.

The example service account is provided for the controller only because the controller MUST have permission to create and delete access points. Documentation about the controller is coming: https://github.com/kubernetes-sigs/aws-efs-csi-driver/pull/297/files .

[...]

PR #512, opened 2021-07-14
Add node/daemonset service account to helm chart

wongma7:

node should have option to have its own serviceaccount (besides "default"). Moreover there is a specific situation where you need to give the node permissions, so for EKS users the serviceaccount can be used for IAM for SA. https://github.com/kubernetes-sigs/aws-efs-csi-driver/tree/master/examples/kubernetes/cross_account_mount#prerequisite-setup
https://github.com/kubernetes-sigs/aws-efs-csi-driver/issues/503
https://github.com/kubernetes-sigs/aws-efs-csi-driver/issues/397
https://github.com/kubernetes-sigs/aws-efs-csi-driver/issues/316

**What testing is done?**


kbasv (commented):
LGTM. Can you update the helm change log as well?

PR #497, opened 2021-06-18
fix log collector

wreed4:
**Is this a bug fix or adding new feature?**

**What is this PR about? / Why do we need it?**

**What testing is done?**

PR #525, opened 2021-07-29
Fix (and format) log collector script

wongma7:
/bug
**What is this PR about? / Why do we need it?** supersedes https://github.com/kubernetes-sigs/aws-efs-csi-driver/pull/497 which we an't merge due to CLA.

BTW, I guess we should prioritize feeding/tailing/reading efs-utils logs into the driver stderr so that users don't have to run this script at all!

**What testing is done?**

### DIFF
--- a/troubleshooting/log_collector.py
+++ b/troubleshooting/log_collector.py
@@ -9,16 +9,17 @@ parser = argparse.ArgumentParser(description="Troubleshooting EFS CSI Driver")
 
 parser.add_argument("--driver-pod-name", required=True, help="The EFS CSI driver pod name")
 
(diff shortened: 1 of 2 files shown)


troubleshooting/log_collector.py
```diff
@@ -9,17 +9,18 @@
 
 parser.add_argument("--driver-pod-name", required=True, help="The EFS CSI driver pod name")
 
-args = parser.parse_args(['--driver-pod-name', 'efs-csi-node-5dpc4'])
+args = parser.parse_args()
```
wongma7, line 12: here is the fix. rest is formatting (black)

kbasv: Thank you for the fix!

PR #530, opened 2021-08-04
Release v1.3.3

kbasv:

Bump release v1.3.3

**What testing is done?**

### DIFF
--- a/CHANGELOG-1.x.md
+++ b/CHANGELOG-1.x.md
@@ -1,3 +1,12 @@
+# v1.3.3
+
+### Misc.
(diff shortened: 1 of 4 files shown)


deploy/kubernetes/overlays/stable/kustomization.yaml
```diff
@@ -4,7 +4,7 @@ bases:
   - ../../base
 images:
   - name: amazon/aws-efs-csi-driver
-    newTag: v1.3.2
+    newTag: v1.3.3
```
wongma7, line 7: for the new release process we want to update kustomize only after the image has been pushed https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/docs/RELEASE.md (so we will need 2 PRs for every release)

wongma7: I guess for this repo since the helm release action acts on the relkease branch instead of master branch we also have t odelay helm changes until after image build

PR #514, opened 2021-07-15
Only reap zombie stunnel processes

wongma7:
**Is this a bug fix or adding new feature?**

**What is this PR about? / Why do we need it?**

I suspect our reaper is causing (rarely) mounts to error with 'wait: no child processes'' because it might be catching SIGCHLD and waiting for `mount` child processes before os.exec can wait for them: https://cs.opensource.google/go/go/+/refs/tags/go1.16.6:src/os/exec_unix.go;l=43;drc=refs%2Ftags%2Fgo1.16.6.

Since the intention of reaper is to clean up zombie stunnel processes which as explained in code comment "become zombies because their parent is the
// driver process but they get terminated by the efs-utils watchdog process:", this PR limits our reaper to cleaning up those.

Arguably maintaining this reaper and having this reaper constantly reading the process table is overkill. In most cases it should be fine to simply let the zombie processes be, the pid limit is 4194303, but in interest of not changing too much to o quickly I'd rather we try this reaper change before we remove the reaper altogether.

(previous try to simply remove reaper: https://github.com/kubernetes-sigs/aws-efs-csi-driver/pull/509)

**What testing is done?** 
build and deploy image.
use dynamic provisioning example (with my own fs of course)
```
$ k apply -f examples/kubernetes/dynamic_provisioning_test_cluster_1/specs/
$ k exec -n kube-system efs-csi-node-8xgsw -it -- sh
sh-4.2# ps aux
USER         PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
root           1  0.0  0.4 119932 17296 ?        Ssl  01:10   0:00 /bin/aws-efs-csi-driver --endpoint=unix:/csi/csi.sock --logtostderr --v=4
root          17  0.1  0.4 110812 18876 ?        S    01:10   0:00 python3 /usr/bin/amazon-efs-mount-watchdog
root          31  0.0  0.1 180072  5888 ?        Ss   01:11   0:00 /usr/bin/stunnel /var/run/efs/stunnel-config.fs-176f8111.var.lib.kubelet.pods.bad00a46
root          55  0.0  0.0  11672  2836 pts/0    Ss   01:12   0:00 sh
root          74  0.0  0.0  51848  3456 pts/0    R+   01:13   0:00 ps aux
```

delete the pod from the example and then check logs.
```
$ k delete po efs-app
$ k logs -n kube-system efs-csi-node-8xgsw efs-plugin
[...]

PR #544, opened 2021-08-30
[release-1.3] cherry-pick: Only reap zombie stunnel processes

chrishenzie:
Fixing a bug.

This is a cherry-pick of https://github.com/kubernetes-sigs/aws-efs-csi-driver/pull/514, which fixes an issue where mounts occasionally fail with:
```
mount_linux.go:150] Mount failed: waitid: no child processes
```

**What testing is done?**
Currently smoke-testing this in an AWS cluster and running the EFS E2E test suite repeatedly. For context, this issue has been cropping up occasionally and causing test-flakiness. Will post back with the results.

/assign @wongma7 
/assign @kbasv


wongma7:
/lgtm
/hold

feel free to cancel hold when testing is done if testing shows that this fixes the issue. then we can make a release.

@kbasv and I were going to hold this patch until 1.4.0 as we couldn't reproduce the issue and verify this was actually a bug fix but if your testing shows otherwise we should make a 1.3.x fix.

chrishenzie:
@wongma7 Thank you! For clarification, we have some E2E tests that run this test suite periodically and have encountered some test flakes due to timeouts. When I inspect the logs I see this line about no child processes. I was able to repro locally by re-running the E2E test-suite repeatedly until it exited with code != 0.

chrishenzie:
@wongma7 The test ran successfully 34 times over a 17 hour span without any failures. Removing the hold.

/hold cancel

Issue #547, opened 2021-09-02
Block volumes can be provisioned


jsafrane:
/kind bug

**What happened?**
E2E test `External Storage [Driver: efs.csi.aws.com] [Testpattern: Dynamic PV (block volmode)] volumeMode should fail in binding dynamic provisioned PV to PVC [Slow][LinuxOnly]` fails, because the CSI driver will provision a volume when CreateVolume requests a block volume. The driver should return an error in this case.

**How to reproduce it (as minimally and precisely as possible)?**
Run Kubernetes external CSI tests.

PR #548, opened 2021-09-02
Fix block provisioning

jsafrane:
/kind bug
Fixes: #547

Return error when CO requests a block volume to be dynamically provisioned. EFS does not support them. When at it, add more error reporting when validating volume capabilities, so the caller (or user) knows what's wrong with a capability instead of generic "Volume capability not supported".

**What testing is done?** 
Tested with the Kubernetes external CSI tests.

### DIFF
--- a/pkg/driver/controller.go
+++ b/pkg/driver/controller.go
@@ -18,6 +18,7 @@ package driver
 
 import (
 	"context"
+	"fmt"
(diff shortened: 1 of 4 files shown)


jsafrane:
cc @wongma7 @kbasv
PTAL

Issue #543, opened 2021-08-30
Dynamic basePath


wolffberg:
**Is your feature request related to a problem? Please describe.**
When working with multi tenant clusters where each tenant can provision x amount of PVCs, identifying files for cleanup, restoration, etc. Becomes hard as directories can only be identified by the PVC id.

The built-in option to delete data with access points is too risky for our use case.

**Describe the solution you'd like in detail**
We would like a way of defining a path per namespace/tenant e.g. `/basePath/namespace/pvc-id` through annotations or other attributes in the `PersistentVolumeClaim` resource or define a dynamic `basePath` format in the `StorageClass` resource.

If attributes in the `PersistentVolumeClaim` is a posibility, correct values can be enforced with the use of [OPA Gatekeeper](https://github.com/open-policy-agent/gatekeeper) or similar.

This way you could easily create cleanup logic for deleted namespaces and copy files to new PVC volumes if required.

**Additional context**
Looking through [controller.go](https://github.com/kubernetes-sigs/aws-efs-csi-driver/blob/master/pkg/driver/controller.go) I haven't been able to identify whether the controller actually get data related to the `PersistentVolumeClaim` used to generate access points.

I'm hoping I have either missed something or it can be easily implemented.

wongma7:
the controller does have access to the PVC namespace so this is theoretically possible. otherwise the storageclass parameter idea also sounds reasonable to me.

richerve:
I also have a need for this, what I'm thinking is having an admission controller that will add a `/<namespace>` prefix to any PVC using storageClass `efs-<something>` or whatever I use as storageClass name. But this is an extra step.

Having said that it seems that this kind of behaviour is better handled by an operator than by a csi driver but in this specific case it'd probably be just an extra flag in the StorageClass, something like `prefixNamespaceInPath` or something similar, or maybe a templatized value like: `pathPrefix: "$namespace"`.

@wongma7 I can have a go in adding this feature, is there interest in it?

PR #568, opened 2021-10-18
Allow more control of the name of the directory created by the Access Point under Dynamic Provisioning

jonathanrainer:
This adds a new feature, mostly in response to discussions on Issue #543 

At the moment if you use dynamic provisioning the name of the folder created on the EFS is the name of the generated PV however in some use cases this isn't desirable and leads to a lot of confusion. Under this scheme the name of the directory can be controlled by users using a new parameter `rootDirectoryPath` and deprecating the old `basePath`. This allows users a higher level of control over the directories that are created as they can specify any combination of fixed strings and a set of predefined variables based on what the CSI Spec supports.

**What testing is done?** 
I have added to the unit tests for the module and I believe this is sufficient because all that is being changed is the directory being added to the AccessPointOpts struct. There might want to be some E2E testing about what happens if that directory already exists but I'm happy to take pointers on that as this is my first PR to the repo.


jonathanrainer:
Also thinking about it maybe the best idea is to remove "basePath" if and swap it for something like "pathTemplate"? I think that's much clearer as to what that parameter does, that way we can support both and then deprecate basePath if we want to or go the whole hog and deprecate it in one fell swoop if people want to take the major version bump.

wolffberg:
If we want backwards compatibility we could keep `basePath` as is and join another path from a new parameter like `subPath`, `subPathTemplate` or similar.
This would also make it easier to avoid name clashes if you for some reason want to use the EFS share for other things than your PVCs.

wongma7:
I like wolffberg's idea, it is easy to reason about.

My highest priority is backwards compatibility, I think it is too soon for a major version update and an upgrade story that entails recreating storageclasses is hard to reason about and could be troublesome.

Another option is to maintain both basePath and rootDirectoryPath and make them mutually exclusive. I don't like this because it is confusing and I want to avoid doing validation of storageclass in general (because validation cannot occur at storageclass creation time, it can only happen later at pv provision time, it is unpleasant UX). But I mention it for completeness

Also, I slightly prefer reusing go templating syntax over coming up with something new. helm should let us escape expressions. Another option is to copy https://github.com/kubernetes-sigs/nfs-subdir-external-provisioner pathPattern parameter templating syntax. I don't know how they arrived at that syntax. But there's probably a nonzero overlap of users who would benefit from consistency since https://github.com/kubernetes-sigs/nfs-subdir-external-provisioner does technically work with EFS too.

jonathanrainer:
Ok, sorry about the mess with this one, I forgot that fiddling with `master` was probably going to affect the PR but anyway. 

I've re-written this to keep `basePath` around in line with @wolffberg's suggestion and have incorporated the syntax from `nfs-subdir-external-provisioner` as a starting point. Their version is more expansive because you're allowed to use things like annotations and arbitrary labels and if we wanted to add that in the future we could but it looked quite complicated and I think it would be scope creep for what we're doing here.

Let me know what you think and we can work from there.

iAnomaly:
Thanks for continuing to work through this @jonathanrainer. My only question currently is how does the current implementation support a desired path of just `/`? Reading through the code it seems like the way to achieve that would be with:
```
basePath: nil
subPathPattern: ""
```
But it looks like subPathPattern doesn't currently support a `length == 0` case? https://github.com/kubernetes-sigs/aws-efs-csi-driver/pull/568/files#diff-4e4f8ad2f057d794bd5842ed50e358451e6e1b3b20e5f2836289665bdfbf5625R252

pkg/driver/controller.go
```diff
@@ -234,7 +247,25 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
 	}
 
 	rootDirName := volName
-	rootDir := basePath + "/" + rootDirName
+	// Check if a custom structure should be imposed on the access point directory
+	if value, ok := volumeParams[SubPathPattern]; ok {
+		if len(value) == 0 {
+			return nil, status.Errorf(codes.InvalidArgument, "No pattern specified for subPath.")
```
iAnomaly, line 253: Nit: this should say `No pattern specified for subPathPattern.` instead of `No pattern specified for subPath.`.

jonathanrainer: Resolved this by removing the check completely to support the use case referred to above.

jonathanrainer:
@iAnomaly Thanks for the feedback! You're right that is certainly a use-case I overlooked, I've now updated the code so it's possible to have a 0-length pattern and added a test to show that it gives you the correct constructed directory (i.e. /) once it's been through the process.

pkg/driver/controller.go
```diff
@@ -233,8 +242,27 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
...
+			klog.Infof("Using user-specified structure for access point directory.")
+			rootDirName = val
+		} else {
+			return nil, err
+		}
+	} else {
+		klog.Infof("Using PV name for access point directory.")
+	}
+
+	rootDir := path.Join("/", basePath, rootDirName)
```
wongma7, line 264: since it's possible now for names to collide whereas before we could assume that the path would be unique since volName would be the pvc+uuid, I'm wondering whether we need to handle it. Especially if deleteAccessPointRootDir is set to false then we could have a situation where say I have iteration 1 of pod+pvc-a writing data to directory pvc-a and then iteration 2 of pod+pvc-a (not necessarily in the same namespace as iteration 1) unexpectedly reading/writing to the same directory pvc-a.

It puts quite a burden on administrators to set up the storageclasses 'correctly' so that boundaries are respected. If I just lazily create a storageclass with subpath /{PVC.name} then users from namespace A or B could both access that which is probably not what I intend.

Just want to start a discussion for now, I anticipate we might get blocked releasing this on security grounds in its current form.

jonathanrainer: That's reasonable, and I'm certainly not an expert in this stuff which is why I rely on other people who are! In which case, since the EBS driver already does this do they not have the same problem? And what do you think is the best way to deal with a situation like this?

* Can we query the EFS to check if any access points overlap? How similar do we need to check that they are? i.e. It might be completely legitimate to have 2 access points pointing at the same directory if they had different GUID/UID properties for example
* Were this to be violated what's an acceptable response, should the driver just return an error, as it does if you misconfigure the subPathPattern?

wolffberg: I think the path should still contain the UUID to prevent collisions e.g. `/cluster/namespace/pvc-1d8451fe-ee96-421a-b55f-c90a346e1215`

Also, I just found this paragraph on the [EFS docs](https://docs.aws.amazon.com/efs/latest/ug/efs-access-points.html#enforce-root-directory-access-point) that might require a check or two.

> You enable this feature by setting the access point Path attribute when creating an access point. The Path attribute is the full path of the root directory of the file system for all file system requests made through this access point. The full path can't exceed 100 characters in length. It can include up to four subdirectories.

jonathanrainer:
This PR being linked to master is making it difficult to manage. I'm going to close this PR now and re-open it properly so re-bases etc. will not cause me to spend in-ordinate amounts of time fixing it.

PR #581, opened 2021-11-03
Build target platform from build platform

wongma7:

**What is this PR about? / Why do we need it?** https://github.com/kubernetes-sigs/aws-fsx-csi-driver/issues/218 this shoudl make the build faster since we can build arm (target) binaries from a amd (build) node/host: https://docs.docker.com/buildx/working-with-buildx/#build-multi-platform-images.

similar to fsx PR (trying to be consistent): https://github.com/kubernetes-sigs/aws-fsx-csi-driver/pull/221

**What testing is done?**


wongma7:
/lifecycle frozen

still want this just don't have time to rebase ATM

PR #582, opened 2021-11-04
Adds more permissions to avoid errors during deployment

heitorrbarros:
This is an adjustment to the IAM policy example. During the deployment I received some errors by missing some permissions.


heitorrbarros:
Check my CLA sign, pls ❤️

Issue #564, opened 2021-10-12
Logs flooded by Connecting to unix...


renanqts:
/kind bug

**What happened?**
Logs are flooded by:
efs-csi-node-k6lkl liveness-probe I1012 10:17:54.811549       1 connection.go:153] Connecting to unix:///csi/csi.sock

**What you expected to happen?**
Don't see this log, at least in log level 1

**How to reproduce it (as minimally and precisely as possible)?**
Just deploy and see the logs

**Anything else we need to know?**:

**Environment**
- Kubernetes version (use `kubectl version`):
- Driver version:
v2.2.0-eks-1-18-2

tanvp112:
For a sizeable cluster, such rapid logging can eat up A LOT of diskspace.

keifgwinn:
please address this, it's such an easy fix

janaurka:
BTW: this can be fixed with overwriting the default values file with:

```
        sidecars:
          livenessProbe:
            image:
              tag: v2.6.0-eks-1-21-13
```
so https://github.com/kubernetes-sigs/aws-efs-csi-driver/pull/595 is essentially too complicated, since it’s also changing the Container Registry.

If the above mentioned solution does not work for you _and_ you are using a k8s version <1.12, then look at the compatible images under: https://gallery.ecr.aws/eks-distro/kubernetes-csi/livenessprobe

If I find the time, I’ll create a PR, to update the `livenessProbe` Image to v2.6.

ilvez:
When driver was installed with Helm, I used this command to update image:
```
kubectl -nkube-system set image deployment/efs-csi-controller liveness-probe=public.ecr.aws/eks-distro/kubernetes-csi/livenessprobe:v2.6.0-eks-1-18-17
```

Your setup may vary -- namespace, deployment name.

ilvez:
The issue is still in the latest chart. I managed to fix it with downloading the chart and replaced sidecars.livenessProbe.image.tag
```
sidecars:
  livenessProbe:
    image:
      repository: public.ecr.aws/eks-distro/kubernetes-csi/livenessprobe
      tag: v2.6.0-eks-1-18-17 # v2.2.0-eks-1-18-13
```
Then applied the upgrade:
```
[...]

PR #595, opened 2021-12-07
Moving to livenessProbe 2.4 to stop logs being flooded with unnecessary messages

jonathanrainer:
I've had a go at this one to resolve the issue mentioned below. I'm not enormously familiar with Kustomize and our reasons for not using `k8s.gcr.io/sig-storage/livenessprobe` so if I've done something wrong please let me know so we can move forward to fix it together! This is largely based on the [fix for the EBS driver](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/pull/1054/files)

This is a bug fix to address the issues raised in #564, moving to a new version of the liveness probe will stop the large volume of logs being generated.

See above

**What testing is done?** 
Using the PR process to run the E2E tests will give confidence that this change will not break current installations. Once the PR is more stable/reviewed I will spin up the driver on my own EKS cluster to check the volume of logs is reduced as expected.

### DIFF
--- a/charts/aws-efs-csi-driver/values.yaml
+++ b/charts/aws-efs-csi-driver/values.yaml
@@ -15,8 +15,8 @@ image:
 sidecars:
   livenessProbe:
     image:
-      repository: public.ecr.aws/eks-distro/kubernetes-csi/livenessprobe
-      tag: v2.2.0-eks-1-18-13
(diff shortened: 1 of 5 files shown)


jonathanrainer:
Closing in preference to #688

PR #599, opened 2021-12-09
Release v1.3.5

Ashley-wenyizha:
Release of v.1.3.5

PR #601, opened 2021-12-10
Release v1.3.5

wongma7:

**What is this PR about? / Why do we need it?** merging https://github.com/kubernetes-sigs/aws-efs-csi-driver/pull/599 and https://github.com/kubernetes-sigs/aws-efs-csi-driver/pull/600 back into master branch following https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/docs/RELEASE.md#merge-the-release-and-post-release-commits-to-the-main-branch

**What testing is done?**

### DIFF
--- a/charts/aws-efs-csi-driver/CHANGELOG.md
+++ b/charts/aws-efs-csi-driver/CHANGELOG.md
@@ -1,7 +1,10 @@
 # Helm chart
 
-# v2.2.1
+# master (unreleased)
 * Add controller.volMetricsOptIn for emitting volume metrics
+
+# v2.2.1
+* Bump app/driver version to `v1.3.5`
 
 # v2.2.0
 * Allow health ports to be configured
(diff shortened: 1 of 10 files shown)


charts/aws-efs-csi-driver/CHANGELOG.md
```diff
@@ -1,8 +1,11 @@
 # Helm chart
 
-# v2.2.1
+# master (unreleased)
```
wongma7, line 3: conflict had to be resolved here.

for the efs repo, unlike fsx and ebs, the helm canonical source of helm chart is the release branch. So if a helm chart is edited in master branch it isn't actually released unless the change is backported to the release branch. In separate PRs, I'll backport the "Add controller.volMetricsOptIn for emitting volume metrics" change to release branch and then reconcile release and master branch.

The reason for this is because I found during dynamic provisioning release there was a long time between when helm chart was changed and when we could actually release it so it made sense to branch the helm chart to keep it isolated from master branch. Not sure if it's worth continuing this pattern though

PR #612, opened 2022-01-10
Should not display FSType in describe pv for AWS EFS driver

RomanBednar:
This is a bug fix but it changes how StorageClass is checked. StorageClass objects with invalid fstype capability will no longer pass the check and volume creation will be prevented. This applies to **all new PVs created** - similar to other checks that were added in the past (e.g. [access mode](https://github.com/kubernetes-sigs/aws-efs-csi-driver/blob/fc4d43d0f6ac3ff6619672c9862ef88b31ab3506/pkg/driver/node.go#L318)).

This change is needed to prevent PVs being created with incorrect values, currently the driver let's users create PVs with any value which is not correct. We should check if supported fstype was used.

**What testing is done?** 
Adding unit tests to cover the new check, both negative and positive flow. Also added a test for multiple messages which can happen when there are multiple unsupported capabilities found.

### DIFF
--- a/pkg/driver/node.go
+++ b/pkg/driver/node.go
@@ -36,7 +36,8 @@ var (
 		csi.VolumeCapability_AccessMode_SINGLE_NODE_WRITER,
 		csi.VolumeCapability_AccessMode_MULTI_NODE_MULTI_WRITER,
 	}
-	volumeIdCounter = make(map[string]int)
+	volumeIdCounter  = make(map[string]int)
+	supportedFSTypes = []string{"efs", ""}
 )
 
 func (d *Driver) NodeStageVolume(ctx context.Context, req *csi.NodeStageVolumeRequest) (*csi.NodeStageVolumeResponse, error) {
@@ -312,6 +313,10 @@ func (d *Driver) isValidVolumeCapabilities(volCaps []*csi.VolumeCapability) erro
 	if err := d.validateAccessType(volCaps); err != nil {
 		return err
 	}
+
+	if err := d.validateFStype(volCaps); err != nil {
+		return err
+	}
 	return nil
 }
 
@@ -342,6 +347,28 @@ func (d *Driver) validateAccessType(volCaps []*csi.VolumeCapability) error {
 		if c.GetMount() == nil {
 			return fmt.Errorf("only filesystem volumes are supported")
 		}
+	}
+	return nil
+}
+
+func (d *Driver) validateFStype(volCaps []*csi.VolumeCapability) error {
+	isSupportedFStype := func(cap *csi.VolumeCapability) bool {
+		for _, m := range supportedFSTypes {
+			if m == cap.GetMount().FsType {
+				return true
+			}
+		}
+		return false
+	}
+
+	var invalidFStypes []string
+	for _, c := range volCaps {
+		if !isSupportedFStype(c) {
+			invalidFStypes = append(invalidFStypes, c.GetMount().FsType)
+		}
+	}
+	if len(invalidFStypes) != 0 {
+		return fmt.Errorf("invalid fstype: %s", strings.Join(invalidFStypes, ","))
 	}
 	return nil
 }
--- a/pkg/driver/node_test.go
+++ b/pkg/driver/node_test.go
@@ -317,6 +317,26 @@ func TestNodePublishVolume(t *testing.T) {
 			mountSuccess:  true,
 		},
 		{
+			name: "success: supported volume fstype capability",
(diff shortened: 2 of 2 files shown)


RomanBednar:
The driver should now accept only "efs", "nfs, "" values for fstype in storage class. If incorrect value is used provisioning will fail.

Here is what I used for testing the patch manually:


1. StorageClass with incorrect "fstype: abc"

$ oc get sc/efs-sc-fstype -o yaml
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  creationTimestamp: "2022-01-12T14:02:15Z"
  name: efs-sc-fstype
  resourceVersion: "53362"
  uid: 938c14c9-2552-461d-8497-eea4fe5f7af1
mountOptions:
- tls
parameters:
  FStype: abc
  basePath: /dynamic_provisioning
  directoryPerms: "700"
  fileSystemId: fs-0f3a7db6f271f92e6
  provisioningMode: efs-ap
provisioner: efs.csi.aws.com
reclaimPolicy: Delete
volumeBindingMode: Immediate


2. Claim that references the incorrect StorageClass

$ oc get pvc/efs-claim-fail -o yaml
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  annotations:
    volume.beta.kubernetes.io/storage-provisioner: efs.csi.aws.com
  creationTimestamp: "2022-01-12T14:17:58Z"
  finalizers:
  - kubernetes.io/pvc-protection
  name: efs-claim-fail
  namespace: default
  resourceVersion: "58155"
  uid: 22828df9-c5e2-4907-9c45-8cf4a170529d
spec:
  accessModes:
  - ReadWriteMany
  resources:
    requests:
      storage: 5Gi
  storageClassName: efs-sc-fstype
  volumeMode: Filesystem
status:
  phase: Pending


3. Verify new error message about incorrect fstype

$ oc describe pvc/efs-claim-fail
Name:          efs-claim-fail
Namespace:     default
StorageClass:  efs-sc-fstype
Status:        Pending
Volume:        
Labels:        <none>
Annotations:   volume.beta.kubernetes.io/storage-provisioner: efs.csi.aws.com
Finalizers:    [kubernetes.io/pvc-protection]
Capacity:      
Access Modes:  
VolumeMode:    Filesystem
Used By:       efs-app-fail
Events:
  Type     Reason                Age                  From                                                                  Message
  ----     ------                ----                 ----                                                                  -------
  Normal   Provisioning          20s (x8 over 2m27s)  efs.csi.aws.com_ip-10-0-149-254_b82e2156-1754-432b-92d1-739c1ddb4480  External provisioner is provisioning volume for claim "default/efs-claim-fail"
  Warning  ProvisioningFailed    20s (x8 over 2m27s)  efs.csi.aws.com_ip-10-0-149-254_b82e2156-1754-432b-92d1-739c1ddb4480  failed to provision volume with StorageClass "efs-sc-fstype": rpc error: code = InvalidArgument desc = Volume capabilities not supported: invalid fstype: abc
  Normal   ExternalProvisioning  2s (x11 over 2m27s)  persistentvolume-controller                                           waiting for a volume to be created, either by external provisioner "efs.csi.aws.com" or manually created by system administrator

jsafrane:
> The driver should now accept only "efs", "nfs", "" values

@wongma7 @nckturner what do you think? Do you consider "efs" as the only `fsType` or  is "nfs" valid too?
Note that this is only about validation, choosing `nfs` won't disable encryption.

The code LGTM.

wongma7:
I think only efs should be accepted. Otherwise lgtm.

i dont' want to give impression that there is a difference between fstype nfs and fstype efs. sure, under the hood, mount.efs is just calling mount.nfs so maybe they should be treated the same..but i think it's simpler to just accept efs nd nothing else. Can loosen this restriction later if somebody complains!

RomanBednar:
Sounds good, removing nfs as suggested. Thank you for your feedback Matthew.

pkg/driver/node.go
```diff
@@ -36,7 +36,8 @@ var (
 		csi.VolumeCapability_AccessMode_SINGLE_NODE_WRITER,
 		csi.VolumeCapability_AccessMode_MULTI_NODE_MULTI_WRITER,
 	}
-	volumeIdCounter = make(map[string]int)
+	volumeIdCounter  = make(map[string]int)
+	supportedFsTypes = []string{"efs", ""}
```
jsafrane, line 40: nit: `supportedFSTypes`. Abbreviations have all characters in uppercase.

PR #619, opened 2022-01-20
Security patch & upgrade of k8s.io/kubernetes, linux and golang

Ashley-wenyizha:
- Security patch
- Instruct docker build with no-cache

Security patching on
1.  k8s.io/kubernetes: upgrade to 1.23
2. linux+golang security update

**What testing is done?** 
Uploaded the image to dev account and mitigated all high risks and the majority of medium risks


Dockerfile
```diff
@@ -28,7 +28,7 @@ ENV EFS_CLIENT_SOURCE=$client_source
 
 RUN GOOS=${TARGETOS} GOARCH=${TARGETARCH} make aws-efs-csi-driver
 
-FROM amazonlinux:2
+FROM amazonlinux:2.0.20211201.0
```
wongma7, line 31: we should not need to pin it (the latest at buildtime should get picked up )

Ashley-wenyizha: Thanks, let me still keep `FROM amazonlinux:2`

Ashley-wenyizha:
ECR enhanced scan after the upgrade
Screen Shot 2022-01-20 at 4 49 46 PM

wongma7:
After the 1.3.x release this change also should go in master branch as well ( I changed the default branch to release-1.3)

Issue #516, opened 2021-07-16
subPath mkdir


droctothorpe:
**Is your feature request related to a problem?/Why is this needed**
If I mount a subPath that doesn't exist, the mount fails. This is not true for the default provisioner. 

/feature

**Describe the solution you'd like in detail**
Specifying a subPath that doesn't exist should create the specified directory. 

**Describe alternatives you've considered**
Creating and chowning via an initContainer, but that's somewhat absurd. 

**Additional context**
I think K8s handles this natively here: 
https://github.com/kubernetes/kubernetes/blob/ea0764452222146c47ec826977f49d7001b0ea8c/pkg/volume/util/subpath/subpath_linux.go#L355
Maybe we can port some of the logic over. 

I'd be open to taking a stab at a PR if ya'll think this is a good idea.

droctothorpe:
Some additional context: JupyterHub/Kubespawner/Z2JH rely on subpaths to scope what subfolders users can mount on an EFS. When a new user onboards, the native controller creates the specified subPath. That doesn't happen with the EFS CSI driver. I would just use the native controller, but I need to mount as `1000:1000`, which is only possible with the EFS CSI driver. The JH workaround is to mount as root, create the subPath (in the K8s worker node / kubelet, not the container), then chown in the container, then deescalate. That's an unfortunate model because it requires granting root privileges to user containers which we want to avoid.

wongma7:
What do you mean default provisioner? The CSI driver is ignorant of subPaths and Pods, subPaths are handled at the kubelet level (you can think of the driver as one level "below" kubelet). So technically speaking I believe this is impossible.

droctothorpe:
If I set storageClass to "", kubelet creates the subPath without issue. If I set storageClass to the efs csi driver, kubelet complains that the subpath doesn't exist. The question, I guess, is why does the default storageClass support subPath creation, and the efs csi driver storage class not?

wongma7:
Could you share your PV, PVC, and Pod yamls if possible, plus the exact "doesn't exist" error ? (storageClass being empty string is different from storageClass being nil i.e. not having storageClass in your PVC at all, so I want to make sure what storage driver/plugin (kubernetes.io/aws-ebs?) exactly has different subPath behavior. )

[...]

PR #640, opened 2022-02-23
Allow more control of the name of the directory created by the Access Point under Dynamic Provisioning

### LINKED ISSUE #543: Dynamic basePath

jonathanrainer:
This is a continuation of #568, as that PR was opened incorrectly and messing around with it was much more trouble than just opening a new PR

***Is this a bug fix or adding new feature?***
This adds a new feature, mostly in response to discussions on Issue https://github.com/kubernetes-sigs/aws-efs-csi-driver/issues/543

***What is this PR about? / Why do we need it?***
At the moment if you use dynamic provisioning the name of the folder created on the EFS is the name of the generated PV however in some use cases this isn't desirable and leads to a lot of confusion. Under this scheme the name of the directory can be controlled by users using a new parameter subPathPattern. This allows users a higher level of control over the directories that are created as they can specify any combination of fixed strings and a set of predefined variables based on what the CSI Spec supports. To stop this causing problems with security, there is additionally a new featureFlag added called `ensureUniqueDirectories` which appends a UUID to the created structure so clashes will be avoided. This is true by default but allows users to explicitly opt into an unsafe mode if they wish to provide the guarantee that there will be not be clashes themselves rather than having the driver enforce it for them.

***What testing is done?***
I have added to the unit tests for the module and I believe this is sufficient because all that is being changed is the directory being added to the AccessPointOpts struct. There might want to be some E2E testing about what happens if that directory already exists but I'm happy to take pointers on that as this is my first PR to the repo.

fixes #543

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -26,19 +26,21 @@ The following CSI interfaces are implemented:
 * Identity Service: GetPluginInfo, GetPluginCapabilities, Probe
 
 ### Storage Class Parameters for Dynamic Provisioning
-| Parameters          | Values | Default | Optional  | Description |
-|---------------------|--------|---------|-----------|-------------|
-| provisioningMode    | efs-ap |         | false     | Type of volume provisioned by efs. Currently, Access Points are supported. |
-| fileSystemId        |        |         | false     | File System under which access points are created. | 
-| directoryPerms      |        |         | false     | Directory permissions for [Access Point root directory](https://docs.aws.amazon.com/efs/latest/ug/efs-access-points.html#enforce-root-directory-access-point) creation. |
-| uid                 |        |         | true      | POSIX user Id to be applied for [Access Point root directory](https://docs.aws.amazon.com/efs/latest/ug/efs-access-points.html#enforce-root-directory-access-point) creation and for [user identity enforcement](https://docs.aws.amazon.com/efs/latest/ug/efs-access-points.html#enforce-identity-access-points). |
-| gid                 |        |         | true      | POSIX group Id to be applied for [Access Point root directory](https://docs.aws.amazon.com/efs/latest/ug/efs-access-points.html#enforce-root-directory-access-point) creation and for [user identity enforcement](https://docs.aws.amazon.com/efs/latest/ug/efs-access-points.html#enforce-identity-access-points). |
-| gidRangeStart       |        | 50000   | true      | Start range of the POSIX group Id to be applied for [Access Point root directory](https://docs.aws.amazon.com/efs/latest/ug/efs-access-points.html#enforce-root-directory-access-point) creation and for [user identity enforcement](https://docs.aws.amazon.com/efs/latest/ug/efs-access-points.html#enforce-identity-access-points). Not used if uid/gid is set. For user identity enforcement, this value will be applied as both the uid and the gid. |
-| gidRangeEnd         |        | 7000000 | true      | End range of the POSIX group Id. Not used if uid/gid is set. |
-| basePath            |        |         | true      | Path under which access points for dynamic provisioning is created. If this parameter is not specified, access points are created under the root directory of the file system |
-| az                  |        |   ""    | true      | Used for cross-account mount. `az` under storage class parameter is optional. If specified, mount target associated with the az will be used for cross-account mount. If not specified, a random mount target will be picked for cross account mount |
+| Parameters            | Values | Default         | Optional | Description                                                                                                                                                                                                                                                                                                                                                                                   |
+|-----------------------|--------|-----------------|----------|-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| provisioningMode      | efs-ap |                 | false    | Type of volume provisioned by efs. Currently, Access Points are supported.                                                                                                                                                                                                                                                                                                                    |
+| fileSystemId          |        |                 | false    | File System under which access points are created.                                                                                                                                                                                                                                                                                                                                            | 
+| directoryPerms        |        |                 | false    | Directory permissions for [Access Point root directory](https://docs.aws.amazon.com/efs/latest/ug/efs-access-points.html#enforce-root-directory-access-point) creation.                                                                                                                                                                                                                       |
+| uid                   |        |                 | true     | POSIX user Id to be applied for [Access Point root directory](https://docs.aws.amazon.com/efs/latest/ug/efs-access-points.html#enforce-root-directory-access-point) creation.                                                                                                                                                                                                                 |
+| gid                   |        |                 | true     | POSIX group Id to be applied for [Access Point root directory](https://docs.aws.amazon.com/efs/latest/ug/efs-access-points.html#enforce-root-directory-access-point) creation.                                                                                                                                                                                                                |
+| gidRangeStart         |        | 50000           | true     | Start range of the POSIX group Id to be applied for [Access Point root directory](https://docs.aws.amazon.com/efs/latest/ug/efs-access-points.html#enforce-root-directory-access-point) creation. Not used if uid/gid is set.                                                                                                                                                                 |
+| gidRangeEnd           |        | 7000000         | true     | End range of the POSIX group Id. Not used if uid/gid is set.                                                                                                                                                                                                                                                                                                                                  |
+| basePath              |        |                 | true     | Path under which access points for dynamic provisioning is created. If this parameter is not specified, access points are created under the root directory of the file system                                                                                                                                                                                                                 |
+| subPathPattern        |        | `/${.PV.name}`  | true     | The template used to construct the subPath under which each of the access points created under Dynamic Provisioning. Can be made up of fixed strings and limited variables, is akin to the 'subPathPattern' variable on the [nfs-subdir-external-provisioner](https://github.com/kubernetes-sigs/nfs-subdir-external-provisioner) chart. Supports `.PVC.name`,`.PVC.namespace` and `.PV.name` |
+| ensureUniqueDirectory |        | true            | true     | **NOTE: Only set this to false if you're sure this is the behaviour you want**.<br/> Used when dynamic provisioning is enabled, if set to true, appends the a UID to the pattern specified in `subPathPattern` to ensure that access points will not accidentally point at the same directory.                                                                                                |
+| az                    |        | ""              | true     | Used for cross-account mount. `az` under storage class parameter is optional. If specified, mount target associated with the az will be used for cross-account mount. If not specified, a random mount target will be picked for cross account mount                                                                                                                                          |
 
-**Note**  
+**Note**
 * Custom Posix group Id range for Access Point root directory must include both `gidRangeStart` and `gidRangeEnd` parameters. These parameters are optional only if both are omitted. If you specify one, the other becomes mandatory.
 * When using a custom Posix group ID range, there is a possibility for the driver to run out of available POSIX group Ids. We suggest ensuring custom group ID range is large enough or create a new storage class with a new file system to provision additional volumes. 
 * `az` under storage class parameter is not be confused with efs-utils mount option `az`. The `az` mount option is used for cross-az mount or efs one zone file system mount within the same aws account as the cluster.
@@ -179,13 +181,13 @@ This procedure requires Helm V3 or later. To install or upgrade Helm, see [Using
    helm repo add aws-efs-csi-driver https://kubernetes-sigs.github.io/aws-efs-csi-driver/
    ```
 
-1. Update the repo.
+2. Update the repo.
 
    ```sh
    helm repo update aws-efs-csi-driver
    ```
 
-1. Install a release of the driver using the Helm chart.
+3. Install a release of the driver using the Helm chart.
 
    ```sh
    helm upgrade --install aws-efs-csi-driver --namespace kube-system aws-efs-csi-driver/aws-efs-csi-driver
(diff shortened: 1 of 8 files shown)


jonathanrainer:
Right, sorry about the PR wrangling, but now we're back I can continue the conversation from #568.

@iAnomaly, @wolffberg - This is where the PR continues. @wolffberg I've addressed your points about validating the created EFS by explicitly validating it before we send it off the API and have added tests to that effect.

@wongma7 - I had a long think about your comments re: security and think I've come up with a compromise. My general view is always that things should be secure by default but should allow you to turn off those controls in the knowledge that you're explicitly opting in to having to manage those guarantees yourself. So to that end I've introduced another parameter `ensureUniqueDirectories` which adds a UUID to the end of any directory path specified to ensure that directories are unique. This is on by default so I imagine most users won't notice. You can turn this off to get the behaviour originally specified but throughout the README's and examples it makes it clear that then you are opting in to the responsibility to manage the guarantees yourself. This was it's not a surprise for users and it would require you to make a bad change in two places for something bad to happen.

I hope this is satisfactory and if not I'm more than happy to have a discussion about it. Further I realise some of the Go may not be 100% correct as I'm fairly new to the language and some of the most common idioms still elude me so feel free to comment away but hopefully this is a much better proposition than that proposed previously.

docs/README.md
```diff
@@ -26,18 +26,19 @@ The following CSI interfaces are implemented:
...
+| subPathPattern        |        | `/${.PV.name}`  | true     | The template used to construct the subPath under which each of the access points created under Dynamic Provisioning. Can be made up of fixed strings and limited variables, is akin to the 'subPathPattern' variable on the [nfs-subdir-external-provisioner](https://github.com/kubernetes-sigs/nfs-subdir-external-provisioner) chart. Supports `.PVC.name`,`.PVC.namespace` and `.PV.name` |
+| ensureUniqueDirectory |        | true            | true     | **NOTE: Only set this to false if you're sure this is the behaviour you want**.<br/> Used when dynamic provisioning is enabled, if set to true, appends the a UID to the pattern specified in `subPathPattern` to ensure that access points will not accidentally point at the same directory.                                                                                                |
```
iAnomaly, line 40: Minor typo? Should `appends the a UID to the pattern specified` say `appends the UID to the pattern specified`?

iAnomaly (approved):

iAnomaly:
This is awesome @jonathanrainer , really appreciate all your time on this and all the cases you covered with tests. I went looking for the omitted `basePath` empty `subPathPattern` which achieves the AWS EFS Console default behavior of just `/` as the root path for the Access Point and sure enough you've got that in there.

Thanks again, this will be an awesome flexible feature once merged.

wongma7:
/ok-to-test

I will not have time to review for a while but in principle yes  I want this merged, thanks for patience

docs/README.md
jonathanrainer: I went with `appends a UUID to the pattern specified` but you're right it's definitely a typo!

thomaspaulin:
I've run into a use case where these changes would be highly desirable. Thanks team!

examples/kubernetes/dynamic_provisioning/README.md
```diff
@@ -20,13 +20,23 @@ parameters:
   gidRangeStart: "1000"
   gidRangeEnd: "2000"
   basePath: "/dynamic_provisioning"
+  subPathPattern: "${.PVC.namespace}/${PVC.name}"
```
wolffberg, line 23: I think you're missing a `.` in `${PVC.name}`

jonathanrainer:
Just to prove this actually works, I spun up this on an EFS cluster I built with the following storageClass

```
allowVolumeExpansion: true
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  creationTimestamp: "2022-06-12T15:32:21Z"
  labels:
    provisioning-type: dynamic
  name: efs-dynamic
  resourceVersion: "14170"
  uid: 470ddcd4-4df4-42b5-a691-e99520c1a708
parameters:
  basePath: /dynamic
  directoryPerms: "777"
  ensureUniqueDirectories: "false"
  fileSystemId: fs-05427a7b4af6c1fcc
  gidRangeEnd: "2000"
  gidRangeStart: "1000"
  provisioningMode: efs-ap
  subPathPattern: /dynamic/${.PVC.name}/foo
provisioner: efs.csi.aws.com
reclaimPolicy: Delete
volumeBindingMode: Immediate
```

Which managed to create the following directory on disk (apologies for the long directory name, just wanted to stress test it a little bit):
Screenshot 2022-06-12 at 16 34 22

jonathanrainer:
Just to prove this actually works, I spun this up on an EKS cluster with the following StorageClass

```
allowVolumeExpansion: true
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  creationTimestamp: "2022-06-12T15:46:35Z"
  labels:
    provisioning-type: dynamic
  name: efs-dynamic
  resourceVersion: "16864"
  uid: 983866ac-c731-45ac-a22d-1d1edaf7dbda
parameters:
  basePath: ""
  directoryPerms: "777"
  ensureUniqueDirectory: "false"
  fileSystemId: fs-05427a7b4af6c1fcc
  gidRangeEnd: "2000"
  gidRangeStart: "1000"
  provisioningMode: efs-ap
  subPathPattern: /dynamic/${.PVC.name}/foo
provisioner: efs.csi.aws.com
reclaimPolicy: Delete
volumeBindingMode: Immediate
```

Which then generated the following on disk:
Screenshot 2022-06-12 at 16 53 07

So my Unit Test predictions were good and it also showed the value of all the error handling as I got the StorageClass wrong a few times and every time my error handling directed me to the issue quickly.

andre-lx:
Hey @jonathanrainer 

Here at YData we are testing your solution, and fix a big issue for our problem, so thank you and hope this PR get merged soon in order to use the official version. 

Our storage class looks like this:

```
kind: StorageClass
apiVersion: storage.k8s.io/v1
metadata:
  name: shared-sc
provisioner: efs.csi.aws.com
parameters:
  provisioningMode: efs-ap
  fileSystemId: fs-xxxxx
  directoryPerms: "0770"
  gid: 1000
  uid: 1000
  subPathPattern: "/efs"
  ensureUniqueDirectory: "false"
reclaimPolicy: Retain
volumeBindingMode: Immediate
```

About the issue:  https://github.com/kubernetes-sigs/aws-efs-csi-driver/issues/715

We are testing, and we notice that new access points are created each time. For example, here where we have multiple PVC's and each PVC using the same directory.

image


About this issue: https://github.com/kubernetes-sigs/aws-efs-csi-driver/issues/516

We found out, and maybe is worthy of put in the docs, is that with SubPath we can't write to the root path:

SubPaths in the pods will not work with:
```
  subPathPattern: "/efs"
  ensureUniqueDirectory: "false"
```

But after setting as `/efs`everything works great.

jonathanrainer:
Hi @andre-lx, thanks for the very kind words,

With regards your first point, I'd never really considered that access points could be re-used, but you make a very fair point. I'll need to be a bit careful about that intersects with other features, for example if you use the GID range features then you would need a new access point each time because each one would be for a different GID but I don't think it's beyond the realms of possibility to check that is the case, it's also probably a case where adding an E2E test would be a good idea so thanks for spotting that. 

In regards the second point that seems like a bug/gap in my unit testing. I'm going to have a good look at that and hopefully they'll be a fix fairly soon. I'm quite busy this week but hopefully will get a chance to look over this at the weekend and will post the results of my findings when I have them.

Thanks so much for taking an interest, and I look forward to getting to the bottom of these issues.

jonathanrainer:
Hi @andre-lx,

I've looked into both the issues you've raised and I'll respond to them in turn:

1. With the issue around setting 
```
subPathPattern: "/"
ensureUniqueDirectory: "false"
```
I've run some tests and it appears the code is operating exactly as I'd expect, i.e. the request to AWS requests an accessPoint that's root directory is "/". What error are you seeing when this happens? And further what use case does this service? Setting the subPath pattern like that would simply allow you to gain access to the root of the directory, which I guess is useful, but if that's what you want then why not just create a PV/PVC statically and mount that PVC into multiple pods? It seems simpler and like it would accomplish what you want without needing this feature. This might also be helped by the work done in #732.

2. With the issue around re-use of access points, I've had a look into this and I'm not sure it's a good idea, for one thing it introduces the need for concurrency controls and entails quite a lot of book-keeping on the drivers side to make sure the set of access points is correct with respect to the cluster. You have to guarantee that you can lock access to the AWS API so that two concurrent requests from a PVC end up sharing one and also you'd have to ensure that when one PVC got deleted the access point didn't disappear from the other PVCs that were using it.

However I think that the recent PR I've raised in #732 might help you out with the problem you're having, and that one combined with this one will give a similar kind of functionality without depleting the access points available. If this is something that would still be useful then it's probably something a future PR could tackle but I think it's scope creep for this current PR.

gfafonso:
Hey! This is highly desirable by my side... Any time frame on it going merged?

tgeo-cambrian:
Hello, any news on merging this?
This will save a lot of trouble. Currently sharing an EFS access point for shared services let's say in a cluster, leads us to create a PV and a PVC in each of the 30 namespaces (one per microservice) that reside in a cluster.

If this could get merged, we can use a shared services storage class with dynamic provisioning instead and a fixed subpath so we could create all those dynamically using only the PVC's and reducing the clutter.

luismy:
Any news on this one? This will also save us loads of work

jonathanrainer:
@Ashley-wenyizha Sorry I've been so long to sort this out, am really eager to get this merged soon if we can and happy to do whatever's necessary to make that happen :)

luismy:
Exciting times! So looking forward to this!

jz0912:
Are there any updates? I'm waiting to be merged.

jonathanrainer:
Hi @Ashley-wenyizha, I know you said https://github.com/kubernetes-sigs/aws-efs-csi-driver/pull/732#issuecomment-1536595665 that this issue would be progressing as well, has there been any further progress? More than happy to help in any way I can :)

SpoddyCoder:
Thought it would be useful just to point out this PR resolves a Disaster Recovery issue that can arise when using dynamic provisioning.

Consider - a cluster running disparate workloads, with dynamic PVC's defined in mutiple repo's. Now imagaine the cluster completely dies and requires rebuilding all the deployments from manifests. The present dynamic provisioning model would be very awkward to reconnect the EFS state directories back up to the correct deployment. You'd need to modify all the PVC's (in all their different repos) and essentially convert them to static provisioning to define the path where the files still exist. In which case you may as well just start with static provisioning model....

Thanks to the people who have put considerable effort into this PR - hope it gets merged soon.

pkg/driver/controller.go
```diff
@@ -236,8 +247,41 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
...
 	rootDirName := volName
-	rootDir := basePath + "/" + rootDirName
+	// Check if a custom structure should be imposed on the access point directory
+	if value, ok := volumeParams[SubPathPattern]; ok {
+		// Try and construct the root directory and check it only contains supported components
+		val, err := interpolateRootDirectoryName(value, volumeParams)
+		if err == nil {
+			klog.Infof("Using user-specified structure for access point directory.")
+			rootDirName = val
+			if value, ok := volumeParams[EnsureUniqueDirectory]; ok {
```
Ashley-wenyizha, line 262: When we release this new feature and customer upgraded to adopt this feature, controller can not really find the subPathPattern or EnsureUniqueDirectory in its existing volume param.

From the logic here, it seems that will still use PV name for access point directory like before correct?

So theoretically this should not break any existing customer use cases

jonathanrainer: Yes that's correct, once this is released a customer has to explicitly opt into it for it to become active, by setting a subPath pattern. This was one of the key reasons for using a newly defined key so that this behaviour is an enhancement rather than replacement for what the driver already does. 

`ensureUniqueDirectory` is the same, setting it on its own won't do anything, only when the combination of `subPathPattern` and `ensureUniqueDirectory` are set does it work. If neither are set then the code behaves exactly as it was before, with respect to the behaviour of basePath etc. So yes when this releases unless a customer actively changes their StorageClass to turn this new behaviour on they won't notice any difference.

Ashley-wenyizha:
Hi Jonathan @jonathanrainer,

Thanks for the patience and apologize about the delay, this PR lgtm, could you give it a rebase and squash into 1-3 commits? We can merge this afterwards and you can continue to revise PR 732

thanks!

harshad-jpmc:
bump!

Ashley-wenyizha:
Hi Jonathan @jonathanrainer,

this PR lgtm, could you give it a pull + rebase and squash into 1-3 commits? We can merge this afterwards and you can continue to revise PR 732

thanks!

jonathanrainer:
@Ashley-wenyizha Hi, sorry for the delay, real life stuff got a bit on top of me! But have done what you wanted now and this is ready to merge as and when you'd like :)

jonathanrainer:
@Ashley-wenyizha FYI I had to make a small change to the sanity tests because the EFS API Request won't accept paths more than 100 characters but the sanity attempts to use a volume name of 128 characters. In this case I think it's fine because this is the mechanism by which you'd get around exactly this problem but if you think there's a better solution then let me know

joebowbeer:
@jonathanrainer It looks like the new properties (`subPathPattern`, `ensureUniqueDirectory`) were released with helm chart 2.4.9, but **_not_** in the driver appVersion 1.6.0 that this helm chart uses.

When will these properties be supported in a released appVersion?

Also, note the typo in the helm docs: `s/ensureUniqueDirectories/ensureUniqueDirectory/`

PR #644, opened 2022-03-11
Add -v to controller liveness-probe container args

alex-shafer-1001:

Bug fix -- this flag is consistently applied to other Pods and containers, but not this one.

Liveness probes make very noisy logs

**What testing is done?** 

None, assumed to work as expected based on ample precedent.


alex-shafer-1001:
Getting the corporate CLA done is going to be too much effort, sorry.

Issue #489, opened 2021-06-10
Missing permission in the example IAM policy file


foreignmeloman:
/kind bug

**What happened?**
Attaching [provided](https://github.com/kubernetes-sigs/aws-efs-csi-driver/blob/master/docs/iam-policy-example.json) IAM policy example to an EC2 instance did not grant enough permissions for it to mount the EFS and create a PV.
Instead the pod reported the following error:
```
Output: Failed to resolve "fs-bdfc7009.efs.us-east-1.amazonaws.com". The file system mount target ip address cannot be found, please pass mount target ip address via mount options. 
User: arn:aws:sts::888888888888:assumed-role/LordOfTheVolumes/i-0dc4f12f9fdb8dcd3 is not authorized to perform: elasticfilesystem:DescribeMountTargets on the specified resource
```

**What you expected to happen?**
I expected the provisioner to create a PV for the pod.

**How to reproduce it (as minimally and precisely as possible)?**
Just use the current version of [docs/iam-policy-example.json](https://github.com/kubernetes-sigs/aws-efs-csi-driver/blob/master/docs/iam-policy-example.json) with [examples/kubernetes/dynamic_provisioning](https://github.com/kubernetes-sigs/aws-efs-csi-driver/tree/master/examples/kubernetes/dynamic_provisioning).

**Anything else we need to know?**:
I've actually found the solution but will leave it for the maintainer to commit as I am too lazy to sign the CLA.
The error message above actually gives the solution. The first object of  the "Statement" list should look like this:

```json
        {
            "Effect": "Allow",
            "Action": [
                "elasticfilesystem:DescribeAccessPoints",
                "elasticfilesystem:DescribeFileSystems",
                "elasticfilesystem:DescribeMountTargets"
            ],
            "Resource": "*"
        },
```
After I made this change to the policy the dunamic provisioning started to work as expected.

**Environment**
- Kubernetes version (use `kubectl version`):
```
Client Version: version.Info{Major:"1", Minor:"21", GitVersion:"v1.21.1", GitCommit:"5e58841cce77d4bc13713ad2b91fa0d961e69192", GitTreeState:"clean", BuildDate:"2021-05-12T14:18:45Z", GoVersion:"go1.16.4", Compiler:"gc", Platform:"linux/amd64"}
[...]

PR #647, opened 2022-03-22
URGENT / Docs: Fixing IAM policy example, missing required permissions

AndrewFarley:
Urgent documentation update

Without this, using the default documentation role this service does not function properly

**What testing is done?**
Install EFS driver with recommended permissions, it does not work, add permissions, it magically works.  Get frustrated that no one has updated the documentation, submitted this PR so no one else has to be so frustrated as me.

Ref: #489

### DIFF
--- a/docs/iam-policy-example.json
+++ b/docs/iam-policy-example.json
@@ -5,7 +5,9 @@
       "Effect": "Allow",
       "Action": [
         "elasticfilesystem:DescribeAccessPoints",
-        "elasticfilesystem:DescribeFileSystems"
+        "elasticfilesystem:DescribeFileSystems",
+        "elasticfilesystem:DescribeMountTargets",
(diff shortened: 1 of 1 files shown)


AndrewFarley:
Uhm,  is this project abandonware?

Ashley-wenyizha:
/lgtm

Add @wongma7 for review

vmga-crisnevares:
@wongma7 This has been fixed in the master branch, but none of the subsequent releases ([v1.3.8](https://github.com/kubernetes-sigs/aws-efs-csi-driver/blob/v1.3.8/docs/iam-policy-example.json), [helm-chart-aws-efs-csi-driver-2.2.6](https://github.com/kubernetes-sigs/aws-efs-csi-driver/blob/helm-chart-aws-efs-csi-driver-2.2.6/docs/iam-policy-example.json)) include this fix.

PR #651, opened 2022-03-25
Check if publish target path already mounted to be idempotent

wongma7:

While debugging some kubelet issue and restarting kubelet a couple of times I noticed some errors from the driver like 
`x is already mounted or y busy`. kubelet might call NodePublish on the same volume multiple times, the spec says the driver must be idempotent. So to be idempotent, no-op if a target path is already mounted instead of printing errors.

The implementation is copied from EBS driver https://github.com/kubernetes-sigs/aws-ebs-csi-driver https://github.com/kubernetes-sigs/aws-ebs-csi-driver/pull/1019

**What testing is done?** 
CI, plus WIP manual test to reproduce the issue.

### DIFF
--- a/hack/update-gomock
+++ b/hack/update-gomock
@@ -13,10 +13,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-
 set -euo pipefail
 
 IMPORT_PATH=github.com/kubernetes-sigs/aws-efs-csi-driver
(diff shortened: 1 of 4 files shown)


wongma7:
node.go:214: Unexpected call to *mocks.MockMounter.IsLikelyNotMountPoint

have to update all tests to expect it.

PR #654, opened 2022-04-04
[release-1.3] Add release changelog auto-generatoin script  

Ashley-wenyizha:
**Is this a bug fix or adding new feature?**
 N/A
**What is this PR about? / Why do we need it?**
Add release changelog auto-generatoin script
**What testing is done?** 
Local test by executing CLI `python3 hack/release --github-user=ashley-wenyizha --github-token=$TOKEN note --since v1.3.6 --branch release-1.3`

Output
```
wenyizha@3c22fb3758da aws-efs-csi-driver % python3 hack/release --github-user=ashley-wenyizha --github-token=$TOKEN note --since v1.3.6 --branch release-1.3

* Release v1.3.6 part 2/3 ([#625](https://github.com/kubernetes-sigs/aws-efs-csi-driver/pull/625), [@Ashley-wenyizha](https://github.com/Ashley-wenyizha))
* go.mod: fix non-existing k8s.io/kubernetes version ([#645](https://github.com/kubernetes-sigs/aws-efs-csi-driver/pull/645), [@bertinatto](https://github.com/bertinatto))
```

PR #655, opened 2022-04-04
changelog/readme/makefile udpate for v1.3.7 release

Ashley-wenyizha:
Release note of v1.3.7
changelog/readme/makefile udpate for v1.3.7 release
**What testing is done?**

### DIFF
--- a/CHANGELOG-1.x.md
+++ b/CHANGELOG-1.x.md
@@ -1,3 +1,6 @@
+# V1.3.7
+* go.mod: fix non-existing k8s.io/kubernetes version ([#645](https://github.com/kubernetes-sigs/aws-efs-csi-driver/pull/645), [@bertinatto](https://github.com/bertinatto))
+* New efs-utils version of v1.32.1 (https://github.com/aws/efs-utils/releases/tag/1.32.1)
 # v1.3.6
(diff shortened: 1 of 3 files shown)


Cappuccinuo:
@Ashley-wenyizha There is [watchdog configuration change in efs-utils 1.32.1](https://github.com/aws/efs-utils/blob/master/dist/efs-utils.conf#L67-L70), do we need to update driver side https://github.com/kubernetes-sigs/aws-efs-csi-driver/blob/master/pkg/driver/efs_watch_dog.go as well?

wongma7:
just waiting for
https://github.com/kubernetes-sigs/aws-efs-csi-driver/pull/654
https://github.com/kubernetes-sigs/aws-efs-csi-driver/pull/662
/retest

The efs-utils version is dictated at time the image is built, so if u pull 1.3.6 now it has 1.31.3, so I do not think it is necessary to pin efs-utils. Rather we can continue the current model where we just pull whatever happens to be newest at time the image is built, with some exceptions (e.g. when there was some bug in version Y and we deliberately pinned to older version X to rollback)

~ $ docker run --entrypoint sh -it amazon/aws-efs-csi-driver:v1.3.6
sh-4.2# yum info amazon-efs-utils
Loaded plugins: ovl, priorities
Repodata is over 2 weeks old. Install yum-cron? Or run: yum makecache fast
Installed Packages
Name        : amazon-efs-utils
Arch        : noarch
Version     : 1.31.3

PR #663, opened 2022-04-05
Revise to add new params in efs-utils config rendering template 

Ashley-wenyizha:
N/A

Revise to add new params in efs-utils config rendering template to support this code change from utils side - https://github.com/aws/efs-utils/blob/master/dist/efs-utils.conf#L67-L70

[efs-utils] original issue description - https://github.com/aws/efs-utils/issues/126

[efs-utils] ^^ commit - https://github.com/aws/efs-utils/commit/443476f84d03f53f9c11f31a110681ebc31623cf

What testing is done?


wongma7:
/hold

for squashing

PR #664, opened 2022-04-06
Post-release version(v1.3.7) update

Ashley-wenyizha:

**What testing is done?**


Cappuccinuo (approved):
/lgtm

charts/aws-efs-csi-driver/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v2
 name: aws-efs-csi-driver
 version: 2.2.3
```
wongma7, line 3: this should be bumped to 2.2.4

wongma7: or I guess 2.2.5. 

since master branch is source of truth wrt helm charts

PR #665, opened 2022-04-06
Post-release version(v1.3.7) update

Ashley-wenyizha:

**What testing is done?**

### DIFF
--- a/CHANGELOG-1.x.md
+++ b/CHANGELOG-1.x.md
@@ -1,3 +1,6 @@
+# V1.3.7
+* go.mod: fix non-existing k8s.io/kubernetes version ([#645](https://github.com/kubernetes-sigs/aws-efs-csi-driver/pull/645), [@bertinatto](https://github.com/bertinatto))
+* New efs-utils version of v1.32.1 (https://github.com/aws/efs-utils/releases/tag/v1.32.1)
 # v1.3.6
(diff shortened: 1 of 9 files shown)


charts/aws-efs-csi-driver/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v2
 name: aws-efs-csi-driver
 version: 2.2.4
```
wongma7, line 3: this should be bumped to 2.2.5

CHANGELOG-1.x.md
```diff
@@ -1,3 +1,6 @@
+# V1.3.7
+* go.mod: fix non-existing k8s.io/kubernetes version ([#645](https://github.com/kubernetes-sigs/aws-efs-csi-driver/pull/645), [@bertinatto](https://github.com/bertinatto))
+* New efs-utils version of v1.32.1 (https://github.com/aws/efs-utils/releases/tag/1.32.1)
```
wongma7, line 3: include change https://github.com/kubernetes-sigs/aws-efs-csi-driver/pull/666/files here as well ?

charts/aws-efs-csi-driver/Chart.yaml
wongma7: (otherwise chart will not be released)

PR #674, opened 2022-04-21
Disable getting all secrets from ns by default

Ashley-wenyizha:
Disable getting all secrets from namespace by default due to security concern

**What testing is done?**


wongma7:
/ok-to-test
/lgtm
/approve

i would submita pr to master branch as well and update https://github.com/kubernetes-sigs/aws-efs-csi-driver/tree/master/examples/kubernetes/cross_account_mount to say that this permission must be added/uncommented in order for https://github.com/kubernetes-sigs/aws-efs-csi-driver/tree/master/examples/kubernetes/cross_account_mount to work

PR #681, opened 2022-04-24
Removing Dependency on IMDS, allowing `hostNetwork: true` to be removed

jonathanrainer:
This PR is a fix that allows the driver to keep functioning even in situations where IMDS is not available. It does this by falling back to the Kubernetes API to get the required information if it cannot get it from the EC2Metadata Service.

This PR is in response to #313, and will allow the driver to function in environments that do not have access to IMDS. The code was based heavily on https://github.com/kubernetes-sigs/aws-ebs-csi-driver/pull/907, but refactored slightly to use a provider pattern for the metadata and factor out each part into their own file to allow more ease of testing.

**What testing is done?** 
Have written additional unit tests to cover the change but need some help running the E2E tests as running them on my local setup has been fraught with problems. Any guidance on this (can we run them on the usual E2E infrastructure as a one off perhaps?) would be greatly appreciated.


hack/eksctl-patch.yaml
```diff
@@ -0,0 +1,9 @@
+iam:
+  vpcResourceControllerPolicy: true
+  withOIDC: true
+  serviceAccounts:
+    - metadata:
+        name: efs-csi-controller-sa
+        namespace: kube-system
+      wellKnownPolicies:
+        ebsCSIController: true
```
Ashley-wenyizha, line 9: Should here be [efsCSIController](https://eksctl.io/usage/schema/#iam-serviceAccounts-wellKnownPolicies-efsCSIController) instead?

jonathanrainer: Yes, thought I'd done a find and replace everywhere but clearly not!

pkg/cloud/ec2_metadata_test.go
```diff
@@ -0,0 +1,119 @@
...
+
+	"github.com/aws/aws-sdk-go/aws/ec2metadata"
+	"github.com/golang/mock/gomock"
+
+	"github.com/kubernetes-sigs/aws-efs-csi-driver/pkg/cloud/mocks"
+)
+
+var (
+	stdInstanceID       = "instance-1"
+	stdRegion           = "instance-1"
```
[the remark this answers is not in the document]
jonathanrainer: No worries, the codebase needs to be understandable! Have changed this

pkg/cloud/cloud.go
```diff
@@ -101,7 +104,21 @@ type cloud struct {
 // It panics if session is invalid
 func NewCloud() (Cloud, error) {
 	sess := session.Must(session.NewSession(&aws.Config{}))
-	metadata, err := NewMetadataService(sess)
+	svc := ec2metadata.New(sess)
+	api, err := DefaultKubernetesAPIClient()
+
+	if err != nil && !isDriverBootedInECS() {
+		return nil, fmt.Errorf("could not instantiate k8s api client: %v", err)
```
wongma7, line 111: shouldn't error here, to maintain backwards compatibility the driver should tolerate k8s api being unavailable. it should only require kubernetes API to be available after it has determined the other sources are unavailable

Makefile
```diff
@@ -71,6 +71,20 @@ test-e2e:
 	GINKGO_FOCUS="\[efs-csi\]" \
 	./hack/e2e/run.sh
 
+.PHONY: test-e2e-external-eks
+test-e2e-external-eks:
```
wongma7, line 75: this works for me. however if you want to trick the CI into executing it, remove the `test-e2e` rule above and rename this to `test-e2e` because basically what the CI does is run `make test-e2e`. Then in the logs we should be able to validate that an EKS cluster with IMDS disabled passed the test

wongma7: reference: https://github.com/kubernetes/test-infra/blob/5f5d5a6d1130e44c28f4481d47ac00c55ca50f55/config/jobs/kubernetes-sigs/aws-efs-csi-driver/aws-efs-csi-driver-presubmits.yaml#L57-L59

wongma7: never mind it didn't work either, it only worked up until the test itself then I got 

```

  creating file system
  Unexpected error:
      <*errors.errorString | 0xc000557450>: {
          s: "no subnets found for cluster test-cluster-23625-k8s-local",
      }
      no subnets found for cluster test-cluster-23625-k8s-local
  occurred
```

will debug, I guess there could be a bug here https://github.com/kubernetes-sigs/aws-efs-csi-driver/blob/5e1fcd3e915d62d3b091c6de780ff9e6816f3a7b/test/e2e/cloud.go#L215-L231

pkg/cloud/cloud.go
jonathanrainer: Have fixed this in latest version. Realised the error handling around MetadataProvider wasn't the best either so that's been changed, let me know if that fits your requirements.

pkg/cloud/cloud.go
```diff
@@ -108,13 +108,13 @@ func NewCloud() (Cloud, error) {
 	api, err := DefaultKubernetesAPIClient()
 
 	if err != nil && !isDriverBootedInECS() {
-		return nil, fmt.Errorf("could not instantiate k8s api client: %v", err)
+		klog.Warning("Could not create Kubernetes Client")
```
wongma7, line 111: :+1: but the is error not being printed anymore

also, NewCloudWithRole needs the same update (better yet, refactored so NewCloudWithRole and NewCloud call the same internal function so since they only differ by one line)

Makefile
wongma7: I opened https://github.com/kubernetes-sigs/aws-efs-csi-driver/pull/707 to try to fix this

jonathanrainer:
Have tested this in my own EKS cluster. Did a standard install of the Helm Chart plus my changes and the controller booted fine and could serve both static and dynamically provisioned PVs. The IPs also were pod IPs rather than Node IPs as can be evidenced in the below by comparing the pods IPs to the Node Names.
Screenshot 2022-06-12 at 11 19 13

Makefile
wongma7: https://github.com/kubernetes-sigs/aws-efs-csi-driver/pull/707 has merged, running the tests via the hack scripts should be fixed now although i see below manual tests also succeeded

Makefile
```diff
@@ -71,6 +71,20 @@ test-e2e:
...
+	CLUSTER_TYPE=eksctl \
+	K8S_VERSION="1.20" \
+	DRIVER_NAME=aws-efs-csi-driver \
+	HELM_VALUES_FILE="./hack/values_eksctl.yaml" \
+	CONTAINER_NAME=efs-plugin \
+	TEST_EXTRA_FLAGS='--cluster-name=$$CLUSTER_NAME' \
+	AWS_REGION=us-west-2 \
+	AWS_AVAILABILITY_ZONES=us-west-2a,us-west-2b,us-west-2c \
```
wongma7, line 83: could you add this line https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/Makefile#L205 here as well

It is a hack so that clusters created by the CI IAM role can be cleaned up by my "cleaner" IAM role "Infra-prod-KopsDeleteAllLambdaServiceRoleF1578477-1ELDFIB4KCMXV"

hack/values_ekstcl.yaml
```diff
@@ -0,0 +1,8 @@
+enableVolumeSnapshot: true
+controller:
+  logLevel: 5
+node:
+  logLevel: 5
+serviceAccount:
+  controller:
```
wongma7, line 7: with latest helm chart, should be controller.serviceAccount not serviceAccount.controller similar to the EBS one.
https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/hack/values_eksctl.yaml

Also the enalbeVolumeSnapshot line can be removed

wongma7:
@jonathanrainer need to add https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/charts/aws-ebs-csi-driver/templates/controller.yaml#L93-L96 to the template (and kustomize , by running `make generate-kustomize`)
```
            - name: CSI_NODE_NAME
              valueFrom:
                fieldRef:
                  fieldPath: spec.nodeName
```
```

###
## Printing pod efs-csi-controller-695dc45655-tr8rc efs-plugin container logs
#
I0615 08:52:33.911293       1 config_dir.go:62] Mounted directories do not exist, creating directory at '/etc/amazon/efs'
I0615 08:52:33.912260       1 metadata.go:63] getting MetadataService...
I0615 08:52:37.089439       1 metadata.go:71] retrieving metadata from Kubernetes API
F0615 08:52:37.089492       1 driver.go:56] could not get metadata from AWS: CST_NODE_NAME env var not set 
```
this is good in sense that the CI test is working as intended...metadata is not available so it's looking for nodeName

jonathanrainer:
Thanks that's really useful to know I can pull the pod logs out of prow now, pretty sure I'll be much faster/self-sufficient in future!

jonathanrainer:
@wongma7 Yay it all passed, thanks a lot for your help

wongma7:
THX lgtm, could you squash the commits and i'll drop the "real" lgtm. Ideally one commit with vendor changes and one commit with the rest, but putting it all together in 1 commit is fine as well.

jonathanrainer:
@wongma7 All squashed and ready to go

Issue #692, opened 2022-05-10
Add tags to dynamically created Access Points using YAML file


dheerajgss:
**Is your feature request related to a problem? Please describe.**

Whenever an access point is created dynamically using PV/PVC with the help of the EFS CSI driver, there are no tags associated with it to identify which PVC it belongs to (or) there is no way to add tags using the YAML file to categorize which pods/deployments are using those access points.

For example, if I want to use an access point with one of my application pods while creating a PVC for it, I want to tag that access point from the YAML file itself to know and notify other users that this particular access point is being used by this application pods/deployment.

**Describe the solution you'd like in detail**

Tweaking the YAML file definition to add the 'tags' parameter as well.

**Describe alternatives you've considered**

Open the AWS EFS console, search for the dynamically created access point, and then use the 'Manage Tags' option to add more tags. There is also the 'extraCreateMetadata' option which can be added to the helm chart, but it only provides static tags to all the access points created and does not support adding customized tags as per access point usage.

dheerajgss:
#689 partially covers the requirement, but if I want to categorize my access points in a much-refined way (like application level or pod level as mentioned in my initial description) then these static tags would not help me.

jonathanrainer:
Ah yes I see what you mean. Would you be happy to be able to set the tags per StorageClass? Or do you need it more fine grained than that?

dheerajgss:
StorageClass settings would help I guess. I don't have any ideas about fine graining it as of now😅.

PR #689, opened 2022-05-06
Adding tags to accessPointsOptions

jonathanrainer:
**Is this a bug fix or adding new feature?**
This is a very small new feature, just adding some tags into the Access Point Creation

**What is this PR about? / Why do we need it?**
As per #596 we want to add the tags for the PV name, PVC name and namespace to the created access point. This PR simply adds these tags, very much based on the code in #349 

**What testing is done?** 
#349 was tested against an EKS cluster and the code is identical, plus all unit tests still pass and E2E tests will run as part of this PR

fixes #596

PR #700, opened 2022-05-27
Add Ability to Set Tags on StorageClass, Expand Range of Characters Supported in Tags

### LINKED ISSUE #613: Helm: No support for tags with spaces
/kind bug

**What happened?**

Similar to https://github.com/kubernetes-sigs/aws-efs-csi-driver/issues/533 , if you provide a "controller.tags" value to the Helm chart which has values that contain spaces, you will get an error in the controller's "efs-plugin" container like:

> driver.go:134] Failed to parse input tag string: Application:foo Brand:Value With Spaces LaunchedBy:terraform-api

The tag value eg. `Brand:Value With Spaces` should be interpreted as having the value `"Value With Spaces"` and no error should be logged.

Install the driver by using the Helm chart and providing a tag value with spaces in it. For example, in your values.yaml:

```
controller:
  # Add additional tags to access points
  tags:
    mytag: Value With Spaces
```

It seems like the CSI driver for EBS has a solution for this which works a little better. See https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/charts/aws-ebs-csi-driver/templates/_helpers.tpl it appears to use commas to separate the pairs, and `=` to separate the key/value. For example `<key1>=<value1>,<key2>=<value2>`. It appears to be parsed here https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/25edc9779a6bf8608902908473d3868e05b6efb8/cmd/options/controller_options.go#L41 though I'm not very familiar with Go so I don't really see how that's happening, it looks like it's somewhere in https://github.com/spf13/pflag Edit: actually, it looks like it's in https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/25edc9779a6bf8608902908473d3868e05b6efb8/vendor/k8s.io/component-base/cli/flag/map_string_string.go#L65 although it's not clear to me how one would escape an `=` character in the value, which could also be annoying.

It would certainly make more sense if both of AWS' two CSI drivers followed the same config formatting/parsing approach for parameters that they both share, such as "extra tags".

**Environment**
- Kubernetes version (use `kubectl version`): 1.18
- Driver version: 2.2.1

jonathanrainer:
This allows users to set tags on the StorageClass so they can have more granular tagging for their accessPoints should they desire, this addresses the feature requested in #692. In doing so the parsing of tags has been expanded to include characters like ':' and ' ' and so addresses #613 as well. It doesn't go quite as far as the issue suggests (i.e. aligning with the EBS driver on a standard) but it allows tags to have ':' and ' ' in them with a modicum of effort on the part of the user. That being said if we did decide in future we wanted to align and both use "=" and the key/value separator and "," as the tag separator because they're pulled out into constants that would not be very difficult.

This PR allows users to specify tags on StorageClasses and further, where tags could be specified previously, they can now include  spaces, and colons as well as all the other special characters dictated by the AWS standard: https://docs.aws.amazon.com/general/latest/gr/aws_tagging.html and https://docs.aws.amazon.com/efs/latest/ug/manage-fs-tags.html

**What testing is done?** 
I have added extra Unit tests to cover this behaviour. If it needs extra E2E tests adding please let me know

fixes #613


jonathanrainer:
I have End-To-End tested this myself in an EKS cluster. Installed the driver with the following helm overrides 

```
controller:
  tags:
    environment: prod
    region: us-east-1
    "'I have spaces'": "'and:I:have:colons'"
```

Then created a StorageClass to facilitate dynamic provisioning like so:

```
allowVolumeExpansion: true
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  creationTimestamp: "2022-06-12T08:40:57Z"
  labels:
    provisioning-type: dynamic
  name: efs-dynamic
  resourceVersion: "13482"
  uid: eb329bc4-8a69-4a02-b523-3648cdb7ec67
parameters:
  basePath: /dynamic
  directoryPerms: "777"
  extraTags: SCTag:'Look at the spaces'
  fileSystemId: fs-0a2f099fa92ff360d
  gidRangeEnd: "2000"
  gidRangeStart: "1000"
  provisioningMode: efs-ap
provisioner: efs.csi.aws.com
reclaimPolicy: Delete
volumeBindingMode: Immediate
```

Created a PVC in the cluster to kickoff dynamic provisioning and it created an Access Point with the following tags:
Screenshot 2022-06-12 at 09 35 18

Then tweaked the Chart Overrides and StorageClass to show the overriding behaviour:

```
controller:
  tags:
    environment: prod
    region: us-east-1
    "'I have spaces'": "'and:I:have:colons'"
    OverrideMe: foo
```
allowVolumeExpansion: true
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  creationTimestamp: "2022-06-12T08:40:57Z"
  labels:
    provisioning-type: dynamic
  name: efs-dynamic
  resourceVersion: "13482"
  uid: eb329bc4-8a69-4a02-b523-3648cdb7ec67
parameters:
  basePath: /dynamic
  directoryPerms: "777"
  extraTags: SCTag:'Look at the spaces' OverrideMe:bar
  fileSystemId: fs-0a2f099fa92ff360d
  gidRangeEnd: "2000"
  gidRangeStart: "1000"
  provisioningMode: efs-ap
provisioner: efs.csi.aws.com
reclaimPolicy: Delete
volumeBindingMode: Immediate
```

And then got the following tags in AWS:
<img width="1331" alt="Screenshot 2022-06-12 at 09 48 05" src="https://user-images.githubusercontent.com/1529660/173225387-17578938-8e22-4d64-83aa-11230eb070fa.png">

Ben-Mark:
Before fixing the conflicts,
will this PR support the below functionality like the ebs storage class has?


ebs storageclass can auto add the tags like this:
parameters:
  tagSpecification_1: "pvc_namespace={{ .PVCNamespace }}"
  tagSpecification_2: "pvc_service={{ .PVCService }}"
  tagSpecification_3: "pvc_name={{ .PVCName }}"
  tagSpecification_4: "pv_name={{ .PVName }}"

would efs storageclass allow it as well like this? 
parameters:
   extraTags: "pvc_namespace={{ .PVCNamespace }} pvc_service={{ .PVCService }} pvc_name={{ .PVCName }} pv_name={{ .PVName }}

jonathanrainer:
@Ben-Mark I'm not sure that the functionality described is the best fit for this PR, but I think #689 does what you want?

jonathanrainer:
@Ashley-wenyizha Could we get an ok to test here as well?

PR #701, opened 2022-05-28
Adding ability to set volMetricsOptIn on Daemonset

### LINKED ISSUE #698: volMetricsOptIn not working correctly in helm chart
/kind bug

**What happened?**
If you set controller.volMetricsOptIn to true, it adds `--vol-metrics-opt-in=true` to controller deployment.
But that parameter has to be set for the nodes.

The helm chart should add `--vol-metrics-opt-in=true` to node daemonset.

jonathanrainer:
This is a bug-fix in response to #698. 

It seems as though `volMetricsOptIn` is set for the controller container when the code that uses the flag isn't executed in that context. I've added the flag to the daemonset as well in the first instance because I'm not 100% sure that the flag is never used in the context of the controller. If that is the case I'm happy to also delete those settings in a separate commit.

**What testing is done?** 
As this is a helm change that is very simple, and replicates code that already exists in the codebase, this is untested other than the automated tests that run as part of the PR. However if there are further tests you would like performing I'm happy to do so.

fixes #698

### DIFF
--- a/charts/aws-efs-csi-driver/templates/node-daemonset.yaml
+++ b/charts/aws-efs-csi-driver/templates/node-daemonset.yaml
@@ -72,6 +72,8 @@ spec:
             - --logtostderr
             - --v={{ .Values.node.logLevel }}
             - --vol-metrics-opt-in={{ hasKey .Values.node "volMetricsOptIn" | ternary .Values.node.volMetricsOptIn false }}
+            - --vol-metrics-refresh-period={{ hasKey .Values.node "volMetricsRefreshPeriod" | ternary .Values.node.volMetricsRefreshPeriod 240 }}
+            - --vol-metrics-fs-rate-limit={{ hasKey .Values.node "volMetricsFsRateLimit" | ternary .Values.node.volMetricsFsRateLimit 5 }}
           env:
             - name: CSI_ENDPOINT
               value: unix:/csi/csi.sock
@@ -166,7 +168,7 @@ spec:
             type: Directory
         - name: efs-state-dir
           hostPath:
-            path: /var/run/efs
+            path: /var/run/efs-csi-driver
             type: DirectoryOrCreate
         - name: efs-utils-config
           hostPath:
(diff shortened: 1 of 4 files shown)


z0rc:
Hey, this change looks simple enough. Can it be reviewed?

makomatic:
Works  as expected if I use it in our QA cluster with `--vol-metrics-opt-in=true` as arg for the daemonset binary.
_kubelet_volume_stats_available_bytes_ is set to EFS max capacity, of course (8388607Ti), but _kubelet_volume_stats_used_bytes_ seems to be valid.

jonathanrainer:
Hi @mskanth972 can we have an `ok-to-test` on this one?

WDaan:
Any update on this one? What is still blocking?

mskanth972:
@jonathanrainer, thanks for the PR. Can you remove that parameter in controller helm chart in a separate commit?

jonathanrainer:
@mskanth972 That's now done, do I need to re-generate the Kustomize as well?

mskanth972:
@jonathanrainer not necessary I believe but if you can that would be great.

jonathanrainer:
@mskanth972 That's now fixed too, I also realised there was an inconsistency between the helm chart and the kustomize so that's now fixed up as well.

charts/aws-efs-csi-driver/templates/node-daemonset.yaml
```diff
@@ -165,7 +168,7 @@ spec:
             type: Directory
         - name: efs-state-dir
           hostPath:
-            path: /var/run/efs
+            path: /var/run/efs-csi-driver
```
mskanth972, line 171: why the path has been changed?

jonathanrainer: So I thought this was really weird, it's an inconsistency between the Kustomize and the Helm Chart. If you look at `master`, the entry in the Helm Chart was added at some point 3 years ago and appears to have never changed. But the Kustomize was changed in PR #792, I assumed since the Kustomize was more up-to-date that was correct but if the Kustomize is incorrect then I'm happy to excise it.

mskanth972: Not a problem. Will merge it

PR #702, opened 2022-05-28
Allow support for custom CA bundles

### LINKED ISSUE #699: Support custom CA bundles in disconnected regions

When using the driver in disconnected partitions/regions, there is a need to mount a volume for a custom CA bundle and sometimes to also set `AWS_CA_BUNDLE` when AWS calls are needed for describing mount target IP addresses. The helm chart does not allow variables and templating for these, and I have resorted to manually editing the `DaemonSet` post-install or relying on a policy engine like Kyverno to inject the changes for me.

The helm chart should expose additional variables such as:
* node.caBundle.name --> name of a config map holding the CA bundle
* node.caBundle.key --> key within the config map holding the CA bundle
* node.caBundle.mountPath --> path within the container to mount the bundle

When these are set, new volumes and volumeMounts are templated into the `DaemonSet` at the given path. Furthermore, `AWS_CA_BUNDLE` env is set to the mount path.

Alternatives have been manual edits and Kyverno mutating policies.

jonathanrainer:
This is a new feature, which allows a custom CA bundle to be specified to the driver, as suggested by @zonybob in #699

When the driver runs in disconnected regions/partitions it needs custom CA certificates to be provided, but at present the driver has no way of doing that. This PR adds that capability and consequently configures the driver to use those certificates.

**What testing is done?** 
Have run the driver through `helm` template to ensure it renders correctly and have installed it into a local KiND cluster to ensure the driver boots but no more testing other than that. @zonybob if you wanted to take this branch and see if it solves your problem/have any comments I'd be eager to hear them!

fixes #699

### DIFF
--- a/charts/aws-efs-csi-driver/templates/node-daemonset.yaml
+++ b/charts/aws-efs-csi-driver/templates/node-daemonset.yaml
@@ -100,10 +100,10 @@ spec:
               mountPath: /var/amazon/efs
             - name: efs-utils-config-legacy
               mountPath: /etc/amazon/efs-legacy
-            {{- if .Values.node.caBundle }}
+          {{- if .Values.node.caBundle }}
             - name: custom-ca-bundle
               mountPath: {{ .Values.node.caBundle.mountPath }}
-            {{- end }}
+          {{- end }}
           ports:
             - name: healthz
               containerPort: {{ .Values.node.healthPort }}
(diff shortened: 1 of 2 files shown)


jonathanrainer:
Hi @mskanth972 could we get an `ok-to-test` on this?

PR #707, opened 2022-06-08
Support e2e test EFS create on EKS clusters by finding EKS node subnets

wongma7:

**What is this PR about? / Why do we need it?** The e2e test can run against an EKS cluster if an EFS file system is provided (i.e., --create-file-system=false). However, if an EFS file system is not provided then the e2e test EFS create fails because it fails to find EKS node subnets which are needed for creating EFS mount targets.

I don't know of a tag that is guaranteed to exist on ALL EKS clusters so I am trying

```

				Name: aws.String("tag:Name"),
				Values: []*string{
					aws.String(fmt.Sprintf("*%s*", clusterName)),
				},
```
eksctl does tag clusters with alpha.eksctl.io/cluster-name but not all EKS clusters are created by eksctl

**What testing is done?** 
- [ ] local test `TEST_ID=1 CLEAN=false make test-e2e
CLUSTER_TYPE=eksctl \
K8S_VERSION="1.20" \
DRIVER_NAME=aws-efs-csi-driver \
HELM_VALUES_FILE="./hack/values_eksctl.yaml" \
CONTAINER_NAME=efs-plugin \
TEST_EXTRA_FLAGS='--cluster-name=$CLUSTER_NAME' \
AWS_REGION=us-west-2 \
AWS_AVAILABILITY_ZONES=us-west-2a,us-west-2b,us-west-2c \
TEST_PATH=./test/e2e/... \
GINKGO_FOCUS="\[efs-csi\]" \
./hack/e2e/run.sh`
- [ ] CI test EKS (trick CI into executing EKS by modifying Make rule)
- [ ] CI test Kops (make sure my refactoring didn't break anything)


wongma7:
EKS test passed 

https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_aws-efs-csi-driver/707/pull-aws-efs-csi-driver-e2e/1534677627753730048

Dockerfile
```diff
@@ -26,9 +26,9 @@ ADD . .
 ARG client_source=k8s
 ENV EFS_CLIENT_SOURCE=$client_source
 
-RUN GOOS=${TARGETOS} GOARCH=${TARGETARCH} make aws-efs-csi-driver
+RUN OS=${TARGETOS} ARCH=${TARGETARCH} make $TARGETOS/$TARGETARCH
```
Ashley-wenyizha, line 29: curious the purpose of the change on make here?

Makefile
```diff
@@ -13,29 +13,90 @@
...
+bin/aws-efs-csi-driver: | bin
+	CGO_ENABLED=0 GOOS=linux GOARCH=$(ARCH) go build -mod=vendor -ldflags ${LDFLAGS} -o bin/aws-efs-csi-driver ./cmd/
+
+.PHONY: all
+all: all-image-docker
+
+.PHONY: all-push
+all-push:
+	docker buildx build \
+		--no-cache-filter=linux-amazon \
```
Ashley-wenyizha, line 68: nice to build the no cache! would this be helpful on avoiding unnecessary image cache triggered CVEs?

hack/e2e/eksctl.sh
```diff
@@ -6,7 +6,7 @@ function eksctl_install() {
   INSTALL_PATH=${1}
   EKSCTL_VERSION=${2}
   if [[ ! -e ${INSTALL_PATH}/eksctl ]]; then
-    EKSCTL_DOWNLOAD_URL="https://github.com/weaveworks/eksctl/releases/download/${EKSCTL_VERSION}/eksctl_$(uname -s)_amd64.tar.gz"
+    EKSCTL_DOWNLOAD_URL="https://github.com/weaveworks/eksctl/releases/download/v${EKSCTL_VERSION}/eksctl_$(uname -s)_amd64.tar.gz"
```
Ashley-wenyizha, line 9: Should we add 'v' here?

looks like +v is 404 not found --> https://github.com/weaveworks/eksctl/releases/download/v0.11.1/eksctl_Darwin_amd64.tar.gz

vs https://github.com/weaveworks/eksctl/releases/download/0.11.1/eksctl_Darwin_amd64.tar.gz is the correct download link

hack/e2e/eksctl.sh
```diff
@@ -68,6 +71,7 @@ function eksctl_create_cluster() {
 
   if [[ "$WINDOWS" == true ]]; then
     ${BIN} create nodegroup \
+      --managed=false \
```
Ashley-wenyizha, line 74: reading this flag, do we change it to false because this is test nodegroups and we do not want to make it show up from the console?

```
setting --managed=false or using the nodeGroups field creates an unmanaged nodegroup. 
Bear in mind that unmanaged nodegroups do not appear in the EKS console, which as a general rule only knows about EKS-managed nodegroups.
```

hack/e2e/run.sh
```diff
@@ -57,7 +57,7 @@ KOPS_STATE_FILE=${KOPS_STATE_FILE:-s3://k8s-kops-csi-e2e}
 KOPS_PATCH_FILE=${KOPS_PATCH_FILE:-./hack/kops-patch.yaml}
 KOPS_PATCH_NODE_FILE=${KOPS_PATCH_NODE_FILE:-./hack/kops-patch-node.yaml}
 
-EKSCTL_VERSION=${EKSCTL_VERSION:-0.56.0-rc.1}
+EKSCTL_VERSION=${EKSCTL_VERSION:-0.69.0}
```
Ashley-wenyizha, line 60: I see, 0.69.0 download link actually needs the v prefix, please ignore previous comment on the download link

Makefile
wongma7: yes, this is to avoid case where somehow you have cache with older yum packages

hack/e2e/eksctl.sh
wongma7: yes, they changed it recently so since we updated to 0.69.0 it's using the v format

hack/e2e/eksctl.sh
wongma7: it's because "Windows is only supported for self-managed (--managed=false flag) nodegroups." (note this is in the WINDOWS clause) . https://eksctl.io/usage/windows-worker-nodes/ . I will add a comment in a follow-up PR for hack/e2e changes

Dockerfile
wongma7: i wanted to make it more consistent with fsx and ebs Makefiles so I copied this pattern from there. Specifically I need the `make image` rule for the e2e test to work beacuse the e2e test needs to build/push an image to a per-test tag and the current rule was pushing it to hardcoded `master` tag which is not usable since multijple tests could be running in parallel.

As for why ebs and fsx are this in the first place, it's because ebs must build for windows so I made a bunch of changes to its Makefile to use these generic OS/ARCH env variables. That comes at cost of readability/simplicity since for efs we only care about arm/x64... so I am open to simplifying the Makefile but probably not in the near future :sweat_smile:

Makefile
Ashley-wenyizha: nice, I will give it a build and push to my dev account to see this can help sort out most of our CVE mitigation

wangnyue (approved):
LGTM

test/e2e/cloud.go
```diff
@@ -270,47 +280,77 @@ func (c *cloud) getKopsSecurityGroup(clusterName string) (string, error) {
...
+	)
+}
+
+func (c *cloud) getEksSubnetIds(clusterName string) ([]string, error) {
+	return c.getFilteredSubnetIds(
+		[]*ec2.Filter{
+			{
+				Name: aws.String("tag:Name"),
+				Values: []*string{
+					aws.String(fmt.Sprintf("*%s*", clusterName)),
```
nckturner, line 327: Are all the eks clusters created with eksctl?

nckturner: Another problem is that you may get more subnets than expected.  In the event that you have `cluster1` and `cluster10`, you will get additional subnets when looking for those that belong to `cluster1`.

wongma7: Updated:

first, call getEksctlSubnetIds. eksctl gives us reliable way to find the subnets "tag:alpha.eksctl.io/cluster-name"

second, call getEksCloudFormationSubnetIds. this is necessarily very generic since the name could be anything but for now i narrowed it to "%s-*", clusterName because that's how **I** name mine

PR #710, opened 2022-06-14
EFS-CSI pod impersonation implementation

lmouhib:
new feature

Add the ability for efs-csi to impersonate pods

**What testing is done?** 
Tested with pods that have service account annotated with a role ARN. The efs-utils assume the role annotated to the service account and is able to mount an access point that is protected with an IAM resource policy.

### DIFF
--- a/charts/aws-efs-csi-driver/templates/csidriver.yaml
+++ b/charts/aws-efs-csi-driver/templates/csidriver.yaml
@@ -8,3 +8,10 @@ metadata:
     "helm.sh/resource-policy": keep
 spec:
   attachRequired: false
+  {{- if .Values.podIAMAuthorization }}
+  podInfoOnMount: true
+  tokenRequests:
+      - audience: "sts.amazonaws.com"
+        expirationSeconds: 3600
+  requiresRepublish: true
+  {{- end}}
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -60,6 +60,7 @@ func main() {
 	if err != nil {
 		klog.Fatalln(err)
 	}
+
 	drv := driver.NewDriver(*endpoint, etcAmazonEfs, *efsUtilsStaticFilesPath, *tags, *volMetricsOptIn, *volMetricsRefreshPeriod, *volMetricsFsRateLimit, *deleteAccessPointRootDir)
 	if err := drv.Run(); err != nil {
 		klog.Fatalln(err)
--- a/examples/kubernetes/pod_iam_impersonation/README.md
+++ b/examples/kubernetes/pod_iam_impersonation/README.md
@@ -1,0 +1,83 @@
+## CSI Driver POD IMPERSONATION
+
+This example shows how you can use the pod impersonation feature available 
+in the CSI Driver to enforce access control when mounting EFS Access point protected with EFS resource policy.
+This feature is for EKS only and leverages IAM roles for service accounts (IRSA), your EKS cluster must be enabled for IRSA. 
+You can follow this [link](https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html) on how to setup and use IRSA.
+
+**Note:** This feature only works when installed with Helm and `podIAMAuthorization` is set to `true`. 
+
+### IMPORTANT
+This feature will not work if you are upgrading the Helm chart. It MUST be a clean install as the CSI Driver object is immutable. 
+The field `podIAMAuthorization` in the HELM chart is set to keep backward compatibility in case you need to upgrade/patch an already existing 
+deployment of the EFS CSI Driver or do not want to use the Pod impersonation feature.
+
+## Example
+
+This example assumes you have an access point set up and is protected with a resource policy. 
+If you would like to learn how to configure IAM policies for access points you can follow this [link](https://docs.aws.amazon.com/efs/latest/ug/efs-access-points.html#access-points-iam-policy).
+
+### Create the Persistent Volume Spec
+
+```
+kind: PersistentVolume
+metadata:
+  name: efs-pv1
+spec:
+  capacity:
+    storage: 5Gi
+  volumeMode: Filesystem
+  accessModes:
+    - ReadWriteMany
+  persistentVolumeReclaimPolicy: Retain
+  storageClassName: efs-sc
+  csi:
+    driver: efs.csi.aws.com
+    volumeHandle: [FileSystemId]::[AccessPointId]
+    volumeAttributes:
+      podIAMAuthorization: "true"
+---
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: efs-claim1
+spec:
+  accessModes:
+    - ReadWriteMany
+  storageClassName: efs-sc
+  resources:
+    requests:
+      storage: 5Gi
+```
+
+Here it is important to set the property `podIAMAuthorization` in `volumeAttribute` to `true`. 
+This indicates to the EFS CSI Driver when mounting the Access Point to use 
+the role annotated to the service account associated with the `pod`.
+
+### Deploy an Application
+
+You can use the sample application in  `examples/kubernetes/pod_iam_impersonation/specs/example.yaml` 
+to create a Persistent Volume (PV), Persistent Volume Claim (PVC) and a sample application. 
+
+**Note:** The pod defined in the example has a service account associated with it. 
+In the example the service account is `efs-app-sa`. The service account needs to be create and annotated with a role 
+that is allowed in the EFS resource policy associated with Access Pointed referenced above in the PV.
+
+Once you have created the service account you can apply the following command:
+
+```sh
+>> kubectl apply -f examples/kubernetes/pod_iam_impersonation/specs/example.yaml
+```
+
+### Check EFS filesystem is used
+After the objects are created, verify the pod is running:
+
+```sh
+>> kubectl get pods
+```
+
+You can also verify that data is being written into the EFS file system.
+
+```sh
+>> kubectl exec -ti efs-app-1 -- tail -f /data-dir1/out.txt
+```
--- a/pkg/driver/node.go
+++ b/pkg/driver/node.go
@@ -18,6 +18,7 @@ package driver
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"os"
 	"path"
@@ -28,8 +29,23 @@ import (
 	"github.com/container-storage-interface/spec/lib/go/csi"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	k8sv1 "k8s.io/client-go/kubernetes/typed/core/v1"
 	"k8s.io/klog"
 )
+
+const (
+	jwtTokenFilePath  = "/home/token"
+	roleAnnotationKey = "eks.amazonaws.com/role-arn"
+	docURL            = "https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html"
+)
+
+type podAuth struct {
+	nameSpace, serviceAccount string
+	k8sClient                 k8sv1.CoreV1Interface
+	ctx                       context.Context
+	jwtToken                  string
+}
 
 var (
 	volumeCapAccessModes = []csi.VolumeCapability_AccessMode_Mode{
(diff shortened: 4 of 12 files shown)


cmd/main.go
```diff
@@ -60,7 +63,18 @@ func main() {
...
+	if err != nil {
+		klog.Fatal(err)
+	}
+
+	clientset, err := kubernetes.NewForConfig(cfg)
+	if err != nil {
+		klog.Fatal(err)
+	}
+
+	drv := driver.NewDriver(*endpoint, etcAmazonEfs, *efsUtilsStaticFilesPath, *tags, *volMetricsOptIn, *volMetricsRefreshPeriod, *volMetricsFsRateLimit, *deleteAccessPointRootDir, clientset.CoreV1())
```
jonathanrainer, line 77: I wonder here what happens if the Kubernetes Client fails to initialise, so clientSet comes back as `nil`. If that happens the driver will crash before it boots, would it be better to pass the clientset through without calling any methods on it and then have that initialise something else (some kind of AuthManager or something like that) that can not do anything if it fails to initialise?

pkg/driver/node.go
```diff
@@ -73,12 +93,41 @@ func (d *Driver) NodePublishVolume(ctx context.Context, req *csi.NodePublishVolu
 	// TODO when CreateVolume is implemented, it must use the same key names
 	subpath := "/"
 	encryptInTransit := true
+	iamAuth := false
```
jonathanrainer, line 96: With regards to this, would it make more sense to feature flag this entire idea, then you can assume that if the user has turned the feature flag on this is what they want which gets rid of the need for this?

pkg/driver/node.go
```diff
@@ -73,12 +93,41 @@ func (d *Driver) NodePublishVolume(ctx context.Context, req *csi.NodePublishVolu
...
+
 		switch strings.ToLower(k) {
+		case "csi.storage.k8s.io/pod.name":
+			continue
+		case "csi.storage.k8s.io/pod.namespace":
+			continue
+		case "csi.storage.k8s.io/pod.uid":
+			continue
+		case "csi.storage.k8s.io/serviceaccount.name":
+			continue
```
jonathanrainer, line 109: Would it be worth collapsing these into one case with commas?

pkg/driver/node.go
```diff
@@ -419,6 +474,28 @@ func parseVolumeId(volumeId string) (fsid, subpath, apid string, err error) {
 	return
 }
 
+func (p podAuth) setPodCredentials() (e error) {
+
+	response, err := k8sv1.CoreV1Interface.ServiceAccounts(p.k8sClient, p.nameSpace).Get(p.ctx, p.serviceAccount, metav1.GetOptions{})
```
jonathanrainer, line 479: I don't think it's enormously important to pass the context through here, you could remove it and then it would make the type simpler, I could be wrong though!

pkg/driver/node.go
```diff
@@ -73,12 +93,41 @@ func (d *Driver) NodePublishVolume(ctx context.Context, req *csi.NodePublishVolu
...
+
+			podauth.setPodCredentials()
+		case "iamauth":
+			var err error
+			iamAuth, err = strconv.ParseBool(v)
+			if err != nil {
+				return nil, status.Error(codes.InvalidArgument, fmt.Sprintf("Volume context property %q must be a boolean value: %v", k, err))
+			}
+		case "csi.storage.k8s.io/ephemeral":
+			continue
```
jonathanrainer, line 127: You could collapse this into the case above as well

pkg/driver/node.go
```diff
@@ -73,12 +93,41 @@ func (d *Driver) NodePublishVolume(ctx context.Context, req *csi.NodePublishVolu
...
+		case "csi.storage.k8s.io/pod.uid":
+			continue
+		case "csi.storage.k8s.io/serviceaccount.name":
+			continue
+		case "csi.storage.k8s.io/serviceaccount.tokens":
+			//TODO get the token in a cleaner way, not like this.
+			result := make(map[string]interface{})
+			json.Unmarshal([]byte(volContext["csi.storage.k8s.io/serviceAccount.tokens"]), &result)
+			tokenInfo := fmt.Sprintf("%v", result["sts.amazonaws.com"].(map[string]interface{})["token"])
+
```
jonathanrainer, line 115: Could result be a `map[string]map[string]string` here? Not the prettiest data type but it avoids the use of `interface{}`. Further I wonder if there's a JWT library that would make this extraction a little bit easier?

pkg/driver/node.go
```diff
@@ -419,6 +474,28 @@ func parseVolumeId(volumeId string) (fsid, subpath, apid string, err error) {
...
+func (p podAuth) setPodCredentials() (e error) {
+
+	response, err := k8sv1.CoreV1Interface.ServiceAccounts(p.k8sClient, p.nameSpace).Get(p.ctx, p.serviceAccount, metav1.GetOptions{})
+	if err != nil {
+		klog.Errorf("Unable to get the service account description %s", err)
+		return err
+	}
+	roleArn := response.Annotations["eks.amazonaws.com/role-arn"]
+
+	errWrite := os.WriteFile("/home/token", []byte(p.jwtToken), 0644)
```
jonathanrainer, line 486: Is it worth having the token location be configurable? (Genuine question I'd be interested in your views)

jonathanrainer (changes requested):
Looks like a really good solution in general! I was working on something similar but thought it'd be better if I got behind what you were doing instead, which looks great. There are a few things to look at but feel free to push back on any of them, they're just my thoughts and I'm not expert.

I'd say a couple of general things as well as the specific points I made in the review:

* It'd be worth looking into feature flagging this whole thing and setting it to false in the Helm Values. This means people have to explicitly opt-into it and will give them time to migrate, one of the most important things in this driver is backwards compatibility and if people have specific stuff setup to do with the serviceAccounts that already exist we want to give them a smooth upgrade path. I'd extend that to the new Helm Resources too.
* I think it would be work extending PodAuth out and split it up into functions a bit more. This will make it easier to Unit Test because you can test one small bit at a time, while still keeping a nice public interface in the `setPodCredentials` function.
* This definitely needs unit tests adding to it, the kubernetes library has a really nice `fake` library that I've added in a few PRs already. It means you can set up a fake clientset and make unit tests that react to it without a lot of effort. I'd look into that.
* Does any work need to be done when `NodeUnpublishVolume` is called?

Let me know if some of that doesn't make sense or you need any help, it's a really great feature to see added to the driver!

pkg/driver/node.go
lmouhib: Ok for the feature flag, because the CSIDriver object is immutable and we need to keep backward compatibility as you mentioned. We need `iamAuth`  there will be  cases where `iam auth` is not needed. The Driver is used across the cluster and you might have a mix of EFS that use IAM or not. If it is set to true, `efs-utils` behavior when provided with `iam` mount option but there is no `IAM` credential available will try to use whatever credential available including instance profile which is not something desirable.

pkg/driver/node.go
lmouhib: This is not part of my change, it should be kept as is not merged with the other CSIDriver parameters.

pkg/driver/node.go
lmouhib: The tokens are returned as an array of json string, not sure the JWT library would help here.

cmd/main.go
lmouhib: We can probably move it as part of the driver initialization include it in the `Cloud` [object initialization](https://github.com/lmouhib/aws-efs-csi-driver/blob/8b2895563b8b5102c9e3641524fb102d41402e18/pkg/driver/driver.go#L56)?

deploy/kubernetes/base/node-daemonset.yaml
```diff
@@ -46,7 +46,7 @@ spec:
           args:
             - --endpoint=$(CSI_ENDPOINT)
             - --logtostderr
-            - --v=2
+            - --v=5
```
Ashley-wenyizha, line 49: we reduced default log level to 2 last year as we think this is bit too high. 

These messages are set at level 4/5 which is meant for debugging and does not need to be set to 5

pkg/driver/node.go
lmouhib: done

cmd/main.go
lmouhib: Moved the init of the client to driver.go in NewDriver call

deploy/kubernetes/base/node-daemonset.yaml
Ashley-wenyizha: Can we change this back to 2? Also can we squash  and only keep the crucial commits?

examples/kubernetes/pod_iam_impersonation/README.md
```diff
@@ -0,0 +1,3 @@
+## IRSA POD IAM IMPERSONATION
+
+TBC
```
Ashley-wenyizha, line 3: Do we have plan to fill this?

lmouhib: Yes, this will serve as example/documentation of the feature. It will be ready in the next couple of days.

pkg/driver/node.go
```diff
@@ -425,6 +491,83 @@ func parseVolumeId(volumeId string) (fsid, subpath, apid string, err error) {
...
+
+	if err != nil {
+		return err
+	}
+
+	err = setRoleArn(p)
+
+	if err != nil {
+		return err
+	}
```
Ashley-wenyizha, line 506: Same as above

pkg/driver/node.go
```diff
@@ -425,6 +491,83 @@ func parseVolumeId(volumeId string) (fsid, subpath, apid string, err error) {
 	return
 }
 
+func (p podAuth) setPodCredentials() (e error) {
+
+	err := writeToken(p)
+
+	if err != nil {
+		return err
```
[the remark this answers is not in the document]
Ashley-wenyizha: oh nvm you have it inside of writeToken

examples/kubernetes/pod_iam_impersonation/README.md
lmouhib: Fixed

Ashley-wenyizha:
I have performed test locally on my side, with service account associated role have permission, mount successes, otherwise fails

charts/aws-efs-csi-driver/templates/csidriver.yaml
```diff
@@ -8,3 +8,10 @@ metadata:
     "helm.sh/resource-policy": keep
 spec:
   attachRequired: false
+  {{- if .Values.iamAuth }}
```
wongma7, line 11: maybe something  a bit more descriptive for the helm value / volumeAttribute? `podIAMAuth`? `podIAMAuthorization? since this is b asically "mounting using the IAM authorization using the pod's IAM credentials" https://docs.aws.amazon.com/efs/latest/ug/mounting-IAM-option.html

examples/kubernetes/pod_iam_impersonation/README.md
```diff
@@ -0,0 +1,83 @@
+## CSI Driver POD IMPERSONATION
+
+This example shows how you can use the pod impersonation feature available 
+in the CSI Driver to enforce access control when mounting EFS Access point protected with EFS resource policy.
+This feature leverages IAM roles for service accounts (IRSA) and your EKS cluster must be enabled for IRSA.
```
wongma7, line 5: it should be stated explicitly that this feature is exclusive to EKS. right now it's implied because you need IRSA and only EKS has IRSA

wongma7:
lgtm I just have one nitpick about the naming of the attribute which will be impossible to  change later a it's basically an API ....

charts/aws-efs-csi-driver/templates/csidriver.yaml
lmouhib: FIxed, used `podIAMAuthorization` as it carry more meaning.

examples/kubernetes/pod_iam_impersonation/README.md
lmouhib: Fixed, specified that it only work with EKS.

examples/kubernetes/pod_iam_impersonation/README.md
```diff
@@ -0,0 +1,83 @@
...
+  volumeMode: Filesystem
+  accessModes:
+    - ReadWriteMany
+  persistentVolumeReclaimPolicy: Retain
+  storageClassName: efs-sc
+  csi:
+    driver: efs.csi.aws.com
+    volumeHandle: [FileSystemId]::[AccessPointId]
+    volumeAttributes:
+      iamAuth: "true"
```
Ashley-wenyizha, line 38: can you revise here as well?

examples/kubernetes/pod_iam_impersonation/README.md
```diff
@@ -0,0 +1,83 @@
...
+spec:
+  accessModes:
+    - ReadWriteMany
+  storageClassName: efs-sc
+  resources:
+    requests:
+      storage: 5Gi
+```
+
+Here it is important to set the property `iamAuth` in `volumeAttribute` to `true`.
```
Ashley-wenyizha, line 53: here as well

pkg/driver/node.go
Ashley-wenyizha: nit - suggest to change the var names here and later as well

examples/kubernetes/pod_iam_impersonation/README.md
lmouhib: fixed

examples/kubernetes/pod_iam_impersonation/README.md
lmouhib: fixed

Ashley-wenyizha:
/approve

Adding approve tag for now and let Matthew drop the `lgtm` tag

wongma7 (approved):

jvdadda:
Nice idea to put a breaking change inside a minor release, nice work guys

andrecastro:
Why was this never released?

PR #712, opened 2022-06-16
Updating documentation and adding storage class

### LINKED ISSUE #639: Add Storage Class to static_provisioning and statefulset examples

The static_provisioning and statefulset examples do not have a storageclass in the example.  This may result in confusion for people when looking at these examples when all others have storageclasses.

/feature

Add storageclass objects to the static_provisioning and statefulset examples

If there is a reason not to include these, please describe it in the readme.

**Additional context**

jonathanrainer:
Just a small documentation fix to add storage classes to two of the example as otherwise they will bind to the default StorageClass in EKS and fail to run.

See above, we want our examples to be easy to follow and clear.

**What testing is done?** 
Have run through the examples with a new cluster (1.22) and they all work fine now whereas without the storageClass they were a problem

fixes #639

### DIFF
--- a/examples/kubernetes/statefulset/specs/example.yaml
+++ b/examples/kubernetes/statefulset/specs/example.yaml
@@ -1,4 +1,10 @@
 apiVersion: v1
+kind: StorageClass
+metadata:
+  name: efs-sc
+provisioner: efs.csi.aws.com
+---
+apiVersion: v1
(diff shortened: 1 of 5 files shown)


jonathanrainer:
Hi @Ashley-wenyizha could we try and get this merged as it's just a docs change?

PR #714, opened 2022-06-19
Support ReadWriteOncePod

### LINKED ISSUE #540: Support ReadWriteOncePod (kube v1.22+)
The new ReadWriteOncePod access mode will be added to PVC, not sure what needs to change if much at all. 

[KEP-2485: ReadWriteOncePod PersistentVolume AccessMode](https://github.com/kubernetes/enhancements/tree/master/keps/sig-storage/2485-read-write-once-pod-pv-access-mode#csi-specification-changes-volume-capabilities
)

https://github.com/kubernetes/enhancements/issues/2485 

/feature

The driver to support ReadWriteOncePod. 

Not using ReadWriteOncePod. 

**Additional context**

jonathanrainer:
New Feature, now that ReadWriteOncePod has entered Alpha testing we should add support for it in the driver.

This PR enables us to serve users that want to use ReadWriteOncePod in their manifests. It adds new E2E tests to cover the use cases the ReadWriteOncePod covers.

**What testing is done?** 
More E2E testing in Kops only, as EKS has no way to set the feature gates required. The first draft of this may not be exactly what's needed from E2E tests but I need to use the CI framework to run them and then debug.

fixes #540 

@chrishenzie tagging for visibility and assistance.

### DIFF
--- a/test/e2e/e2e.go
+++ b/test/e2e/e2e.go
@@ -20,6 +20,7 @@ import (
 	clientset "k8s.io/client-go/kubernetes"
 	"k8s.io/kubernetes/test/e2e/framework"
 	e2epod "k8s.io/kubernetes/test/e2e/framework/pod"
+	e2epv "k8s.io/kubernetes/test/e2e/framework/pv"
 	storageframework "k8s.io/kubernetes/test/e2e/storage/framework"
 	"k8s.io/kubernetes/test/e2e/storage/testsuites"
 	"k8s.io/kubernetes/test/e2e/storage/utils"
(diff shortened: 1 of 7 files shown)


test/e2e/e2e.go
```diff
@@ -372,11 +373,60 @@ var _ = ginkgo.Describe("[efs-csi] EFS CSI", func() {
...
+			immediateBinding := storagev1.VolumeBindingImmediate
+			sc := storageframework.GetStorageClass("efs.csi.aws.com", map[string]string{
+				"provisioningMode": "efs-ap",
+				"fileSystemId":     FileSystemId,
+				"directoryPerms":   "777",
+				"gidRangeStart":    "1000",
+				"gidRangeEnd":      "2000",
+				"basePath":         "/dynamic_provisioning",
+			}, &immediateBinding, EfsDriverNamespace)
+			_, err := f.ClientSet.StorageV1().StorageClasses().Create(context.TODO(), sc, metav1.CreateOptions{})
```
chrishenzie, line 415: Do we need any automatic cleanup logic for this `StorageClass` after the test completes?

chrishenzie (approved):
This looks great, thanks for adding! Your work on the E2E test here may be a good candidate for porting into the [K8s E2E test](https://github.com/kubernetes/kubernetes/tree/master/test/e2e) library.

test/e2e/e2e.go
jonathanrainer: That sounds like a sensible idea, once I get someone on the repo to enable testing I can start debugging the tests as I'm not 100% sure they're right at the minute. Thank you for looking though

jonathanrainer:
@Ashley-wenyizha would I able to get this approved for testing so I can start getting the E2E tests sorted out?

Ashley-wenyizha:
> @Ashley-wenyizha would I able to get this approved for testing so I can start getting the E2E tests sorted out?

sure, thanks for working on this, tag added

test/e2e/e2e.go
```diff
@@ -373,11 +374,74 @@ var _ = ginkgo.Describe("[efs-csi] EFS CSI", func() {
...
+			podSpec.Spec.RestartPolicy = v1.RestartPolicyNever
+			pod1, err := f.ClientSet.CoreV1().Pods(f.Namespace.Name).Create(context.TODO(), podSpec, metav1.CreateOptions{})
+			framework.ExpectNoError(err, "creating pod")
+
+			err = e2epod.WaitForPodRunningInNamespace(f.ClientSet, pod1)
+			framework.ExpectNoError(err, "pod started running successfully")
+
+			pod2, err := f.ClientSet.CoreV1().Pods(f.Namespace.Name).Create(context.TODO(), podSpec, metav1.CreateOptions{})
+			err = e2epod.WaitForPodNotPending(f.ClientSet, f.Namespace.Name, pod2.Name)
+			framework.ExpectError(err, "second pod could not mount PVC as well")
```
chrishenzie, line 394: `pod2` in this case will fail during scheduling because we see `pod1` is already using the PVC. If you want to test enforcement on a single node you can try setting `Pod.Spec.NodeName`.

jonathanrainer: Hi Chris, thanks for the comments. Isn't this what we want to test? Or are you saying we should be more specific and allow the pod to be allocated to a node to then make sure it fails because of the PVC and not other conditions?

chrishenzie: I'm saying there's two ways ReadWriteOncePod can be enforced, 1) is during scheduling, and 2) is during mounting. The only way to test 2) is to force two pods on the same node by setting the node name.

A separate heads up, I'm working on implementing a similar e2e test in upstream k8s (I have to add it now so we can collect signal for 1.26 before beta graduation in 1.27). I'll tag you on the PR when it's available.

jonathanrainer: Oh of course, I'll see if I can cook something up that does that, it feels like an important test. If you want to just cherry pick what's here feel free

chrishenzie:
Hi! Just a heads up, an e2e test for ReadWriteOncePod was added here:
https://github.com/kubernetes/kubernetes/pull/112987

I think in order to make this work we'll need to:

- Update our k8s testsuite dependency (to pull in the capability + test suite), v1.26 should work
- Update `InitEFSCSIDriver()` to also advertise `storageframework.CapReadWriteOncePod`
- Add [`testsuites.InitReadWriteOncePodTestSuite()`](https://github.com/kubernetes/kubernetes/blob/v1.26.0/test/e2e/storage/testsuites/readwriteoncepod.go#L66) to `csiTestSuites`
- Verify the e2e test uses at least these sidecar versions:
  - csi-provisioner:v3.0.0+
  - csi-attacher:v3.3.0+
  - csi-resizer:v1.3.0+

And I think these tests will be pulled in for-free. Note that for this to succeed the ReadWriteOncePod feature will need to be enabled. In v1.27 this feature will be moved to beta and will be enabled by default.

jonathanrainer:
Hi @chrishenzie,

Sorry it's taken a long time for me to reply, I've had a look at this but I think that updating the Kubernetes version throughout the driver, even if it's just to access new E2E tests is something that should be done in a separate PR, not this one as it's likely to introduce breaking changes etc. Adding the test is definitely something we could do in the future though, might be best to raise a separate issue for it so when we do upgrade to later Kubernetes versions we can sort it then :)

chrishenzie:
ReadWriteOncePod is now graduated to beta in K8s v.127 and enabled by default. See [this comment on the Github issue](https://github.com/kubernetes-sigs/aws-efs-csi-driver/issues/540#issuecomment-1536468510) for more details.

The only thing required of CSI driver users is to [update their sidecars](https://kubernetes.io/blog/2023/04/20/read-write-once-pod-access-mode-beta/#how-can-i-start-using-readwriteoncepod).

There is also an [E2E test suite for the ReadWriteOncePod feature](https://github.com/kubernetes-sigs/aws-efs-csi-driver/pull/714#issuecomment-1352373687). See the steps in the above comment for what's required to pull in these tests and it should start working with the AWS EFS CSI driver. Please let me know if you encounter any issues and I can help.

Ashley-wenyizha:
Hi Jonathan, this PR lgtm, could you give a fix of the conflicts? I will merge this in

thank you!

jonathanrainer:
@Ashley-wenyizha I've fixed this up now, had some weird errors with the new E2E tests but I think that was a result of a bad merge conflict/adding things I didn't need because we're now using 1.27 in our E2E tests.

chrishenzie:
Update -- [`ReadWriteOncePod` is now GA/Stable](https://kubernetes.io/blog/2023/12/18/read-write-once-pod-access-mode-ga/) as of the K8s v1.29. So long as this driver is deployed with [these minimum CSI sidecar versions](https://kubernetes.io/blog/2023/12/18/read-write-once-pod-access-mode-ga/), `ReadWriteOncePod` will work without any driver changes here.

Additionally, the K8s E2E test suite includes tests for this access mode that should work for this driver. In order to pull in and use these tests you will need to first update your K8s dependencies to at least v1.27.

An important thing this PR is tackling is the migration to the new CSI access modes. Since support for the K8s access mode and E2E testing is handled outside of this repo, I would consider refocusing this PR on the CSI access mode migration.

Issue #300, opened 2021-01-11
chown/chgrp on dynamic provisioned pvc fails


gazal-k:
/kind bug

**What happened?**

Using dynamic provisioning feature introduced by #274 with applications that try to `chown` their pv fails.

**What you expected to happen?**

With the old [efs-provisioner](https://github.com/kubernetes-retired/external-storage/tree/master/aws/efs), this caused no issues.
But with dynamic provisioning in this csi driver, the chown command fails. I must admit, I don't understand how the uid/gid thing works with EFS access points. The pod user does not seem to have any association with the uid/gid on the access point, however pods can read & write mounted pv just fine.

**How to reproduce it (as minimally and precisely as possible)?**

- Use the dynamic provisioning feature introuduced by #274
- Create a storageClasss:
  ```yaml
  apiVersion: storage.k8s.io/v1
  kind: StorageClass
  metadata:
    name: efs-sc
  mountOptions:
  - tls
  parameters:
    directoryPerms: "700"
    fileSystemId: fs-<id>
    provisioningMode: efs-ap
  provisioner: efs.csi.aws.com
  ```
- Use it with application charts that have a chown step in the beginning, like:
  - https://github.com/grafana/helm-charts/blob/8dfa6da2790911ee78e2c9cf62f950f20e4a8129/charts/grafana/templates/_pod.tpl#L22-L40
  - https://github.com/jenkins-x-charts/nexus/blob/f05c2172a550c6bf3daf2c25eef940e67290d346/Dockerfile#L9-L10

For the first chart, we saw the initContainer failing. We just tried disabling the initContainer on the grafana chart with these chart overrides:
```yaml
initChownData:
  enabled: false
```
And the application worked fine.

For the nexus chart, there's a whole bunch of errors from the logs,

```sh
chgrp: changing group of '/nexus-data/elasticsearch/nexus': Operation not permitted
```

Perhaps these charts don't really need that step with dynamic provisioning. Another nexus chart seems to have an option to skip the `chown` step; https://github.com/travelaudience/docker-nexus/blob/a86261e35734ae514c0236e8f371402e2ea0feec/run#L3-L6

**Anything else we need to know?**:

**Environment**
- Kubernetes version (use `kubectl version`):
  ```sh
[...]

Issue #125, opened 2020-01-28
securityContext doesn't applied when creating pvc


Amos-85:
/kind bug

**What happened?**
Deploy Hashicorp [vault helm chart ](https://github.com/hashicorp/vault-helm) that include this `securityContext` below without the specified uid/gid ( the /vault/data folder created by root ):
```
securityContext:
        runAsNonRoot: true
        runAsGroup: {{ .Values.server.gid | default 1000 }}
        runAsUser: {{ .Values.server.uid | default 100 }}
        fsGroup: {{ .Values.server.gid | default 1000 }}
```

**What you expected to happen?**
mount efs persistent volume with the right `securityContext` attributes
**How to reproduce it (as minimally and precisely as possible)?**
deploy vault helm chart with declaring the right storageClass that represents efs.

**Environment**
- `Kubernetes version: v1.15.3`
- `Driver version: 0.2.0`

leakingtapan:
Thx for reporting the issue. What uid/gid of the container do you see when "securityContext doesn't applied when creating pvc"? And I'm assuming you are setting the security container to the csi driver? Could you clarify with more info like the pod spec and logs

Amos-85:
Thanks for response @leakingtapan ,
the uid/gid of the container is:
```securityContext:
    fsGroup: 1000
    runAsGroup: 1000
    runAsNonRoot: true
    runAsUser: 100
```
pod spec:

```apiVersion: v1
kind: Pod
metadata:
  annotations:
    cni.projectcalico.org/podIP: 10.42.133.210/32
  creationTimestamp: "2020-02-10T14:09:40Z"
  generateName: vault-
  labels:
    app.kubernetes.io/instance: vault
    app.kubernetes.io/name: vault
    component: server
    controller-revision-hash: vault-7f64cccd79
    helm.sh/chart: vault-0.3.3
    statefulset.kubernetes.io/pod-name: vault-0
  name: vault-0
  namespace: vault
  ownerReferences:
  - apiVersion: apps/v1
    blockOwnerDeletion: true
    controller: true
    kind: StatefulSet
    name: vault
    uid: 77540e58-ec01-4129-a98a-4f0f05b45da3
  resourceVersion: "3692437"
  selfLink: /api/v1/namespaces/vault/pods/vault-0
  uid: 7325dc66-77fc-47d7-bafb-d1e061a2da35
spec:
  affinity:
    podAntiAffinity:
[...]

PR #727, opened 2022-06-23
Delegate`fsGroup` to driver rather than Kubelet

jonathanrainer:
New feature

The new feature of CSI documented here: https://kubernetes-csi.github.io/docs/support-fsgroup.html#delegate-fsgroup-to-csi-driver, allows us to get the fsGroup from the pod that wants to use the PVC specified. This will be particularly helpful to solve issues like #125 and #300 or at least give a slightly better route to solving them. This is a bit of a speculative feature at the minute as the feature in Kubernetes doesn't go Beta till 1.23, not supported by EKS till August but it would be good to have this on the slate for when 1.23 rolls out.

**What testing is done?** 
This has had new Unit Tests and E2E tests added and I plan to spin this up on my own cluster and will post results as that happens. The E2E might not be quite right yet but once this PR is approved for testing I'll iterate on them at that point.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -119,7 +119,7 @@ test-e2e:
 	K8S_VERSION="1.22.9" \
 	DRIVER_NAME=aws-efs-csi-driver \
 	CONTAINER_NAME=efs-plugin \
-	TEST_EXTRA_FLAGS='--cluster-name=$$CLUSTER_NAME' \
+	TEST_EXTRA_FLAGS='--cluster-name=$$CLUSTER_NAME --created-by=kops' \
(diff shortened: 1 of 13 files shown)


jonathanrainer:
@Ashley-wenyizha Could I have an approval to test here too? Sorry for the spam!

jonathanrainer:
Have played with this and don't think that this solution will work because multiple pods could attempt to override the fsGroup and then we'd end up with a lot of concurrency issues. You could implement locks and that sort of thing but that seems overblown when we have the dynamic provisioning solution with `uid` and `gid` parameters

PR #731, opened 2022-06-28
Helm Chart: Add Pod Labels

iusergii:
**Is this a bug fix or adding a new feature?**
Helm feature

Ability to add labels to Pods.


TAMQuiroz:
We would like to use this feature to better track deployments using a monitoring tool via labels

PR #615, opened 2022-01-17
add updateStrategy

nalbam:
**Is this a bug fix or adding new feature?**

Feature

**What is this PR about? / Why do we need it?**

Added updateStrategy to controller and deamonset.
When deploying to a large number of nodes, the deployment speed can be faster.
Code referenced aws-ebs-csi-driver .

https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/charts/aws-ebs-csi-driver/templates/controller.yaml#L10

**What testing is done?**

PR #740, opened 2022-07-15
feat(node/controller): allow to set updateStrategy

lefterisALEX:
(New Feature)
This PR allow to define `updateStrategy` for `controller deployment` and `node daemonset`

With default update strategy takes too long to update all the pods of the daemonset if you have many worker nodes. This results in helm timeout (after 300 second.)

We would like to have the option to tune the `updateStrategy` if needed. Selecting different `updateStrategy` type or tuning the default`RollingUpdate`.

ps: I see there is an [open PR](https://github.com/kubernetes-sigs/aws-efs-csi-driver/pull/615) already about it, but since is a bit old ,not sure if someone is still working on it. If that can be merged i can close mine.

### DIFF
--- a/charts/aws-efs-csi-driver/templates/controller-deployment.yaml
+++ b/charts/aws-efs-csi-driver/templates/controller-deployment.yaml
@@ -13,6 +13,10 @@ spec:
       app: efs-csi-controller
       app.kubernetes.io/name: {{ include "aws-efs-csi-driver.name" . }}
       app.kubernetes.io/instance: {{ .Release.Name }}
+  {{- with .Values.controller.updateStrategy }}
+  strategy:
+    {{ toYaml . | nindent 4 }}
+  {{- end }}
   template:
     metadata:
       labels:
--- a/charts/aws-efs-csi-driver/templates/node-daemonset.yaml
+++ b/charts/aws-efs-csi-driver/templates/node-daemonset.yaml
@@ -11,6 +11,10 @@ spec:
       app: efs-csi-node
       app.kubernetes.io/name: {{ include "aws-efs-csi-driver.name" . }}
       app.kubernetes.io/instance: {{ .Release.Name }}
+  {{- with .Values.node.updateStrategy }}
+  updateStrategy:
+    {{ toYaml . | nindent 4 }}
+  {{- end }}
   template:
     metadata:
       labels:
--- a/charts/aws-efs-csi-driver/values.yaml
+++ b/charts/aws-efs-csi-driver/values.yaml
@@ -68,6 +68,7 @@ controller:
     #   cpu: 100m
     #   memory: 128Mi
   nodeSelector: {}
+  updateStrategy: {}
   tolerations: []
   affinity: {}
   # Specifies whether a service account should be created
@@ -112,6 +113,10 @@ node:
     #   cpu: 100m
     #   memory: 128Mi
   nodeSelector: {}
+  updateStrategy: {}
+    # Override default strategy (RollingUpdate) to speed up deployment.
+    # This can be useful if helm timeouts are observed.
+    # type: OnDelete
   tolerations:
     - operator: Exists
   # Specifies whether a service account should be created


pierluigilenoci:
@jqmichael @wongma7 any news about this?

pierluigilenoci:
@jonathanrainer please take a look

pierluigilenoci:
@Ashley-wenyizha @RomanBednar can you please take a look?

lefterisALEX:
@wongma7 @jqmichael any news about it?

pierluigilenoci:
@Ashley-wenyizha @RomanBednar @jonathanrainer @makomatic could you please take a look?

charts/aws-efs-csi-driver/values.yaml
```diff
@@ -66,6 +66,11 @@ controller:
     #   cpu: 100m
     #   memory: 128Mi
   nodeSelector: {}
+  updateStrategy: {}
+  # type: RollingUpdate
```
RomanBednar, line 70: Why would we add a comment with a strategy that is the [default](https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#strategy) in Kubernetes? We should probably add a comment with good explanation. Something like: "override default strategy (RollingUpdate) to speed up large deployments" - and below add a comment/example of the strategy that provides better result at scale? Same for the node strategy below.

charts/aws-efs-csi-driver/values.yaml
```diff
@@ -110,6 +115,11 @@ node:
     #   cpu: 100m
     #   memory: 128Mi
   nodeSelector: {}
+  updateStrategy: 
+   type: RollingUpdate
```
RomanBednar, line 119: We should not change the strategy for node deployment but only add the possibility to change it since this PR says exactly that. So this should be `updateStrategy: {}` and the rest commented out.

charts/aws-efs-csi-driver/values.yaml
```diff
@@ -66,6 +66,11 @@ controller:
     #   cpu: 100m
     #   memory: 128Mi
   nodeSelector: {}
+  updateStrategy: {}
+  # type: RollingUpdate
+  # rollingUpdate:
```
RomanBednar, line 71: What strategy did you use to speed up your deployments? Did you test both and compared the results? Could you maybe share it in the PR description please?

RomanBednar (changes requested):

pierluigilenoci:
@lefterisALEX could you please integrate/review @RomanBednar suggestions for the PR? And maybe rebase your branch.

pierluigilenoci:
@RomanBednar your reviews have been integrated, can you take a look again?

charts/aws-efs-csi-driver/values.yaml
lefterisALEX: @RomanBednar i added the comment you suggested. Not sure what to add as an example for large scale deployments , since that might depends of how many worker nodes you have and also what type of nodes (spot or on demand).

charts/aws-efs-csi-driver/values.yaml
lefterisALEX: done

charts/aws-efs-csi-driver/values.yaml
lefterisALEX: I haven't done any test to compare speed between different strategies..
With default strategy the rollout take too long and helm times out after 300 second.
 We would like to use OnDelete. We are using only spot instances and nodes are replaced frequently.
(will update the description as well)

charts/aws-efs-csi-driver/values.yaml
```diff
@@ -66,6 +66,8 @@ controller:
     #   cpu: 100m
     #   memory: 128Mi
   nodeSelector: {}
+  updateStrategy: {}
+    # override default strategy (RollingUpdate) to speed up large deployments"
```
RomanBednar, line 70:
```suggestion
```

charts/aws-efs-csi-driver/values.yaml
```diff
@@ -110,6 +112,7 @@ node:
     #   cpu: 100m
     #   memory: 128Mi
   nodeSelector: {}
+  updateStrategy: {}
```
RomanBednar, line 115:
```suggestion
  updateStrategy:
    {}
    # Override default strategy (RollingUpdate) to speed up deployment. 
    # This can be useful if helm timeouts are observed.
    # type: OnDelete
```

charts/aws-efs-csi-driver/values.yaml
RomanBednar: Now I understand it better, so you tested DaemonSet `OnDelete` strategy only to get past the helm timeout. And the problem does not necessarily come from having huge number of nodes but could be caused by instance type of worker nodes, as you observed.

Then I think leaving controller deployment without any comment is ok because at this point we don't know what alternative to suggest (and controller pods were not causing the issue anyway) and instead it should go under node values - see my suggestions.

charts/aws-efs-csi-driver/values.yaml
lefterisALEX: updated

lefterisALEX:
@RomanBednar can you please take a look again to the latest commit?

lefterisALEX:
> @lefterisALEX: PR needs rebase.

Done

pierluigilenoci:
@RomanBednar @wongma7 could you please take a look?

lefterisALEX:
@RomanBednar @jqmichael can you please take a look into that?

lefterisALEX:
@jqmichael can you please take a look?

charts/aws-efs-csi-driver/values.yaml
```diff
@@ -112,6 +113,11 @@ node:
     #   cpu: 100m
     #   memory: 128Mi
   nodeSelector: {}
+  updateStrategy:
+    {}
```
pierluigilenoci, line 117: Aside from this `{}` which might be put on the top line the only thing this PR is missing, IMHO, is the fact that it is missing a bump of the chart version.

pierluigilenoci (approved):

charts/aws-efs-csi-driver/values.yaml
lefterisALEX: good comments @pierluigilenoci , applied to new commit. 
thanks

lefterisALEX:
rebase  and squash commits. 
@RomanBednar @jqmichael can you please review?

pierluigilenoci (approved):

lefterisALEX:
rebased.
@RomanBednar @wongma7 @jqmichael

lefterisALEX:
@wongma7 @jqmichael i will really appreciate if you can  take a look into that ;)

lefterisALEX:
rebased.
/assign @wongma7

lefterisALEX:
rebased, @RomanBednar can you please take another look?

pierluigilenoci:
@lefterisALEX thank you!

pierluigilenoci (approved):

pierluigilenoci (approved):

cablespaghetti:
/assign mskanth972

As a new owner, fancy reviewing this long languishing PR?

lefterisALEX:
@cablespaghetti  i see github didn't allowed to assign this to @mskanth972 .Any suggestions?

charts/aws-efs-csi-driver/templates/controller-deployment.yaml
```diff
@@ -13,6 +13,10 @@ spec:
       app: efs-csi-controller
       app.kubernetes.io/name: {{ include "aws-efs-csi-driver.name" . }}
       app.kubernetes.io/instance: {{ .Release.Name }}
+  {{- with .Values.controller.updateStrategy }}
+  strategy:
+    {{ toYaml . | nindent 4 }}
```
Ashley-wenyizha, line 18: are we missing a dash `-` before `toYaml` here? same as below

charts/aws-efs-csi-driver/values.yaml
Ashley-wenyizha: here lgtm

charts/aws-efs-csi-driver/templates/controller-deployment.yaml
RyanStan: No, this looks good.  `-` kills white space but I think we want that here

RyanStan:
Testing rendering
```
updateStrategy:
    # Override default strategy (RollingUpdate) to speed up deployment. 
    # This can be useful if helm timeouts are observed.
    # type: OnDelete
    type: RollingUpdate
    rollingUpdate:
      maxUnavailable: 1
      maxSurge: 1
```

And I get
```
  strategy:
    
    rollingUpdate:
      maxSurge: 1
      maxUnavailable: 1
    type: RollingUpdate
```

Not sure why order is backwards and space exists.  I can't seem to get rid of the space no matter where I add dashes.  Can someone confirm, does spacing / order here matter?  I assume it doesn't, but I want to run a quick test to make sure.

dawid-remitly:
> Not sure why order is backwards

I believe it's because it's sorted lexically.
Guys, Is there anything else this PR is waiting? I miss this feature so hard ... and it seems to be a simple change...

lefterisALEX:
also not sure why order is not kept, but since indentation do not change is that a blocking issue?

pierluigilenoci:
@mskanth972 @Ashley-wenyizha @markapruett could someone of you help this PR to reach the finish line?

charts/aws-efs-csi-driver/templates/node-daemonset.yaml
```diff
@@ -11,6 +11,10 @@ spec:
       app: efs-csi-node
       app.kubernetes.io/name: {{ include "aws-efs-csi-driver.name" . }}
       app.kubernetes.io/instance: {{ .Release.Name }}
+  {{- with .Values.node.updateStrategy }}
+  strategy:
```
617m4rc, line 15: This is incorrect. The field name for DaemonSets is `updateStrategy`, not `strategy`

lefterisALEX: good catch, indeed `strategy` is for Deployments and for daemonset should be `updateStrategy`. will fix it soon

lefterisALEX: @617m4rc  fixed, can you please review and resolve if you agree?

lefterisALEX:
pushed a new commit fix the issue @617m4rc spotted. I rebased as well.

 https://github.com/kubernetes-sigs/aws-efs-csi-driver/pull/740#discussion_r1114011753

Issue #741, opened 2022-07-19
Remove deprecated NodeSelector


dschunack:
Hi,

beta.kubernetes.io/os is deprecated since v1.14 and it make no sense to set this in the template.
It's better and also best practice to set this in the values.yaml.

The new NodeSelector is kubernetes.io/os

https://github.com/kubernetes-sigs/aws-efs-csi-driver/blob/c3eaccfc13ff175db8bcdea6da00c3d0b0732c2f/charts/aws-efs-csi-driver/templates/node-daemonset.yaml#L39

dschunack:
PR created

PR #743, opened 2022-07-26
Update deprecated NodeSelector

dschunack:
Signed-off-by: dschunack <dschunack@web.de>

No, it replaced the deprecated NodeSelector labels #741 

It replaced the deprecated NodeSelector labels #741 

**What testing is done?** 
helm chart was applied on a test EKS

### DIFF
--- a/charts/aws-efs-csi-driver/Chart.yaml
+++ b/charts/aws-efs-csi-driver/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 name: aws-efs-csi-driver
 version: 2.2.9
-appVersion: 1.4.2
+appVersion: 1.4.3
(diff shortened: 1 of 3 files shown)


dschunack:
Hey @Ashley-wenyizha @justinsb, this is only a small PR to replace deprecated NodeSelectors. Would be nice, if you Approve the PR.

charts/aws-efs-csi-driver/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v2
 name: aws-efs-csi-driver
 version: 2.2.9
-appVersion: 1.4.2
```
Ashley-wenyizha, line 4: why are we changing here before a release?

PR #752, opened 2022-08-08
Update uid/gid Readme

Ashley-wenyizha:
**Is this a bug fix or adding new feature?**

Doc update - Document existing access point user identity enforcement behaviors in README.

https://github.com/kubernetes-sigs/aws-efs-csi-driver/pull/748

^^ As Daniel is blocked by the EasyCLA issue, re-submitting this PR through my account.

PR #755, opened 2022-08-11
Should not pass in mount option of awscredsuri

Ashley-wenyizha:

Should not pass in mount option of `awscredsuri` as there is no EKS customer should be using it, the uri is used together with ECS metadata endpoint. We should ignore that option if customer passed it in efs-csi-driver.

**What testing is done?**

### DIFF
--- a/examples/kubernetes/access_points/README.md
+++ b/examples/kubernetes/access_points/README.md
@@ -98,3 +98,5 @@ as this could subject you to
     `/ap1/my/subpath`.
   - As with normal volume path, the `[Subpath]` must already exist prior to consuming
     the volume from a pod.
+
+- `awscredsuri` mount option is not supported through efs-csi-driver as it's designed and used by ECS tasks.
--- a/pkg/driver/node.go
+++ b/pkg/driver/node.go
@@ -78,7 +78,7 @@ func (d *Driver) NodePublishVolume(ctx context.Context, req *csi.NodePublishVolu
 		switch strings.ToLower(k) {
 		//Deprecated
 		case "path":
-			klog.Warning("Use of path under volumeAttributes is depracated. This field will be removed in future release")
+			klog.Warning("Use of path under volumeAttributes is deprecated. This field will be removed in future release")
 			if !filepath.IsAbs(v) {
 				return nil, status.Errorf(codes.InvalidArgument, "Volume context property %q must be an absolute path", k)
 			}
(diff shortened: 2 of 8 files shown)


wangnyue (approved):

pkg/driver/node.go
```diff
@@ -165,6 +165,12 @@ func (d *Driver) NodePublishVolume(ctx context.Context, req *csi.NodePublishVolu
 				}
 			}
 
+			if f == "awscredsuri" {
+				klog.Warning("awscredsuri is not accepted as a efs-csi-driver mount option")
+				return nil, status.Errorf(codes.InvalidArgument,
+					"Should not pass in awscredsuri as a mount option")
```
yinsihaws, line 171: nit: align

yinsihaws (approved):

Ashley-wenyizha:
After running `hack/update-gofmt`,  pull-aws-efs-csi-driver-verify still fails, similar in https://github.com/kubernetes-sigs/aws-efs-csi-driver/pull/752

Ashley-wenyizha:
failed test was able to be fixed updating Go to latest version and run `/hack/update-gofmt`

pkg/driver/node.go
```diff
@@ -165,6 +165,12 @@ func (d *Driver) NodePublishVolume(ctx context.Context, req *csi.NodePublishVolu
 				}
 			}
 
+			if f == "awscredsuri" {
```
Cappuccinuo, line 168: 1. I would suggest to have a non supported options list, and check whether that option is in that list
2. If that option is not supported, I think we should fail gracefully by just ignore that option + warn log, instead of return an error

Cappuccinuo:
Added on above comment:

1. Can we add some unit test for this change? Basically validating when the option is passed, the csi-driver won't fail the mount but post a warn log, or that option is not passed to efs-utils at all.
2. I think we should ignore the netns option as well, but that could be in a separate PR.

pkg/driver/node.go
Ashley-wenyizha: thanks, logged this behavior, return nil but do not fail

Updated this behavior in ReadMe

examples/kubernetes/access_points/README.md
```diff
@@ -97,4 +97,6 @@ as this could subject you to
     if you configured your access point with `/ap1`, the above would mount to
     `/ap1/my/subpath`.
   - As with normal volume path, the `[Subpath]` must already exist prior to consuming
-    the volume from a pod.
\ No newline at end of file
+    the volume from a pod.
+
+- `awscredsuri` mount option is not supported through efs-csi-driver due to security concerns.
```
Cappuccinuo, line 102: Not due to security concerns, it is due to this option is designed and used by ECS tasks.

pkg/driver/node.go
```diff
@@ -165,6 +165,11 @@ func (d *Driver) NodePublishVolume(ctx context.Context, req *csi.NodePublishVolu
...
+				klog.Warning("awscredsuri is not accepted as a efs-csi-driver mount option")
```
Cappuccinuo, line 169: awscredsuri mount option is not supported by efs-csi-driver.

PR #761, opened 2022-08-25
Support multiple EFS filesystems

artificial-aidan:
Signed-off-by: Aidan Jensen <aidan@artificial.com>

New feature to allow multiple efs file systems to increase the number of supported PVCs

Currently there is a limit where only 120 pvcs are supported because each PVC is mapped to an AccessPoint

**What testing is done?** 

Running this at our company for some time. Also added unit testing.


Ashley-wenyizha:
Hi @artificial-aidan 

EFS now support 1000 access points and able to support 1000 PVCs now, I will close this PR as we do not plan to support beyond this volume at the moment. Thank you for your work here!

PR #769, opened 2022-09-23
allow specifying timeout duration

mathbruyen:
On an EKS cluster we experienced some node pods entering crashloop while volumes were actually correctly attached to nodes and mounted on pods. This was due to the liveness probe taking too long (its default timeout is 1s), failing with a timeout and not managing to recover. The probe continued to fail even after the container is restarted.

AWS support (case 10734780791) suggested increasing timeouts which was performed by manual updates on the daemonset and pods are running fine since then. But those timeouts could not be defined in the Helm chart.

Add timeout values in the Helm chart with default values set to their current default.

Feature

CSI pods randomly entering crashloop on some clusters

**What testing is done?** 

Change applied on one of our cluster


mathbruyen:
@leakingtapan any opinion on that?

PR #777, opened 2022-09-29
Revert "EFS-CSI pod impersonation implementation"

wongma7:
Reverts kubernetes-sigs/aws-efs-csi-driver#710


adammw:
Is there any more information on the reason for the revert?

andrecastro:
Why this was reverted?

dims:
@adammw @andrecastro can you please share some context on why the sudden interest in a revert after 2-ish years?

andrecastro:
@dims I manage a multi-tenant platform based on k8s, meaning I have applications of different teams sharing the same cluster. I would like to whitelist the access to the applications using the EFS resource policy, restricting the access only to the roles associated to the pods (IRSA) that need access.

Right now I'm doing it via templates and naming patterns that we have in our platform, but I think this feature would be a valid option thinking about security concerns.

The [secrets-store-csi-driver-provider-aws](https://github.com/aws/secrets-store-csi-driver-provider-aws) have something similar. I can create secrets on AWS Secrets Manager and restrict the access to the pods. The provider kind of impersonate the pod mounting the secret, using its permissions.

dims:
@andrecastro thanks for sharing! please open a new issue and point it back to the discussion here (and cut-n-paste the info here). it's just that folks don't really pay attention to old closed PRs and better visibility as a fresh request.

adammw:
#1439 was opened back in August

PR #796, opened 2022-10-20
add daemonset affinity control

camaeel:
New feature

Add possiblity to control daemonset's affinity. 
In my setup I don't need EFS support on control-plane nodes so I would like to modify affinity. 

**What testing is done?** 
I run `helm template` of the helm chart. It was identical with original one.


camaeel:
@nckturner @Ashley-wenyizha how many times am I supposed to rebase this PR before it gets merged?

PR #797, opened 2022-10-21
Add envFrom efs-plugin to container spec in controller deployment

jebbens:
New feature
Adding controller.envFrom to values files so that users can easily specify a ConfigMap (for things like proxy environment variables) or Secrets. Only added envFrom to the efs-plugin container spec in the controller-deployment, but it could be easily added to other container specs. 
**What testing is done?** 
Have tested use of ConfigMap with EFS CSI driver using manifests and confirmed it works. Have tested this same change to the EBS CSI driver charts and confirmed it works.


jebbens:
Pushed a new commit with original changes to correct mix-up after rebase

PR #813, opened 2022-11-04
EFS-54767 fixed: CVE-2021-25740

mskanth972:

**What testing is done?**


Dockerfile
```diff
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM golang:1.17 as builder
+FROM golang:1.19.2 as builder
```
Ashley-wenyizha, line 15: Remember to update `Kubernetes Version Compability Matrix` in Readme as well

PR #840, opened 2022-11-29
Bumps sidecars to 1.23

markapruett:
Bumps sidecars to a more recent version

I'm bumping the sidecar images to version 1.23 to get the conversation started on updating the images used. 1.18 has been out of support for some time. I noticed 1.18 was almost out of support when it was updated in [this PR](https://github.com/kubernetes-sigs/aws-efs-csi-driver/pull/606) so you may not want to go all the way to 1.23. 

We could also go up to 1.24, but the best would be to use the correct version based on the k8s version being used. Adding  switching logic based on version in helm is easy enough, but first we'd need to rethink the approach of building a single version for all clusters, and move to building different versions for each version of K8s supported by EKS. Short of that, this PR addresses several security and support issues from running old versions. 

**What testing is done?** 
No testing yet

### DIFF
--- a/charts/aws-efs-csi-driver/values.yaml
+++ b/charts/aws-efs-csi-driver/values.yaml
@@ -18,7 +18,7 @@ sidecars:
   livenessProbe:
     image:
       repository: public.ecr.aws/eks-distro/kubernetes-csi/livenessprobe
-      tag: v2.8.0-eks-1-23-8 
+      tag: v2.8.0-eks-1-23-8
       pullPolicy: IfNotPresent
     resources: {}
   nodeDriverRegistrar:
--- a/charts/aws-efs-csi-driver/values.yaml
+++ b/charts/aws-efs-csi-driver/values.yaml
@@ -18,19 +18,19 @@ sidecars:
   livenessProbe:
     image:
       repository: public.ecr.aws/eks-distro/kubernetes-csi/livenessprobe
-      tag: v2.2.0-eks-1-18-13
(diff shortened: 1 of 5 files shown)


scottsidwell:
+1 - this would help the >1 old issue where the defaults for this project mean that livenessProbe emits a lot of logging noise for consumers (https://github.com/kubernetes-sigs/aws-efs-csi-driver/issues/564) & the recently raised security concerns https://github.com/kubernetes-sigs/aws-efs-csi-driver/issues/846.

@Ashley-wenyizha  / @justinsb any chance we could get an estimate on when you might have time to get around to this? (or at least, an okay to trigger CI to move this forward more)

npellegrin:
Hello :)

This change was integrated to the `master` branch but is not in the recent updates of the `release-1.4` branch. Will it be deployed in a future `release-1.5` branch ?

See branches diff with the kustomization.yaml file:

* https://github.com/kubernetes-sigs/aws-efs-csi-driver/blob/master/deploy/kubernetes/overlays/stable/kustomization.yaml
* https://github.com/kubernetes-sigs/aws-efs-csi-driver/blob/release-1.4/deploy/kubernetes/overlays/stable/kustomization.yaml

Ashley-wenyizha:
Hi [npellegrin](https://github.com/npellegrin)@

Yes this is not yet released on driver yet, we will release this in version `v1.4.9`, will update here once we have a concrete timeline of next release

PR #885, opened 2023-01-21
feat: Pod impersonation

lmouhib:
New feature

Add the ability for efs-csi-driver to impersonate pods.

Recreating #710 the previous one was closed and reverted after merge. Starting from new one instead of reverting.

What testing is done?
Tested with pods that have service account annotated with a role ARN. The efs-utils assume the role annotated to the service account and is able to mount an access point that is protected with an IAM resource policy.


mhulscher:
Is there any progress on this? Would love to have this feature.

PR #910, opened 2023-02-06
Remove hostNetwork option from controller

mskanth972:

**What testing is done?**


diranged:
@mskanth972 Why was this change implemented? Or at the very least, why was it made as a change that we cannot override?

PR #937, opened 2023-02-24
Helm charts prefix image names with containerRegistry

RyanStan:
Added .values.containerRegistry to the values.yaml file, and prefixed all our images with the container registry.  Also modifed the csi driver image to pull from the aws public ecr instead of dockerhub.

This will require updating the helm installation documentation here as well: https://docs.aws.amazon.com/eks/latest/userguide/efs-csi.html. 

New feature

We need to be able to prefix all the images (sidecars + csi driver image )when using private ECR registries.

**What testing is done?** 
Deployed the helm chart without configuring any of the default values in values.yaml and made sure images pulled successfully.

### DIFF
--- a/charts/aws-efs-csi-driver/templates/controller-deployment.yaml
+++ b/charts/aws-efs-csi-driver/templates/controller-deployment.yaml
@@ -47,7 +47,7 @@ spec:
         - name: efs-plugin
           securityContext:
             privileged: true
-          image: {{ printf "%s:%s" .Values.image.repository (default (printf "v%s" .Chart.AppVersion) (toString .Values.image.tag)) }}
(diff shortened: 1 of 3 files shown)


RyanStan:
Oh. Figured out why the tests were failing.
```
/home/prow/go/src/github.com/kubernetes-sigs/aws-efs-csi-driver/hack/e2e/csi-test-artifacts/bin/helm upgrade --install aws-efs-csi-driver --namespace kube-system --set image.repository=607362164682.dkr.ecr.us-west-2.amazonaws.com/aws-efs-csi-driver --set image.tag=22696 --wait --kubeconfig /home/prow/go/src/github.com/kubernetes-sigs/aws-efs-csi-driver/hack/e2e/csi-test-artifacts/test-cluster-22696.k8s.local.kops.kubeconfig ./charts/aws-efs-csi-driver -f ./hack/values.yaml
Release "aws-efs-csi-driver" does not exist. Installing it now. 
```

The value passed in to image will break how this works now. This is a good call out that we'll have to modify our public documentation along with this PR when it's merged

PR #947, opened 2023-02-28
Updated the k8s and go version

mskanth972:
Updated the k8s and go version

**What testing is done?**


go.mod
```diff
@@ -1,129 +1,138 @@
...
-	k8s.io/apimachinery v0.24.3
+	github.com/onsi/gomega v1.23.0
+	google.golang.org/grpc v1.49.0
+	k8s.io/api v0.26.1
+	k8s.io/apimachinery v0.26.1
 	k8s.io/client-go v1.5.2
 	k8s.io/klog v1.0.0
-	k8s.io/kubernetes v1.22.3
-	k8s.io/mount-utils v0.22.3
+	k8s.io/kubernetes v1.26.1
```
RyanStan, line 16: Latest supported EKS version is 1.25, so I think we should stick to 1.25

mskanth972: Yes, checked in the first part but forgot to close this PR.

PR #958, opened 2023-03-08
Botocore test case

mskanth972:

This PR tests whether the botocore is installed properly in driver

**What testing is done?**

### DIFF
--- a/test/e2e/e2e.go
+++ b/test/e2e/e2e.go
@@ -188,6 +188,7 @@ var _ = ginkgo.SynchronizedBeforeSuite(func() []byte {
 			destroyDriver = true
 		}
 	}
+
 	return []byte(FileSystemId)
 }, func(data []byte) {
 	// allNodesBody: each node needs to set its FileSystemId as returned by node 1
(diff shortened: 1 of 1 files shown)


test/e2e/e2e.go
```diff
@@ -229,6 +230,18 @@ var _ = ginkgo.Describe("[efs-csi] EFS CSI", func() {
 
 	f := framework.NewDefaultFramework("efs")
 
+	ginkgo.It("Validating driver installation", func() {
```
RyanStan, line 233: If there's not a specific reason to have this in a separate Gingko Context than the other tests, than I think we should move this under the existing `ginkgo.Context(storageframework.GetDriverNameWithFeatureTags(driver), func()`

Then, we would have something like 
`ginkgo.It("Validating botocore installation")`

Also, you might want to double check the alignment, I think this block should be indented one more time

RyanStan (changes requested):

test/e2e/e2e.go
mskanth972: `ginkgo.It("Validating driver installation")` is kept as generic. So that many packages can be validated in future.

PR #981, opened 2023-04-04
Add ability to configure daemonset affinity

smcavallo:
Signed-off-by: smcavallo <smcavallo@hotmail.com>

n/a

End users may not want the daemonset running on all nodes.  This PR exposes the daemonset affinity so it can be configurable via helm.

**What testing is done?** 
Before:
`helm template --values=values.yaml . > before.yaml`
After:
`helm template --values=values.yaml . > after.yaml`
Compare:
`diff before.yaml after.yaml`

### DIFF
--- a/charts/aws-efs-csi-driver/CHANGELOG.md
+++ b/charts/aws-efs-csi-driver/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Helm chart
-# V2.4.2
+# v2.4.3
+* Add ability to configure daemonset affinity
+# v2.4.2
 * Bump app/driver version to `v1.5.5` 
 # v2.4.1
 * Bump app/driver version to `v1.5.4`
--- a/charts/aws-efs-csi-driver/Chart.yaml
+++ b/charts/aws-efs-csi-driver/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: aws-efs-csi-driver
-version: 2.4.2
+version: 2.4.3
 appVersion: 1.5.5
 kubeVersion: ">=1.17.0-0"
 description: "A Helm chart for AWS EFS CSI Driver"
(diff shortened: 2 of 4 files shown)


camaeel:
@smcavallo There is already PR for that: https://github.com/kubernetes-sigs/aws-efs-csi-driver/pull/796 
Hope yours gets more luck to be merged.

smcavallo:
@mskanth972 @Ashley-wenyizha - hoping you can review

kanton10062006:
It is still relevant. 
@Ashley-wenyizha @mskanth972 could you please take a look at this PR?

RyanStan (approved):

charts/aws-efs-csi-driver/CHANGELOG.md
```diff
@@ -1,4 +1,6 @@
 # Helm chart
+# v2.4.2
+* Add ability to configure daemonset affinity
```
Ashley-wenyizha, line 3: We do not need the helm chart update here

charts/aws-efs-csi-driver/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: aws-efs-csi-driver
-version: 2.4.1
+version: 2.4.2
```
Ashley-wenyizha, line 3: Same here, we will only bump this in a complete driver release process

RyanStan:
I think it's fine to let a new helm chart release be cut for this, I don't see a reason not to. But we'll need a rebase since we released 2.4.2 today (sorry for the poor timing).

Ashley-wenyizha:
Could we also give a rebase and only keep 1 commit here? otherwise lgtm, thanks!

smcavallo:
@Ashley-wenyizha  - how does it look now?  There's only a single commit now.

PR #983, opened 2023-04-11
Add fsType "efs" in storageclass.yaml example

### LINKED ISSUE #125: securityContext doesn't applied when creating pvc

ileixe:

Bug. Fix #125 

`fsType` is required for kubelet to apply fsGroup. Unless, kubelet failed to change ownership.

**What testing is done?** 

I manually confirmed that POD with fsGroup is correctly applied when storage class has the parameter.

### DIFF
--- a/examples/kubernetes/dynamic_provisioning/specs/storageclass.yaml
+++ b/examples/kubernetes/dynamic_provisioning/specs/storageclass.yaml
@@ -6,6 +6,7 @@ provisioner: efs.csi.aws.com
 parameters:
   provisioningMode: efs-ap
   fileSystemId: fs-92107410
+  fsType: efs
   directoryPerms: "700"
   gidRangeStart: "1000" # optional
   gidRangeEnd: "2000" # optional


ileixe:
This does not work actually as chown could not work: https://github.com/kubernetes-sigs/aws-efs-csi-driver/issues/300

ryan-wendel:
Access Points can be used to solve this issue.

https://github.com/kubernetes-sigs/aws-efs-csi-driver/blob/master/examples/kubernetes/access_points/README.md

PR #988, opened 2023-05-04
Modify efs-utils.conf in-place instead of overwriting

### LINKED ISSUE #737: Allow enabling stunnel_logs_file
Currently the template hard-codes a commented-out stunnel_logs_file, preventing users from turning on stunnel logging if they are encountering issues.

/feature

Ability to have a flag to turn on stunnel logs.

adammw:
New Feature

Modify efs-utils.conf in-place instead of overwriting. 
Expects the `/etc/amazon/efs-static-files/efs-utils.conf` file to exist and is copied over if the efs-utils.conf does not exist. 
Introduces `gopkg.in/ini.v1` ini parser to parse and rewrite the config file. Note that the rewritten file may have slightly different formatting but otherwise should be functionally identical.
Allows users to customise efs-utils.conf to enable debug logging or changing other settings without the driver overwriting the entire file.
Fixes #737 

**What testing is done?** 
Tested on a fork of release-1.4 in our own clusters.

### DIFF
--- a/pkg/driver/efs_watch_dog.go
+++ b/pkg/driver/efs_watch_dog.go
@@ -21,105 +21,12 @@ import (
 	"os/exec"
 	"path/filepath"
 	"sync"
-	"text/template"
 
+	"gopkg.in/ini.v1"
 	"k8s.io/klog"
 )
 
-// https://github.com/aws/efs-utils/blob/v1.30.2/dist/efs-utils.conf
 const (
-	efsUtilsConfigTemplate = `
-#
-# Copyright 2017-2018 Amazon.com, Inc. and its affiliates. All Rights Reserved.
-#
-# Licensed under the MIT License. See the LICENSE accompanying this file
-# for the specific language governing permissions and limitations under
-# the License.
-#
-
-[DEFAULT]
-logging_level = INFO
-logging_max_bytes = 1048576
-logging_file_count = 10
-# mode for /var/run/efs and subdirectories in octal
-state_file_dir_mode = 750
-
-[mount]
-dns_name_format = {az}.{fs_id}.efs.{region}.{dns_name_suffix}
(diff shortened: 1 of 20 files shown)


adammw:
Tests passing, ready for review @Ashley-wenyizha

pkg/driver/efs_watch_dog.go
```diff
@@ -242,16 +149,24 @@ func copyFile(src, dst string) error {
 }
 
 func (w *execWatchdog) updateConfig(efsClientSource string) error {
-	efsCfgTemplate := template.Must(template.New("efs-utils-config").Parse(efsUtilsConfigTemplate))
-	f, err := os.Create(filepath.Join(w.efsUtilsCfgPath, efsUtilsConfigFileName))
+	filePath := filepath.Join(w.efsUtilsCfgPath, efsUtilsConfigFileName)
+	cfg, err := ini.Load(filePath)
```
RyanStan, line 153: Wouldn't this throw an error if the file doesn't exist?

RyanStan:
> Allows users to customise efs-utils.conf to enable debug logging or changing other settings without the driver overwriting the entire file.

What about this change allows users to modify efs-utils.conf differently from how they would before? It seems like either way, users would still need to directly modify this file. I might be misunderstanding the changes here.

adammw:
> What about this change allows users to modify efs-utils.conf differently from how they would before? I

The difference is previously the binary would write out this file every time, making it impossible to customise before the binary has started. In contrast, users wishing to customise the config can now with this change either create a custom Dockerfile or entrypoint that modifies the config, and that modified config will be used rather than overwritten. 

Executing into the container and trying to modify the file after the fact doesn't work for mounts that have already been established before modification, therefore this was needed for us to debug issues in one of our persistent EFS mounts.

PR #1007, opened 2023-05-24
25.6k8s

mskanth972:

This PR updates go and k8s packages and removes SSH from CI

**What testing is done?**


hack/e2e/eksctl.sh
```diff
@@ -68,14 +63,14 @@ function eksctl_create_cluster() {
 
   if [[ "$WINDOWS" == true ]]; then
```
Ashley-wenyizha, line 64: Nit : We do not support windows, but this is a day one issue, might want to remove this in a separate PR later

mskanth972: Sure

PR #1016, opened 2023-05-25
Post-release of v1.5.6

mskanth972:

**What testing is done?**


charts/aws-efs-csi-driver/CHANGELOG.md
mskanth972, line 1: Did not bump the helm version as it is already bumped previously in master by PR raised by a user.

PR #1018, opened 2023-05-30
Test case to make sure that side-car images are located in Private ECR repository

mskanth972:

**What testing is done?** 
Tested with the side car images which are not there in Private ECR repo which resulted in failure.


Ashley-wenyizha:
nice work, /lgtm

Issue #1025, opened 2023-06-05
fsType check during NodePublishVolume will cause issues for volumes created with storageclass set to "" fstype


jingxu97:
/kind bug

**What happened?**

If useing a StorageClass without the parameter fstype, then the component external-provisioner will use the default one set through the flag defaultFsType which is ext4 [1], later the PV ends up setting pv.Spec.PersistentVolumeSource.CSI.FSType = ext4 in [2].

Because of [3] there's now a validation in the controller/node CSI calls to check that the req.VolumeCapabilities is setting a valid fstype which for the EFS CSI Driver is now defined as either ("efs" or "") in [3] supportedFSTypes = []string{"efs", ""}, therefore because the PV fstype is ext4 it fails the validation.

[1] https://github.com/kubernetes-csi/external-provisioner/blob/master/pkg/controller/controller.go#L626
[2] https://github.com/kubernetes-csi/external-provisioner/blob/master/pkg/controller/controller.go#L920
[3] https://github.com/kubernetes-sigs/aws-efs-csi-driver/pull/612

**What you expected to happen?**

Volume create with storageclass using empty fstype should not fail on mount

**How to reproduce it (as minimally and precisely as possible)?**

Create a storageclass with empty fstype, and create volume and a pod to use the volume

**Anything else we need to know?**:

**Environment**
- Kubernetes version (use `kubectl version`):
- Driver version:

**Please also attach debug logs to help us better diagnose**

- Instructions to gather debug logs can be found [here](https://github.com/kubernetes-sigs/aws-efs-csi-driver/tree/master/troubleshooting)

jingxu97:
A possible solution to avoid checking the fstype during NodePublishVolume, and only check it during createVolume

msau42:
cc @jsafrane @RomanBednar

wongma7:
I don't see where ext4 is set, https://github.com/kubernetes-csi/external-provisioner/blob/7eed97a6c45d9542f097adaed0c7c38becd7f0db/cmd/csi-provisioner/csi-provisioner.go#L101, and we don't set default-fstype in the template we ship https://github.com/kubernetes-sigs/aws-efs-csi-driver/blob/master/deploy/kubernetes/base/controller-deployment.yaml#L67, so could it be specific to your efs-csi-controller Deployment, could you double check?

jingxu97:
[...]

PR #1027, opened 2023-06-07
Validate fstype on CreateVolume, but not NodePublishVolume

jingxu97:
bug
https://github.com/kubernetes-sigs/aws-efs-csi-driver/issues/1025

There are existing PVs that could have wrong fstype, adding new validation will break it. This change tries to workaround this to validate fstype during CreateVolume for new volumes, not on NodePublishVolume

**What testing is done?** 
unit test

### DIFF
--- a/pkg/driver/controller.go
+++ b/pkg/driver/controller.go
@@ -76,6 +76,9 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
 	if err := d.isValidVolumeCapabilities(volCaps); err != nil {
 		return nil, status.Error(codes.InvalidArgument, fmt.Sprintf("Volume capabilities not supported: %s", err))
 	}
+	if err := d.validateFStype(volCaps); err != nil {
+		return nil, status.Error(codes.InvalidArgument, fmt.Sprintf("Volume fstype not supported: %s", err))
+	}
 
 	var (
 		azName           string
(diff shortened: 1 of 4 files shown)


pkg/driver/controller.go
```diff
@@ -76,6 +76,9 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
 	if err := d.isValidVolumeCapabilities(volCaps); err != nil {
 		return nil, status.Error(codes.InvalidArgument, fmt.Sprintf("Volume capabilities not supported: %s", err))
 	}
+	if err := d.validateFStype(volCaps); err != nil {
```
wongma7, line 79: realizing now this would be a breaking change. Maybe it is sufficient to remove the check from nodePublish and simply ignore fsType in both nodePublish and createVolume paths

wongma7: even in our example storageclasses we neglect to set fsType. So i think we need to tolerate case where fsType is nonsense, unset, or ext4 :/

jingxu97: do you mean we remove the check for both nodePublish and createVolume?

wongma7: yeah. sorry for lack of clarity, i am not 100% sure. The likelihood of someone having an EFS storageclass with nonsense fsType is very small.

wongma7: let's keep it. As loong as there is a release note we should be OK

wongma7:
/approve

will let other maintainers have a chance to lgtm :+1:

jingxu97:
Just a friendly ping for review. @Ashley-wenyizha @mskanth972 @leakingtapan could you help review and lgtm? Thank you!

PR #1051, opened 2023-06-29
Merged information from EKS User Guide

pgasca:
Documentation update

Merged missing details from https://docs.aws.amazon.com/eks/latest/userguide/efs-csi.html into the existing sections that were noted as Kubernetes specific.

One reason for this is so that Kubernetes users will have full information needed on GitHub even if they aren't aware of the information in the EKS User Guide. Users will be able to more easily contribute to the content when the EKS User Guide no longer has an equivalent GitHub repo.

There was overlap between both sources, with some extra information in one place or another. After this PR, the EKS User Guide will be updated to point to the GitHub readmes where it makes sense to do so rather than have potentially contradictory information that could go out of sync. This is also more consistent with how more recent storage drivers are handled in the EKS User Guide, such as the Amazon File Cache CSI driver.

**What testing is done?** 
Extra details come from what is currently live for EKS User Guide.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -7,7 +7,7 @@
 The [Amazon Elastic File System](https://aws.amazon.com/efs/) Container Storage Interface (CSI) Driver implements the [CSI](https://github.com/container-storage-interface/spec/blob/master/spec.md) specification for container orchestrators to manage the lifecycle of Amazon EFS file systems.
 
 ### CSI Specification Compatibility Matrix
-| AWS EFS CSI Driver \ CSI Spec Version  | v0.3.0| v1.1.0 | v1.2.0 |
+| Amazon EFS CSI Driver \ CSI Spec Version  | v0.3.0| v1.1.0 | v1.2.0 |
 |----------------------------------------|-------|--------|--------|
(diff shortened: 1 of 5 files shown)


Ashley-wenyizha:
Content looks good to me, left comment. Could we also rebase the commits and leave a cleaner commit messages/history?

pgasca:
@Ashley-wenyizha , the numbering got fixed when I corrected the spacing for the Note inbetween the two steps. I've also added detail about how using the master branch isn't recommended for the driver and gave a link to allow users to see the active branches.

Yes, you can rebase the commits. I tried to figure that out but not sure how to do it from the website itself since I'm just using the GitHub forms to edit the readme files. Could you handle the rebase when merging the PR?

pgasca:
I added another commit removing mention that master isn't recommended (but keeping the link to active branches). Originally I added these statements based on my understanding from a different storage driver (https://github.com/kubernetes-sigs/aws-fsx-csi-driver/blob/master/docs/install.md), but this same recommendation may not apply to EFS CSI based on mentions of master throughout rest of readme.

PR #1054, opened 2023-07-03
Set CSI_NODE_NAME for node daemonset

RyanStan:
The CSI_NODE_NAME environment variable allows the driver (either node service or controller service) to avoid making calls to IMDS by instead querying the Kubernetes API for the EC2 instance ID (EKS attaches EC2 Instance ID to ProviderID field on Node spec).

We have already set this for the controller deployment. However, this needs to also be set for the node daemonset.

**What testing is done?** 
I deployed the driver to a EKS cluster where the EC2 nodes had hop limit = 1 (IMDS disabled). I then applied our dynamic provisioning example. The PVC was successfully bound and mounted onto the application Pod.


RyanStan:
/close 

Merged in another PR

PR #926, opened 2023-02-20
feat: add support for additional pod labels

jaygridley:
**Is this a bug fix or adding new feature?**
Feature

**What is this PR about? / Why do we need it?**
Extends Helm chart with additional Pod labels capability

**What testing is done?** 
Locally generated template from the Helm chart

PR #1067, opened 2023-07-14
Add custom pod labels values to csi node and controller

gadiener:
New feature for the HELM chart.

Added a new value in the HELM chart to set custom labels for the CSI node and controller pods, the same way we do with annotations.

**What testing is done?** 
Tested with and without the custom labels applied on EKS.


gadiener:
Duplicated of https://github.com/kubernetes-sigs/aws-efs-csi-driver/pull/926

PR #1086, opened 2023-08-02
Update IAM policy example

mjsoyeon:
Update this example policy to compatible with AWS managed policy.

### DIFF
--- a/docs/iam-policy-example.json
+++ b/docs/iam-policy-example.json
@@ -2,6 +2,7 @@
   "Version": "2012-10-17",
   "Statement": [
     {
+      "Sid": "AllowDescribe",
       "Effect": "Allow",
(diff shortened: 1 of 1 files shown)


docs/iam-policy-example.json
```diff
@@ -2,6 +2,7 @@
   "Version": "2012-10-17",
   "Statement": [
     {
+      "Sid": "AllowDescribe",
```
Ashley-wenyizha, line 5: https://docs.aws.amazon.com/aws-managed-policy/latest/reference/AmazonEFSCSIDriverPolicy.html

RyanStan:
Let's just remove this example IAM Policy and tell users to use the managed policy instead.

PR #1087, opened 2023-08-03
Test PR for hostNetwork

mjsoyeon:
- downgrade k8s version 1.22 to 1.20

**Is this a bug fix or adding new feature?**

**What is this PR about? / Why do we need it?**

**What testing is done?**

PR #1088, opened 2023-08-03
Update reading/writing without interruption after pod restart e2e test

torredil:

This PR has all the necessary components to enable running kOps / EKS e2e CI jobs in the latest version of K8s without `hostNetwork: true`. Will rebase after #1087 and #1054.

**What testing is done?** 

CI

### DIFF
--- a/hack/kops-patch.yaml
+++ b/hack/kops-patch.yaml
@@ -8,8 +8,14 @@ spec:
             "elasticfilesystem:CreateAccessPoint",
             "elasticfilesystem:DeleteAccessPoint",
             "elasticfilesystem:DescribeFileSystems",
-            "elasticfilesystem:DescribeAccessPoints"
+            "elasticfilesystem:DescribeAccessPoints",
+            "elasticfilesystem:DescribeMountTargets",
+            "ec2:DescribeAvailabilityZones"
           ],
           "Resource": "*"
         }
       ]
+
+  cloudConfig:
+    awsEBSCSIDriver:
+      managed: true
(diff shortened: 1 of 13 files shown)


hack/kops-patch.yaml
```diff
@@ -8,8 +8,14 @@ spec:
...
+            "elasticfilesystem:DescribeAccessPoints",
+            "elasticfilesystem:DescribeMountTargets",
+            "ec2:DescribeAvailabilityZones"
           ],
           "Resource": "*"
         }
       ]
+
+  cloudConfig:
+    awsEBSCSIDriver:
```
Ashley-wenyizha, line 20: `awsEFSCSIDriver`?

mjsoyeon: When k8s is installed through kops, ebs driver enables as a default since k8s 1.22

Ashley-wenyizha: when you say `This config managed ebs driver`, could you forward the link of the resource? and what do you mean by `ebs driver enables as a default`?

mjsoyeon: [New clusters running Kubernetes 1.22 will have AWS EBS CSI driver enabled by default.
](https://github.com/kubernetes/kops/blob/master/docs/releases/1.22-NOTES.md#:~:text=New%20clusters%20running%20Kubernetes%201.22%20will%20have%20AWS%20EBS%20CSI%20driver%20enabled%20by%20default.)

Ashley-wenyizha: I guess what I am confused here is I do not fully understand how is this controlled and why do we need to control it through ebs driver?

Could you add more details/explanations here? thanks

torredil: @Ashley-wenyizha To clarify, this patch simply disables the installation of the EBS CSI Driver during cluster creation with kOps. This is not being through the EBS CSI Driver -- it is part of kOps cluster configuration.

Ashley-wenyizha: yep I have communicated offline with Soyeon. lgtm

PR #1117, opened 2023-08-30
Add unit test - subpath pattern, repeated elements

seanzatzdev-amazon:
Adds new unit test.
This adds a unit test for the subpath pattern to test if repeated elements work correctly, with repeated elements referring to using elements such as the PVC name multiple times in the subpath pattern.

The reason for this unit test is to ensure storage classes which use elements such as the PVC name multiple times in their subpath pattern do not encounter any errors.

**What testing is done?** 
Unit tests ran successfully.

### DIFF
--- a/pkg/driver/controller_test.go
+++ b/pkg/driver/controller_test.go
@@ -1248,6 +1248,78 @@ func TestCreateVolume(t *testing.T) {
 			},
 		},
 		{
+			name: "Success: Normal flow with a valid directory structure set, using repeated elements (uses PVC Name in subpath pattern multiple times)",
+			testFunc: func(t *testing.T) {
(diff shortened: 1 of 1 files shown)


pkg/driver/controller_test.go
```diff
@@ -1247,6 +1247,78 @@ func TestCreateVolume(t *testing.T) {
 				mockCtl.Finish()
 			},
 		},
+		{
+			name: "Success: Normal flow with a valid directory structure set, using a repeated elements",
```
Ashley-wenyizha, line 1251: Add bit more info of "elements"? also could you add more description for the test case and initiative in the overview?

seanzatzdev-amazon: Resolved. Thank you.

PR #1120, opened 2023-09-07
Add e2e test for dynamic provisioning

seanzatzdev-amazon:
Adds new e2e test

This PR adds an e2e test for dynamic provisioning. It simply performs dynamic provisioning, then writes to and reads from the persistent volume to ensure that everything works correctly. 

This is important as currently in the EFS-CSI test suite we do not have e2e tests which cover Dynamic Provisioning, and thus we do not have e2e test coverage for EFS-specific CSI Driver features such as "ensureUniqueDirectory". Adding this test allows us more coverage, and will additionally allow us to more easily add further coverage for additional features in the future.

**What testing is done?** 
Test passes on local.

### DIFF
--- a/test/e2e/e2e.go
+++ b/test/e2e/e2e.go
@@ -360,8 +360,107 @@ var _ = ginkgo.Describe("[efs-csi] EFS CSI", func() {
 			encryptInTransit := false
 			testEncryptInTransit(f, &encryptInTransit)
 		})
+
+		ginkgo.It("should successfully perform dynamic provisioning", func() {
+
+			ginkgo.By("Creating EFS Storage Class, PVC and associated PV")
+			params := map[string]string{
+				"provisioningMode":      "efs-ap",
+				"fileSystemId":          FileSystemId,
+				"subPathPattern":        "${.PVC.name}",
+				"directoryPerms":        "700",
+				"gidRangeStart":         "1000",
+				"gidRangeEnd":           "2000",
+				"basePath":              "/dynamic_provisioning",
+				"ensureUniqueDirectory": "true",
+			}
+
+			sc := GetStorageClass(params)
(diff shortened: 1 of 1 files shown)


mskanth972:
I think we already have dynamic provisioning e2e test cases [link](https://github.com/kubernetes-sigs/aws-efs-csi-driver/blob/master/test/e2e/e2e.go#L112). And also we have a test cases to check the read/write functionality [link](https://github.com/kubernetes-sigs/aws-efs-csi-driver/blob/master/test/e2e/e2e.go#L276). Should we need this again?

seanzatzdev-amazon:
> link

Hey Shashi, what you say is true. Though the dynamic provisioning definition you linked is only for use by the kubernetes e2e tests which as I understand do not test EFS-specific features (such as storage class features like subpath pattern, reusingAccessPoints, etc). The reading/writing is also tested currently but for static provisioning.

Combining both of these aspects into a new EFS-specific e2e test will allow us to add more robustness in the tests later on.

mskanth972:
> > I think we already have dynamic provisioning e2e test cases [link](https://github.com/kubernetes-sigs/aws-efs-csi-driver/blob/master/test/e2e/e2e.go#L112). And also we have a test cases to check the read/write functionality [link](https://github.com/kubernetes-sigs/aws-efs-csi-driver/blob/master/test/e2e/e2e.go#L276). Should we need this again?
> 
> Hey Shashi, what you say is true. Though the dynamic provisioning definition you linked is intended to be used by the kubernetes e2e tests which as I understand do not test EFS-specific features (such as storage class features like subpath pattern, reusingAccessPoints, etc). The reading/writing is also tested currently but for static provisioning.
> 
> Combining both of these aspects into a new EFS-specific e2e test will allow us to add more robustness in the tests later on.

Okay, it makes sense.

PR #1133, opened 2023-09-14
Fix ensureUniqueDirectory paramater at docs

wilkerlucio:
I noticed the name for `ensureUniqueDirectory` is wrong in the README, its correct at the example, but wrong here.

Fix in docs

The current docs mislead the parameter due to the wrong naming


mskanth972:
Hi @wilkerlucio, thanks for the PR. We will merge it.

PR #1144, opened 2023-09-20
Pre-release PR for v1.7.0

seanzatzdev-amazon:

Pre-release PR for v1.7.0
**What testing is done?**


runningman84:
hi @seanzatzdev-amazon do you plan to release this within the days? Do you have a prerelease which we can already use?

PR #1162, opened 2023-10-11
Fix storage class api version

mirgj:

bug fix

The previous yaml definition for storage classes were using a wrong api version which resulted in this error: `no matches for kind "StorageClass" in version "v1"`

This PR address this issue by replacing the API version for the storage class to use the right attribute: `storage.k8s.io/v1`

**What testing is done?** 

by using:
```
kubectl apply -f storage-class.yaml
```

it should apply your storage class correctly

verify that with:

```
$ kubectl get storageclass
NAME            PROVISIONER             RECLAIMPOLICY   VOLUMEBINDINGMODE      ALLOWVOLUMEEXPANSION   AGE
efs-sc          efs.csi.aws.com         Delete          Immediate              false                  12m
```

### DIFF
--- a/examples/kubernetes/statefulset/specs/example.yaml
+++ b/examples/kubernetes/statefulset/specs/example.yaml
@@ -20,7 +20,7 @@ spec:
     driver: efs.csi.aws.com
     volumeHandle: fs-4af69aab
 ---
-apiVersion: v1
+apiVersion: storage.k8s.io/v1
 kind: PersistentVolumeClaim
 metadata:
   name: efs-claim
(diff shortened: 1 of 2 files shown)


gcaracuel:
Note this requires Helm changes too, as it was done for Kustomize here https://github.com/kubernetes-sigs/aws-efs-csi-driver/blob/master/charts/aws-efs-csi-driver/templates/storageclass.yaml#L3 change to https://github.com/kubernetes-sigs/aws-efs-csi-driver/blob/master/charts/aws-efs-csi-driver/templates/storageclass.yaml#L3 is required

PR #1163, opened 2023-10-11
Add note on EFS mount target security group

mikereinhold:
Documentation enhancement

Add a note clarifying that the EFS mount target security group may require additional ingress rules for secondary CIDR ranges when secondary networking configurations are used.

**What testing is done?** 
Just finished troubleshooting an EFS CSI driver configuration on a cluster that uses secondary networking.


mskanth972:
Hi @mikereinhold, thanks for bringing here. It's a good point to add in the README for the customers using Secondary CIDR ranges.

Issue #1164, opened 2023-10-16
List access point fails in case there's an AP w/o POSIX user


yaronya:
/kind bug

**What happened?**
I had an AP created manually under the EFS configured for the CSI driver. The AP did not have a Posix User configured.
After I upgraded to 2.5.0 - I saw that the controller is stuck in crash loop.

**What you expected to happen?**
Skip the AP, or at least print an error/warning without panic.

**How to reproduce it (as minimally and precisely as possible)?**
1. Create an AP without Posix User configured for a given EFS
2. Configure a `StorageClass` to use the `fileSystemId` of that EFS
3. Controller pod enters a crash loop, logs will yield the below panic log.

**Anything else we need to know?**:

**Environment**
- Kubernetes version (use `kubectl version`):
```
Client Version: v1.28.1
Kustomize Version: v5.0.4-0.20230601165947-6ce0bf390ce3
Server Version: v1.25.12-eks-2d98532
```
- Driver version: 2.5.0

**Please also attach debug logs to help us better diagnose**
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x16220e6]

goroutine 101 [running]:
github.com/kubernetes-sigs/aws-efs-csi-driver/pkg/cloud.(*cloud).ListAccessPoints(0xc000168380, {0x1e5e9b0, 0xc0008048d0}, {0xc00013ec20, 0xb})
        /go/src/github.com/kubernetes-sigs/aws-efs-csi-driver/pkg/cloud/cloud.go:317 +0x2e6
github.com/kubernetes-sigs/aws-efs-csi-driver/pkg/driver.(*GidAllocator).getUsedGids(0x6acfc0?, {0x1e5e9b0?, 0xc0008048d0?}, {0xc00013ec20, 0xb})
        /go/src/github.com/kubernetes-sigs/aws-efs-csi-driver/pkg/driver/gid_allocator.go:66 +0x53
github.com/kubernetes-sigs/aws-efs-csi-driver/pkg/driver.(*GidAllocator).getNextGid(0xc0006163d8, {0x1e5e9b0, 0xc0008048d0}, {0xc00013ec20, 0xb}, 0x1e708d0?, 0x19cf700?)
        /go/src/github.com/kubernetes-sigs/aws-efs-csi-driver/pkg/driver/gid_allocator.go:42 +0x1fe
[...]

PR #1167, opened 2023-10-20
When listing AccessPoints, check if efs.PosixUser pointer is nil

JeanMassip:

This is a fix for the bug reported in issue #1164 

A simple bugfix 

**What testing is done?** 

All unit test passed.


mskanth972:
Hi @JeanMassip, thanks for bringing here. Can you get EasyCLA Authorization?

JeanMassip:
> Hi @JeanMassip, thanks for bringing here. Can you get EasyCLA Authorization?

Already did twice, doesn't seem to register though :/

PR #1168, opened 2023-10-20
Fix typo in installation with public manifest

ysam12345:
Document fix.

The filename of the downloaded public manifest should be `public-ecr-driver.yaml` rather than `private-ecr-driver.yaml`.

**What testing is done?** 
N/A


mskanth972:
Hi @ysam12345 thanks for bringing here. Merging it.

PR #1170, opened 2023-10-25
Fixed Duplicate Client-token issue

mskanth972:
Currently the client-token of an EFS Access Point is unique only to region and not to a specific File system. So when using `reuseAccessPoint` feature there maybe a chance that two different File systems can use the same PVC name with this parameter enabled and which can lead to `ErrorCode: "AccessPointAlreadyExists"`. To overcome this I just added the FS-ID also to the hashed value along with PVC name.

**What testing is done?** 
Already have the existing test cases around the hash

### DIFF
--- a/pkg/driver/controller.go
+++ b/pkg/driver/controller.go
@@ -95,7 +95,7 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
 			return nil, status.Error(codes.InvalidArgument, "Invalid value for reuseAccessPoint parameter")
 		}
 		if reuseAccessPoint {
-			clientToken = get64LenHash(volumeParams[PvcNameKey])
+			clientToken = get64LenHash(volumeParams[PvcNameKey] + volumeParams[FsId])
 			klog.V(5).Infof("Client token : %s", clientToken)
 		}
 	}
--- a/pkg/driver/controller_test.go
+++ b/pkg/driver/controller_test.go
@@ -810,7 +810,7 @@ func TestCreateVolume(t *testing.T) {
 						Uid: 1000,
 					},
 				}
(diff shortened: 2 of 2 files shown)


joaocc:
Hi. 
We just encountered this issue. In a single EKS cluster, we use several EFS instances which are accessed from different namespaces and with same PVC name (inside the ns). Driver is v1.7.7.
So, for instance
* `EFS-01` and `EFS-02`
* `EKS-A` with 
  * storageClass `dyn-sc-to-efs-01` and `dyn-sc-to-efs-02`, with `reuseAccessPoint=true`
  * namespaces `client-01` and `client-02`
* each client uses PVC names `my-data` (in the corresponding NS) of respectively `dyn-sc-to-efs-01` and `dyn-sc-to-efs-02`

Client-1 is created ok, but when adding client-2, we start getting the errors mentioned above.
It seems even thought the target EFS specified by the `dyn-sc-to-efs-02` is different (`EFS-02` instead of `EFS-01`) the driver is complaining that the access point already exists in `EFS-01`.

If this is the intended behaviour, this means that the potential for conflict is huge, as any cluster can "block" any other cluster in same region that uses the `reuseAccessPoint=true`, but creating a PVC in any other EFS with the same name and settings. Also, if this is intended behaviour, the documentation can be made clearer and be more explicit in explaining the implications (security and operations-wise).

If this was not intended behaviour, could you please let us know when this could be merged?
Thanks

mskanth972:
Hi @joaocc are those file systems in different regions?

joaocc:
Hi. Thanks. We already removed/disabled the param, as the understanding was incorrect and it is not needed. I tried to bump this so that all the work already done could be submitted and save someone else the headache :) Thanks!

PR #1193, opened 2023-11-14
Updated the apiversion in storageclass.yaml

Rajit11:
The SC manifest contained a deprecated ApiVersion, updated the same.

Bug Fix

The sc manifest wouldn't work without updating the apiversion, so the efs-csi-driver check is useless without it.

**What testing is done?** 
I deployed the whole stack after updating the apiversion of the sc manifest and it worked absolutely fine.


mskanth972:
Hi @Rajit11, thank for the PR. You are true, will merge this change.

Rajit11:
Thanks! @mskanth972

Issue #1194, opened 2023-11-15
Add new dedicated regions to watchdog


jdwtf:
/kind bug

**What happened?**
`efs_watch_dog` does not allow rpc in dedicated regions unless they are specified in the config.

**What did you expect to happen?**
The driver to work in APA.

**How to reproduce it (as minimally and precisely as possible)?**
This is only reproducible in the the dedicated region, but basically the driver deploys fine but then begins to refuse rpc connections because the new regions are not present in the config.

**Anything else we need to know?**:
I manually implemented a fix and successfully tested it in the the APA region. Contributing guidelines state I need to open an issue before I can fork/pull req the fix.

**Environment**
- Kubernetes version (use `kubectl version`): 1.23-1.28
- Driver version: 1.7.0

**Please also attach debug logs to help us better diagnose**
Apologies, but I cannot attach logs from dedicated regions.

- Instructions to gather debug logs can be found [here](https://github.com/kubernetes-sigs/aws-efs-csi-driver/tree/master/troubleshooting)

jdwtf:
Fixes kubernetes-sigs/aws-efs-csi-driver#1194

PR #1196, opened 2023-11-16
Adding APA and FFZ regions to watchdog config template

jdwtf:
Sort of both. It's a bug because it's preventing the driver from working in new ADC regions but it will enable the driver in the regions so it's kind of a feature too.

This is to enable the driver to run in two new dedicated regions.

**What testing is done?** 
Changes were manually tested in dedicated regions and fixed the issue.


jdwtf:
Fixes kubernetes-sigs/aws-efs-csi-driver#1194

PR #1201, opened 2023-11-29
Added checks to make sure delete-access-point should not delete entir…

mskanth972:
…e EFS

This is a bug fix

Added checks to make sure delete-access-point should not delete entire EFS

**What testing is done?**

### DIFF
--- a/pkg/driver/controller.go
+++ b/pkg/driver/controller.go
@@ -21,6 +21,7 @@ import (
 	"crypto/sha256"
 	"fmt"
 	"github.com/google/uuid"
+	"io"
 	"os"
 	"path"
 	"sort"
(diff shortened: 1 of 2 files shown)


pkg/driver/controller.go
```diff
@@ -394,11 +395,28 @@ func (d *Driver) DeleteVolume(ctx context.Context, req *csi.DeleteVolumeRequest)
 				}
 			}
 
+			klog.V(2).Infof("Access point root directory : %s", accessPoint.AccessPointRootDir)
+			if accessPoint.AccessPointRootDir == "/" {
```
the-technat, line 399: What happens when `accessPoint.AccessPointRootDir` is an empty string? (The field seems not mandatory when creating an AP, so could in rare cases be empty).

According to my understanding the efs-plugin mounts the entire filesystem in `/var/lib/csi/pv/<fsap-id>` and later calls `os.RemoveAll(target, accessPoint.AccessPointRootDir)`. Now if `accessPoint.AccessPointRootDir` is empty, the left-over path for `os.RemoveAll` is `/var/lib/csi/pv/<fsap-id>` which represents the entire filesystem. I'm I right with that?

RyanStan: We pull AccessPointRootDir from an API call to the AWS EFS console API:

https://github.com/kubernetes-sigs/aws-efs-csi-driver/blob/8b7839300505efcdebcd77dba43e34c5b33b7ded/pkg/driver/controller.go#L374

The API layer has empty access point root dirs default to the root path, so an empty access point root dir shouldn't be possible.

But, I agree, we should validate our assumptions and add a check, just in case.

mkilchhofer:
/remove-lifecycle rotten

@dobsonj @mskanth972 is anything missing here or why did you not merge it?

dankova22:
Addressed in  #1592  and #1532

PR #1152, opened 2023-09-22
Additional Controller Deployment features

nishant221:
**Is this a bug fix or adding new feature?**
New feature

**What is this PR about? / Why do we need it?**
Adding following support/flexibility in Controller Deployment Helm chart template:
- additionalLabels: To append additional Deployment labels.
- podLabels: To append additional Pod labels.
- hostNetwork: For flexibility of defining true or false.
- priorityClassName:  For flexibility of defining custom priorityClassName.
- dnsPolicy:  For flexibility of defining different dnsPolicy.
- dnsConfig: For flexibility of defining custom dnsConfig.
- leaderElectionRenewDeadline & leaderElectionLeaseDuration: For defining csi-provisioner parameters.
- containerSecurityContext:  For flexibility of defining SecurityContext on container.

**What testing is done?** 
Manual/CI

PR #1203, opened 2023-12-01
fix(helm): allow controller to run in hostNetwork

the-technat:

Partially reverts 
- https://github.com/kubernetes-sigs/aws-efs-csi-driver/pull/910 

Allows hostNetwork to optionally be enabled on the helm chart. 

**What testing is done?** 
When upgrading the efs-csi-driver from helm chart version 2.3.6 to 2.4.5 we noticed certain instabilities in our environment that we suppose are due to the controller no longer running in hostNetwork. Thus we would like to thave the option to enable this again.

### DIFF
--- a/charts/aws-efs-csi-driver/templates/controller-deployment.yaml
+++ b/charts/aws-efs-csi-driver/templates/controller-deployment.yaml
@@ -27,6 +27,7 @@ spec:
       annotations: {{- toYaml . | nindent 8 }}
       {{- end }}
     spec:
+      hostNetwork: {{ .Values.controller.hostNetwork }}
       {{- if .Values.imagePullSecrets }}
       imagePullSecrets:
         {{- range .Values.imagePullSecrets }}
--- a/charts/aws-efs-csi-driver/values.yaml
+++ b/charts/aws-efs-csi-driver/values.yaml
@@ -96,7 +96,7 @@ controller:
     runAsUser: 0
     runAsGroup: 0
     fsGroup: 0
-
+  hostNetwork: false 
 
 ## Node daemonset variables
 


RyanStan:
Looks good to me. Could you expand on what kind of instabilities you're seeing though?

RyanStan:
Also, I see that this will be added as part of 
https://github.com/kubernetes-sigs/aws-efs-csi-driver/pull/1152. Let me take a look at that PR and see if I can get it pushed along

the-technat:
> Looks good to me. Could you expand on what kind of instabilities you're seeing though?

In our environment it's mainly due to the CNI. We use cilium in chaining mode with the aws-vpc-cni and experience a lot of situations where pods aren't managed by cilium (due to the technical nature of two CNIs beeing chained together). Pods not managed by cilium means they got an IP from the aws-vpc-cni but cilium doesn't recognize them and thus blocks traffic due to NetworkPolicies beeing implemented by cilium. So we essentially have an unstable networking setup and the controller can't really deal with that I think. Moving the controller to hostNetwork solves this problem for us.

Other potential cause I see is the fact that the controller mounts the EFS inside the pod for cleanup of folders which means the container runs privileged, but still the mount is done in a network sandbox. For me that sounds like a potential for errors and hostNetwork would skip this.

the-technat:
> Also, I see that this will be added as part of #1152. Let me take a look at that PR and see if I can get it pushed along

I know, we've seen this PR but since it hasn't got any updates recently we though we'll try if this one will be merged faster ;).

RyanStan:
Thanks for the details. Yep I just looked at the other PR and there's one thing to clean up, so we'll push along this one.

the-technat:
ping @RyanStan, @mskanth972

the-technat:
Any news on this @RyanStan, @mskanth972?

the-technat:
Seems like this was already implemented with #1152

Issue #1141, opened 2023-09-18
helm chart: `node.hostAliases` does not support multiple AZ for the same FS id


jj-mino:
/kind bug

**What happened?**
When we pass below values to the helm chart to support multiple availability zones, only one entry is added, as the others are overwritten by [this](https://github.com/kubernetes-sigs/aws-efs-csi-driver/blob/v1.6.0/charts/aws-efs-csi-driver/templates/node-daemonset.yaml#L28-L33) logic:
```yaml
node:
  hostAliases:
    fs-010101010101010110: 
      ip: "10.6.0.5" # mount target IP for az1
      region: "us-west-2"
    fs-010101010101010110: 
      ip: "10.6.0.6" # mount target IP for az2
      region: "us-west-2"
    fs-010101010101010110: 
      ip: "10.6.0.7" # mount target IP for az3
      region: "us-west-2"
```

**What you expected to happen?**
I would like to set multiple IPs for the same Filesystem ID

**How to reproduce it (as minimally and precisely as possible)?**
1) Deploy the helm chart with above values
2) Review the daemonset configured hostaliases (`kubectl edit daemonset efs-csi-node -n kube-system`)

**Anything else we need to know?**:

**Environment**
- Kubernetes version (use `kubectl version`): 
  - Client Version: v1.28.1
  - Kustomize Version: v5.0.4-0.20230601165947-6ce0bf390ce3
  - Server Version: v1.27.4-eks-2d98532
- Driver version: aws-efs-csi-driver helm chart `v2.4.9`

**Please also attach debug logs to help us better diagnose**

- Instructions to gather debug logs can be found [here](https://github.com/kubernetes-sigs/aws-efs-csi-driver/tree/master/troubleshooting)

Ashley-wenyizha:
Hi @jj-mino,
 
Could you upload debug logs per instruction (the one you linked)?

thanks

davejfranco:
@Ashley-wenyizha I'm facing the same challenge the PR above should solve the problem

PR #1206, opened 2023-12-02
[Fix] Allow to add multiple IPs per filesystem ID

davejfranco:
This should solves Issue: [1141](https://github.com/kubernetes-sigs/aws-efs-csi-driver/issues/1141)

Currently de `node-daemonset.yaml` loops over a map of host aliases where filesystemID is the key
```
    {{- with .Values.node.hostAliases }}
      hostAliases:
      {{- range $k, $v := . }}
        - ip: {{ $v.ip }}
          hostnames:
            - {{ $k }}.efs.{{ $v.region }}.amazonaws.com
      {{- end }}
    {{- end }}
```
```
node:
  # Number for the log level verbosity
  logLevel: 2
  volMetricsOptIn: false
  volMetricsRefreshPeriod: 240
  volMetricsFsRateLimit: 5
  hostAliases:
    {}
    # For cross VPC EFS, you need to poison or overwrite the DNS for the efs volume as per
    # https://docs.aws.amazon.com/efs/latest/ug/efs-different-vpc.html#wt6-efs-utils-step3
    # implementing the suggested solution found here:
    # https://github.com/kubernetes-sigs/aws-efs-csi-driver/issues/240#issuecomment-676849346
    # EFS Vol ID, IP, Region
    # "fs-01234567":
    #   ip: 10.10.2.2
    #   region: us-east-2
```
This caused that if you add the same filesystem ID that have multiple IPs per AZ as mentioned in the issue that only one entry gets added as you can't have a map with multiple entries with the same key.

This should fix  this and allow multiple host aliases where each filesystem can have multiple IPs per AZ
**What testing is done?** 
Installing the helm chart manually on a EKS cluster.

This is part of a same value file
```
node:
  # Number for the log level verbosity
  logLevel: 2
  volMetricsOptIn: false
  volMetricsRefreshPeriod: 240
  volMetricsFsRateLimit: 5
  hostAliases:
    # For cross VPC EFS, you need to poison or overwrite the DNS for the efs volume as per
    # https://docs.aws.amazon.com/efs/latest/ug/efs-different-vpc.html#wt6-efs-utils-step3
    # implementing the suggested solution found here:
    # https://github.com/kubernetes-sigs/aws-efs-csi-driver/issues/240#issuecomment-676849346
    # EFS Vol ID, IP, Region
    # https://github.com/kubernetes-sigs/aws-efs-csi-driver/issues/1141
    - fsid: fs-01234567
      region: eu-west-1
      ips: 
        - 10.10.10.1
        - 10.10.10.2
        - 10.10.10.3
```
This is the resulting daemonset

```
      dnsPolicy: ClusterFirst
      hostAliases:
      - hostnames:
        - fs-01234567.efs.eu-west-1.amazonaws.com
        ip: 10.10.10.1
      - hostnames:
        - fs-01234567.efs.eu-west-1.amazonaws.com
        ip: 10.10.10.2
      - hostnames:
        - fs-01234567.efs.eu-west-1.amazonaws.com
        ip: 10.10.10.3
```

### DIFF
--- a/charts/aws-efs-csi-driver/templates/node-daemonset.yaml
+++ b/charts/aws-efs-csi-driver/templates/node-daemonset.yaml
@@ -25,12 +25,15 @@ spec:
       annotations: {{ toYaml .Values.node.podAnnotations | nindent 8 }}
       {{- end }}
     spec:
-    {{- with .Values.node.hostAliases }}
+    {{- if .Values.node.hostAliases }}
       hostAliases:
-      {{- range $k, $v := . }}
-        - ip: {{ $v.ip }}
+      {{- range .Values.node.hostAliases }}
+        {{- $alias := . }} 
+        {{- range $alias.ips }}
+        - ip: {{ . }}
           hostnames:
-            - {{ $k }}.efs.{{ $v.region }}.amazonaws.com
+            - {{ $alias.fsid }}.efs.{{ $alias.region }}.amazonaws.com
+          {{- end }}
       {{- end }}
     {{- end }}
     {{- if .Values.imagePullSecrets }}
--- a/charts/aws-efs-csi-driver/values.yaml
+++ b/charts/aws-efs-csi-driver/values.yaml
@@ -108,16 +108,19 @@ node:
   volMetricsOptIn: false
   volMetricsRefreshPeriod: 240
   volMetricsFsRateLimit: 5
-  hostAliases:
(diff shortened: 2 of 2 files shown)


davejfranco:
I was not aware that we cannot have a DNS entry with multiple IPs therefore this PR makes no sense so I'm closing it

PR #850, opened 2022-12-05
Fix GID allocator

RomanBednar:
**Is this a bug fix or adding new feature?**
Bug fix.

**What is this PR about? / Why do we need it?**

Dynamically provisioning EFS volume creates an access point (AP) for which it allocates a GID from default range or a range defined in SC parameters.

The GID allocator did not check what GIDs might be already in use. For example let's say we have an EFS with access point for `/mydir` path which is owned by GID `1000` with permissions of `770` that grant rwx access to group `1000`. Then we create another access point with same path or "higher" (can be also `/`) with the same GID of `1000`. Now whoever mounts the second access point will have rwx access to `/mydir` because the GID matches. The old code does not prevent this and so volume creation done by the driver might result in creating access point with a GID that is already used, and when it gets mounted to a pod this pod might get access to other pods data.

Additionally, when user changed the SC parameters it was not reflected in the driver (controller pod) and GID allocator kept allocating from the range that got in first (for a given FS ID). In order for GID range changes to take effect the pod had to be restarted.

**What testing is done?** 

- current unit tests
- new unit test added
- manual testing

Issue #1225, opened 2023-12-28
Can't mount EFS to a pod: AccessPointAlreadyExists


ig-leonov:
After upgrade the driver version from 1.5.5 to 1.7.0 we started to see some pods failed to mount EFS with exception like:
`I1227 14:23:13.534226       1 event.go:298] Event(v1.ObjectReference{Kind:""PersistentVolumeClaim"", Namespace:""_NAMESPACE_"", Name:""_NAME_"", UID:""60171e27-5916-44a7-a8f2-a93deed684f6"", APIVersion:""v1"", ResourceVersion:""4119172"", FieldPath:""""}): type: 'Warning' reason: 'ProvisioningFailed' failed to provision volume with StorageClass ""_SC_NAME_HERE_"": rpc error: code = Internal desc = Failed to create Access point in File System _FS_ID_HERE_ : Failed to create access point: AccessPointAlreadyExists: `

This is happening during high load (we may have hundreds of pods starting simultaneously), and efs csi driver logs mostly contain Throttling exceptions. We're using Dynamic Provisioning so access points are created and deleted by efs driver.

Attaching logs from csi provisioner for specific pvc and efs csi driver for the same timeframe:

[csi_provisioner_logs.txt](https://github.com/kubernetes-sigs/aws-efs-csi-driver/files/13787309/csi_provisioner_logs.txt)
[efs_csi_driver_logs.txt](https://github.com/kubernetes-sigs/aws-efs-csi-driver/files/13787310/efs_csi_driver_logs.txt)

This issue is not reproducible in v1.5.5, and we suspect that https://github.com/kubernetes-sigs/aws-efs-csi-driver/pull/850/commits may be one of the things that is causing it (increasing the amount of requests for DescribeAccessPoint).

Other things I think may be useful:

- this warning looks weird:
  `W1227 14:19:14.769663       1 gid_allocator.go:93] Requested GID range (50000:7000000) exceeds EFS Access Point limit (1000) per Filesystem. Driver will not allocate GIDs outside of this limit.`
 given that we  use default GID range.
- We noticed that Access Points created for these PVCs are not automatically deleted with associated PVC 

**Environment**
- Kubernetes version v1.26.10-eks-4f4795d:
- Driver version: 1.7.0

otorreno:
[...]

PR #1226, opened 2023-12-29
Reduce calls to EFS API

otorreno:
Bug fix

With #850, the new way of allocating GIDs introduced a new call to the ListAccessPoints endpoint of EFS API, that is generating problems on systems where EFS CSI driver is under high load (lots of PVC created within a short time period). In this PR, we are extracting the ListAccessPoints call from gid_allocator, moving it one level up. In case of dynamic provisioining of GIDs we can reuse the ListAccessPoints call to check for the file system existence (thus removing the DescribeFileSystem call in such case). In case of a fixed UID/GID, we continue calling DescribeFileSystem, and no calls to ListAccessPoints.

In addition to the change explained above, gidMin and gidMax have been converted to int64. #850 made both uid and gid int64, but gidMin and gidMax were nout touched. Also changing the default value for gidMax, as the value of 7000000 was spamming the logs with a message coming from gid_allocator (i.e. range bigger than the max number of access points). Setting the value to 51000 (the value that gid_allocator was setting by adding 1000 to gidMin). Removing an unused function and field from gid_allocator too.

**What testing is done?** 
Coverage with unit tests

### DIFF
--- a/pkg/cloud/cloud.go
+++ b/pkg/cloud/cloud.go
@@ -302,10 +302,10 @@ func (c *cloud) ListAccessPoints(ctx context.Context, fileSystemId string) (acce
 	res, err := c.efs.DescribeAccessPointsWithContext(ctx, describeAPInput)
 	if err != nil {
 		if isAccessDenied(err) {
-			return
+			return nil, ErrAccessDenied
 		}
 		if isFileSystemNotFound(err) {
-			return
+			return nil, ErrNotFound
 		}
 		err = fmt.Errorf("List Access Points failed: %v", err)
 		return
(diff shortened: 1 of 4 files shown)


otorreno:
#1217 should be merged before this one, as it is the reason #1225 is happening I believe. I will then rebase and fix the merge conflicts here

otorreno:
@mskanth972 you may want to review this one as it is in the same area of code of #1217

mskanth972:
Hi @otorreno, the respected [PR](https://github.com/kubernetes-sigs/aws-efs-csi-driver/pull/1217) is merged, can you rebase this PR? 
Thanks

otorreno:
@mskanth972 rebased on top of latest master

otorreno:
> /approve

Thank you for the review!

PR #1244, opened 2024-01-12
fix bottlerocket compatibility

mweberjc:
bottlerocket nodes have a r/o file system:

```
[pod/efs-csi-node-cxcfc/csi-driver-registrar] E0111 23:11:21.812117       1 main.go:107] "Failed to create registration probe file" err="mkdir /var/lib/kubelet/plugins/efs.csi.aws.com: read-only file system" registrationProbePath="/var/lib/kubelet/plugins/efs.csi.aws.com/registration"
```

with this change:

```
[pod/efs-csi-node-kg9dt/csi-driver-registrar] I0111 23:14:01.377424       1 main.go:109] "Kubelet registration probe created" path="/var/lib/kubelet/plugins/efs.csi.aws.com/registration"
```

bug fix

bug fix for bottlerocket node compatibility

**What testing is done?** 

verified deployment on bottlerocket

### DIFF
--- a/charts/aws-efs-csi-driver/templates/node-daemonset.yaml
+++ b/charts/aws-efs-csi-driver/templates/node-daemonset.yaml
@@ -133,6 +133,8 @@ spec:
               mountPath: /csi
             - name: registration-dir
               mountPath: /registration
+            - name: probe-dir
+              mountPath: /var/lib/kubelet/plugins/efs.csi.aws.com/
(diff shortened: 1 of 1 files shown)


z0rc:
As far as I can tell, this error doesn't happen anymore with driver v2.1.6 and Bottlerocket OS 1.32.0. Still unsure when and how exactly this was resolved.

Issue #1091, opened 2023-08-08
Buggy check for non-negative gid


joelthompson:
/kind bug

**What happened?**

See https://github.com/kubernetes-sigs/aws-efs-csi-driver/blob/f52b4c089825d0f4fbb516af71d518a133f2c9ea/pkg/driver/controller.go#L135-L155

It checks for `uid < 0` in both blocks.

**What you expected to happen?**

It should check for `gid < 0` in the second block.

joelthompson:
This was fixed in #1269

PR #1269, opened 2024-02-09
Fix check for gid being less than 0

ashrayjain:
We were checking the uid instead of the gid when we check the configured value against 0

Bug fix

Breaks the ability to specify just a gid without specifying a uid

**What testing is done?**


odinuge:
Good catch, thanks!

/lgtm
/ok-to-test

joelthompson:
FYI, this fixes #1091 (which was left open).

PR #1281, opened 2024-02-29
Templatizing priotityClassName, hostNetwork, and additionalLabels in node-daemonset.yaml

RadOctocode:
**Is this a bug fix or adding new feature?**
This is adding some new features to the aws-efs-csi-driver helm charts

**What is this PR about? / Why do we need it?**
The PR is about adding changes to the aes-efs-csi-driver helm chart. 
- In `node-daemonset.yaml`
  - templatize priorityClassName
  - templatize hostNetwork
  - add additional labels to node-daemonset.yaml

**What testing is done?** 
yes

Issue #1270, opened 2024-02-13
Restarting the efs-csi-node Pod will cause mounts to hang on v1.6.0


RyanStan:
/kind bug

### Issue discovered with v1.6.0 of the aws-efs-csi-driver

When the EFS client mounts a file system, we re-direct a local NFS mount from the Linux Kernel to localhost, and then use a proxy process, stunnel, to receive the NFS traffic and forward it to EFS. The stunnel process runs in the efs-csi-node Pods. 

Version v1.6.0 of the csi driver switched `hostNetwork=true` to `hostNetwork=false`.  This means that Pods in the efs-csi-node Daemonset will launch into a new network namespace whenever they are restarted. This causes an issue. Any time these Pods are restarted, stunnel will launch in a new network namespace, while the local NFS mount from the kernel to localhost remains in the previous network namespace. This causes the mount to hang because the localhost NFS mount will not be able to reach the stunnel process once the Pod has restarted. When mounts hang, they go into uninterruptible sleep.

The issue was resolved in v1.7.0 of the driver, where we reverted the `hostNetwork` change, and set `hostNetwork=true` again. Thus, this issue only affects customers that established mounts while using v1.6.0 of the csi driver.

### Work-arounds
Any attempts to upgrade or restart the v1.6.0 of the efs-csi-node Daemonset will result in EFS mounts on the node hanging. 

To work-around this issue, you can launch new EKS nodes into your cluster, and then deploy a new efs-csi-node Daemonset, with `hostNetwork=true`, that targets these new nodes using Kubernetes Selectors. A rolling migration of your application to these new Nodes will allow you to upgrade to a new aws-efs-csi-driver version while ensuring that your application doesn't experience any downtime due to hanging mounts.

This issue was [originally discovered here](https://github.com/kubernetes-sigs/aws-efs-csi-driver/issues/1127#issuecomment-1734933606), but I'm making this post to raise visibility.

poblahblahblah:
We are also seeing this issue when upgrading from 1.6.0 to 1.7.5

nkryption:
We are also seeing this issue when upgrading from 1.6.0 to 1.7.2, any resolution for this?

nkryption:
[...]

PR #1288, opened 2024-03-20
Revert "Templatizing priotityClassName, hostNetwork, and additionalLabels in node-daemonset.yaml"

seanzatzdev-amazon:
Reverts kubernetes-sigs/aws-efs-csi-driver#1281

Setting hostNetwork=False on the Node Daemonset will cause mounts to hang at upgrade time. See this Github issue:
https://github.com/kubernetes-sigs/aws-efs-csi-driver/issues/1270.

### DIFF
--- a/charts/aws-efs-csi-driver/templates/node-daemonset.yaml
+++ b/charts/aws-efs-csi-driver/templates/node-daemonset.yaml
@@ -21,9 +21,6 @@ spec:
         app: efs-csi-node
         app.kubernetes.io/name: {{ include "aws-efs-csi-driver.name" . }}
         app.kubernetes.io/instance: {{ .Release.Name }}
-        {{- with .Values.node.podLabels }}
-        {{ toYaml . | nindent 8 }}
-        {{- end }}
       {{- if .Values.node.podAnnotations }}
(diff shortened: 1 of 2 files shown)


aaroniscode:
@seanzatzdev-amazon I believe the issue referenced is only about changing `hostNetwork` from `true` to `false` or vice-versa.  Creating a brand new install using `hostNetwork: false` should be fine.  The original PR doesn't change the default so I'm not sure it creates an issue?

levinedaniel:
Echoing @aaroniscode, we are actively using aws-efs-csi-driver with hostnet set to false. We _have_ seen the break when switching, but upon a clean install and actively upgrading using just hostnet=false we have been operating fully functionally.

Adityashar:
Hi @seanzatzdev-amazon 

We faced an incident due to this change of `hostNetwork: true` post the upgrade to `aws-efs-csi-driver-v3.0.6` helm chart. 
Could we please revert this field back to a configurable field so that the users have an option to manually choose between enabling or disabling this option?

```yaml
hostNetwork: {{ .Values.node.hostNetwork }}
```

PR #1323, opened 2024-04-24
Post-release PR for v2.0.1 and also changed the Default image from do…

mskanth972:
…ckerhub to Public ECR

Post-release PR for v2.0.1 and also changed the Default image from dockerhub to Public ECR

**What testing is done?**

### DIFF
--- a/charts/aws-efs-csi-driver/values.yaml
+++ b/charts/aws-efs-csi-driver/values.yaml
@@ -11,7 +11,7 @@ useFIPS: false
 
 image:
   repository: amazon/aws-efs-csi-driver
-  tag: "v2.0.0"
+  tag: "v2.0.1"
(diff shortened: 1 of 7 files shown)


seanzatzdev-amazon (approved):

charts/aws-efs-csi-driver/values.yaml
```diff
@@ -11,7 +11,7 @@ useFIPS: false
 
 image:
   repository: amazon/aws-efs-csi-driver
```
calvinbui, line 13: **issue (blocking):** this needs to be changed to `public.ecr.aws/efs-csi-driver/amazon/aws-efs-csi-driver` as mentioned in [this issue](https://github.com/kubernetes-sigs/aws-efs-csi-driver/issues/1330)

PR #1350, opened 2024-05-22
update dynamic_provisioning README to include reuseAccessPoint

nemani:
looks like this was missed in the commit that added this param.


nemani:
Okay fixing this readme is not worth figuring out the CLA stuff.

PR #1254, opened 2024-01-25
chore: Bump aws-sdk to 1.50 for EKS Pod Identity support

askulkarni2:
**Is this a bug fix or adding new feature?**
Fixes #1257. Bumps up aws-sdk-go to v1.50.3.

**What is this PR about? / Why do we need it?**

AWS SDK [v.1.47.11](https://github.com/aws/aws-sdk-go/releases/tag/v1.47.11) updated credential provider logic support which is required by pods which need to use the newly launched [Amazon EKS Pod Identities](https://aws.amazon.com/about-aws/whats-new/2023/11/amazon-eks-pod-identity/).

Also see https://docs.aws.amazon.com/eks/latest/userguide/pod-id-minimum-sdk.html for further information.

**What testing is done?** 

`make` and `make test`.

PR #1381, opened 2024-06-19
Add Pod Identity documentation in README

arnavgup1:
Feature

Adds documentation for the Pod Identity feature used for customers that want to authenticate with the pod identity agent that is available through EKS Add-ons. 

PR for the pod identity change: https://github.com/kubernetes-sigs/aws-efs-csi-driver/pull/1254


seanzatzdev-amazon (approved):

kkopachev:
I have a question about the language added by this PR. 
It seems that adding `AmazonEFSCSIDriverPolicy` policy to the cluster's node group's IAM role is essentially the same as third method using instance profile policy:
If my node has  `AmazonEFSCSIDriverPolicy` on it, then any pod on that node can use it, regardless of EKS Pod Identity presence. 
Am I missing something here?

leondkr:
Hi @kkopachev 

You can check my [PR#1422](https://github.com/kubernetes-sigs/aws-efs-csi-driver/pull/1422/files) here for further explanation of using EKS Pod Identities feature with EFS CSI.

Basically, it is only required to be added with the IAM role associated with EFS CSI controller.

PR #1401, opened 2024-07-04
Helm chart configurable model for cross account role

ssbostan:
Adding a new feature to the helm chart. Non-breaking.

To allow the user to dynamically manage secret names for cross account or disable it.

**What testing is done?** 
- [x] Deployment to the platform.
- [x] Deployment with the same behaviour before this change.
- [x] Deployment with customized secret names.
- [x] Deployment with allowing access to all secrets in the same namespace.


ssbostan:
@nckturner Any ETA on this?

PR #1434, opened 2024-08-09
Support node.kubeletPath helm chart value

neoaggelos:

Bug fix

Support `aws-efs-csi-driver` in environments where kubelet root dir is not `/var/lib/kubelet`.

It closely mimics `aws-ebs-csi-driver`:
- https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/charts/aws-ebs-csi-driver/templates/_node.tpl#L219-L230
- https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/charts/aws-ebs-csi-driver/values.yaml#L340

The default value remains `/var/lib/kubelet`, so backwards-compatibility would not be an issue.

**What testing is done?** 

Manual testing

### DIFF
--- a/charts/aws-efs-csi-driver/templates/node-daemonset.yaml
+++ b/charts/aws-efs-csi-driver/templates/node-daemonset.yaml
@@ -96,7 +96,7 @@ spec:
             {{- end }}
           volumeMounts:
             - name: kubelet-dir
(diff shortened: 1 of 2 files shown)


neoaggelos:
Thank you @mskanth972 @torredil, should be OK now

PR #1287, opened 2024-03-15
make sure the startup taint will eventually being removed after efs driver ready

abbshr:
**Is this a bug fix or adding new feature?**

bug fix

**What is this PR about? / Why do we need it?**

`removeNotReadyTaint()` just call one time, so there may be accident the taint didn't remove successfully,
lead to node unschedulable for the pod can't tolerate the taint, this PR make sure the startup taint will eventually being removed after efs driver ready.

**What testing is done?** 
test the compensation mechanism

PR #1438, opened 2024-08-13
fix: don't ignore kubernetes API access failure during node taint removal

steved:

Bug fix

Prior to #1287, it made sense to not fail initialization if the driver could not talk to the K8s API. Now that it's being retried, it shouldn't silently fail if the K8s API isn't accessible.

**What testing is done?**


steved:
bump - @mskanth972 any thoughts?

PR #1469, opened 2024-10-10
feat: add portRangeUpperBound to be exposed in the helm chart

cescribanohs:
We have a large amount of pvc deploys in our cluster and often we reach the error `Could not mount` pvc due to available port

Even the port_range_upper_bound property has been increased recently https://github.com/kubernetes-sigs/aws-efs-csi-driver/blob/b0af6bc4965ec01c86cdc3792c4207f24197d65e/pkg/driver/efs_watch_dog.go#L74 still we need this number to be configurable as FIPS it is at the moment.

For this reason, adding a new helm value that can instantiated from a `kustomize.config.k8s.io/v1beta1` component it is our goal.

**What testing is done?** 

We have deployed in our dev environment through argoCD

### DIFF
--- a/pkg/driver/efs_watch_dog.go
+++ b/pkg/driver/efs_watch_dog.go
@@ -20,6 +20,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"strconv"
 	"sync"
 	"text/template"
 
@@ -73,7 +74,7 @@ fips_mode_enabled = {{.FipsEnabled -}}
 
 # Define the port range that the TLS tunnel will choose from
 port_range_lower_bound = 20049
-port_range_upper_bound = 21049
+port_range_upper_bound = {{.PortRangeUpperBound}}
 
 # Optimize read_ahead_kb for Linux 5.4+
 optimize_readahead = true
@@ -179,9 +180,10 @@ type execWatchdog struct {
 }
 
 type efsUtilsConfig struct {
-	EfsClientSource string
-	Region          string
-	FipsEnabled     string
+	EfsClientSource     string
+	Region              string
+	FipsEnabled         string
+	PortRangeUpperBound string
 }
 
 func newExecWatchdog(efsUtilsCfgPath, efsUtilsStaticFilesPath, cmd string, arg ...string) Watchdog {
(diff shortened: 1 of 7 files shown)


pkg/driver/efs_watch_dog.go
```diff
@@ -284,7 +285,11 @@ func (w *execWatchdog) updateConfig(efsClientSource string) error {
 	// used on Fargate, IMDS queries suffice otherwise
 	region := os.Getenv("AWS_DEFAULT_REGION")
 	fipsEnabled := os.Getenv("FIPS_ENABLED")
-	efsCfg := efsUtilsConfig{EfsClientSource: efsClientSource, Region: region, FipsEnabled: fipsEnabled}
+	portRangeUpperBound, found := os.LookupEnv("PORT_RANGE_UPPER_BOUND")
```
avanish23, line 288: You don't have to check the condition here, you can assign the portRangeUpperBound as you have already added it under env in the deployment helm charts.
`portRangeUpperBound := os. Getenv("PORT_RANGE_UPPER_BOUND")`

Also, one check should be added to ensure that portRangeUpperBound is always greater than portRangeLowerBound(20049).

avanish23 (commented):
Add check to ensure `PORT_RANGE_UPPER_BOUND` is greater than 20049.

Mostly the PR looks good.
Thanks!

pkg/driver/efs_watch_dog.go
cescribanohs: Let me know if there is something else or just waiting you resolve the conversation and provide the approval.

avanish23: Tagging @mskanth972 to help with approval.
Thanks!

cescribanohs: Hi all, when do you think it can be approved? Many thanks

cescribanohs:
> /ok-to-test

All test are passing right now. Ty

cescribanohs:
Hi, any approvals or comments are welcomed. Thank you

mskanth972:
Don't we need to update [node-daemonset](https://github.com/kubernetes-sigs/aws-efs-csi-driver/blob/master/deploy/kubernetes/base/node-daemonset.yaml) yaml file?
I did not test this but is it necessary to update both Controller and Daemonset files to include the `PORT_RANGE_UPPER_BOUND` and also did you get chance to test this?

cescribanohs:
> Don't we need to update [node-daemonset](https://github.com/kubernetes-sigs/aws-efs-csi-driver/blob/master/deploy/kubernetes/base/node-daemonset.yaml) yaml file? I did not test this but is it necessary to update both Controller and Daemonset files to include the `PORT_RANGE_UPPER_BOUND` and also did you get chance to test this?

I have updated both daemonset and deployment but in this case they will pick the default value https://github.com/kubernetes-sigs/aws-efs-csi-driver/pull/1469/files#diff-5f5292175c0b29881ee742f0daf838980f771e03b22525cc574b4763b749a0ffR291 We have pushed this changes mapping a new image instead of repo: https://kubernetes-sigs.github.io/aws-efs-csi-driver/ in our dev cluster and working ok. Do you have in mind any other tests to be performed? Thank you in advance

cescribanohs:
Any additional comments? Can we approve this please? Thank you in advance

mskanth972:
can you please squash the commits?

cescribanohs:
> can you please squash the commits?

Done. Thank you

PR #1492, opened 2024-11-01
Discover EFS FileSystemId instead of specifying it in storage class

### LINKED ISSUE #1052: Discover EFS `fileSystemId` / `volumeHandle` instead of specifying it
I create EFS volumes in terraform, grab the ID, and then add it to the values file to use the EFS

I'd like to add a set of unique tags on my EFS and then specify a way of discovering the EFS ID without having to hard code the EFS ID

```diff
# specs/pv.yaml using static
-    volumeHandle: fs-582a03f3
+    volumeHandleDiscovery:
+      tags:
+        service: titan
+        env: dev
```

```diff
# specs/sc.yaml using dynamic
-  fileSystemId: fs-92107410
+  fileSystemDiscovery:
+    tags:
+      service: titan
+      env: dev
```

#### 1. terraform

Create the entire k8s release in terraform so the EFS is created and passed in to a `helm_resource` so we do not have to hard code this value.

#### 2. `initContainer` ?

As a workaround, is it possible to use some kind of `initContainer` using the `aws-cli` container to retrieve the efs ID on-the-fly, save it as an env var, and make it accessible to the pv/pvc/sc ?

**Additional context**
- There is [SSL discovery in the aws-load-balancer-controlelr](https://kubernetes-sigs.github.io/aws-load-balancer-controller/v2.5/guide/ingress/cert_discovery/) so why not EFS discovery in the aws-efs-csi-driver
- https://docs.aws.amazon.com/eks/latest/userguide/efs-csi.html

avanish23:
This PR introduces the feature of discovering EFS FileSystemId instead of having to specify it in the SC parameters.
The way to use this feature is to create a storage class where the parameters have a new field `fileSystemDiscovery`.
Currently, it takes 2 fields `creationToken` and` tags` both being optional. 
creationToken is a string that takes the creation token of the EFS to be discovered as input
tags is a map of strings
Sample Storage Class
```
kind: StorageClass
apiVersion: storage.k8s.io/v1
metadata:
  name: efs-sc
provisioner: efs.csi.aws.com
parameters:
  provisioningMode: efs-ap
  directoryPerms: "700"
  subPathPattern: "${.PVC.namespace}/${.PVC.name}" # optional
  ensureUniqueDirectory: "true" # optional
  reuseAccessPoint: "false" # optional
  fileSystemDiscovery: |
    creationToken: aws-efs-for-discovery
    tags:
      env: prod
      owner: avanishpatil23@gmail.com
```
Note: 
- use EFS Discovery with `|` as it is a multi-line string input 
- fileSystemId will continue to be supported as a parameter
- if more than one EFS is discovered, the driver will throw an error; likewise, if no EFS is discovered
- the EFS discovery results are not cached

**What testing is done?** 
Unit tests added and current unit tests pass

Closes #1052


nitrocode:
Thank you @avanish23 for the pr to close my issue.

Who is the right person to review this pr?

cc @mskanth972 @dankova22

Issue #517, opened 2021-07-17
Failed to create access point: AccessPointLimitExceeded: You have reached the maximum number of access points


chandrab-on:
/kind bug

**What happened?**
aws-efs-csi-driver throws and error saying that "Failed to create access point: AccessPointLimitExceeded: You have reached the maximum number of access points" after migrating to aws-efs-csi-driver version to v1.3.2 

Access Points are not being deleted all PVCs in the namespace are deleted

**What you expected to happen?**
Access Points needs ro deleted when corresponding PVC is deleted

**How to reproduce it (as minimally and precisely as possible)?**
Install the driver using helm:
helm repo add aws-efs-csi-driver https://kubernetes-sigs.github.io/aws-efs-csi-driver/
helm repo update
helm upgrade --install aws-efs-csi-driver --namespace kube-system aws-efs-csi-driver/aws-efs-csi-driver

**Anything else we need to know?**:

**Environment**
AWS EKS
- Kubernetes version (use `kubectl version`):
 v1.20.4
- Driver version:
 v1.3.2

chandrab-on:
{
  RespMetadata: {
    StatusCode: 403,
    RequestID: "3a4f7d4e-4132-4a77-b41c-993152b3c0f5"
  },
  ErrorCode: "AccessPointLimitExceeded",
  Message_: "You have reached the maximum number of access points (120) for your file system fs-4acdf681. Delete an access point and add a new one."
}
  Warning  ProvisioningFailed  14m  efs.csi.aws.com_ip-10-128-93-166.eu-west-1.compute.internal_b2ab1b88-ab90-49ee-a354-adaba28e8170  failed to provision volume with StorageClass "aws-efs": rpc err
or: code = Internal desc = Failed to create Access point in File System fs-4acdf681 : Failed to create access point: AccessPointLimitExceeded: You have reached the maximum number of access points (
120) for your file system fs-4acdf681. Delete an access point and add a new one.
{
  RespMetadata: {
    StatusCode: 403,
    RequestID: "b8ff1a28-bfe5-46d5-9b36-bee6e25528ad"
  },
  ErrorCode: "AccessPointLimitExceeded",
  Message_: "You have reached the maximum number of access points (120) for your file system fs-4acdf681. Delete an access point and add a new one."
}
[...]

PR #1497, opened 2024-11-08
Add "efs-dir" provsioning mode

### LINKED ISSUE #538: Disable Access Point Usage
We've run into an issue with the Dynamic Provisioning of the EFS-CSI-Driver due to hitting the limit of the number of access points one EFS can have. Using EFS for EKS doesn't necessarily require using Access Points to persist data right? I think it severely limits the scalability aspect of using EFS in EKS for certain situations with this driver because it would limit the number of PVCs you can provision with it to 120. Any more than 120 then the user would have to create a new EFS with a new storageclass, which is problematic if they are trying to save money using a long term solution and it'll require manual intervention.

Would it be possible to have an option to disable the Access Points if they are not needed by the user to lift this limit?

Have an option for the StorageClass in the `provisioningMode` section to have a different type for non-access point usage. Maybe call it "nfs" or something else.

mpatlasov:
This PR was inspired by [Jonathan Rainer](https://github.com/jonathanrainer)'s [PR](https://github.com/kubernetes-sigs/aws-efs-csi-driver/pull/732), and [Fabio Bertinatto](https://github.com/bertinatto)'s [PR](https://github.com/kubernetes-sigs/aws-efs-csi-driver/pull/1408).

New feature, requested in https://github.com/kubernetes-sigs/aws-efs-csi-driver/issues/538 and https://github.com/kubernetes-sigs/aws-efs-csi-driver/issues/517 and possibly other issues, this comes up a lot as a feature people would like.

Since its creation the driver has supported Access Point Provisioning as its main means of dynamic provisioning. However this is problematic for a few reasons, it can cause issues with deleting access-points (as reported [here](https://github.com/kubernetes-sigs/aws-efs-csi-driver/issues/517)) and it requires various AWS IAM Permissions that can be complicated to sort out and manage. Customers want a way to set their own permissions from the top level directory, but the mode doesn't allow to do that.

This PR allows a new provisioning mode called efs-dir so that instead of creating EFS access points, directories are created instead. This is achieved by creating a new Interface called Provisioner that is implemented by an AccessPointProvisioner (the original method) and a DirectoryProvisioner (the new method) this also allows in future for different kinds of provisioning to occur, maybe FileSystemProvisioning for example.

**What testing is done?** 
New code is covered by unit-tests, they all pass when I run "go test" locally. New e2e tests are added covering create/delete operation in "efs-dir" mode. They also pass when I run them against OpenShift (OCP) cluster with new driver installed.

fixes https://github.com/kubernetes-sigs/aws-efs-csi-driver/issues/538


seanzatzdev-amazon:
Please rebase

mpatlasov:
Update: last force-push rebased PR (only).

mpatlasov:
Update: last force-push fixed e2e tests for "efs-dir".

dankova22 (commented):
@mpatlasov  Thanks for the PR, we are considering supporting this feature. If could you address these comments and rebase, that would be great

mpatlasov:
Hi @dankova22 , thank you for reviewing this PR, highly appreciated! I've just rebased it, and not addressed your comments yet. Will do it soon.

mpatlasov:
> Hi @mpatlasov Im interested to hear more about one of your concerns with current access point provisioning:
> 
> > it can causes issues with user permissions around deleting provisioned directories
> 
> Is this referring to issues you have faced deleting access point root dirs outside of k8 or when deleting volumes with `deleteAccessPointRootDir`?

This phrase in PR's description is copy/pasta from [Jonathan Rainer](https://github.com/jonathanrainer)'s [PR](https://github.com/kubernetes-sigs/aws-efs-csi-driver/pull/732) description. I thought it referred the issue https://github.com/kubernetes-sigs/aws-efs-csi-driver/issues/517:

> What happened?
> Access Points are not being deleted all PVCs in the namespace are deleted
> What you expected to happen?
> Access Points needs ro deleted when corresponding PVC is deleted

I have to remove "user permissions around deleting provisioned directories" from the description, sorry for that confusion.

mpatlasov:
Hi @dankova22 , all comments addressed, PTAL please!

docs/README.md
```diff
@@ -25,22 +25,27 @@ The following CSI interfaces are implemented:
...
+| reuseAccessPoint      |                | false           | true     | When set to true, it creates the Access Point client-token from the provided PVC name. So that the AccessPoint can be replicated from a different cluster if same PVC name and storageclass configuration are used. Used only for provisioningMode "efs-ap".                                                                                                                                                                            |
 
 **Note**
+* We suggest the following settings for each provisioning mode:
+  * `efs-ap` -> `directoryPerms: 770`, with `gidRangeStart` and `gidRangeEnd` set to sensible values
+  * `efs-dir` -> `directoryPerms: 777` and omit `gidRangeStart` and `gidRangeEnd`
+  * This ensures that important system utilities will still be to access to the directories created
+  * All parameters are available in all modes but please think very carefully about what you're doing otherwise directories could be created that no-one can read or write to.
```
dankova22, line 48: I dont think we should recommend setting permissive 777 permissions (although this is the easiest to manage and users may choose to use this).

I think the recommendation should be for users to either:

1. Set a static gid in the storage class, and add this is as a supplementary group for all pods that will use this storage class.

2. Alternatively, users could use fsGroup to have kubernetes handle ownership and pod access to provisioned dirs. Key points would be:

- Need to set [fsGroup in securityContext for pod ](https://kubernetes-csi.github.io/docs/support-fsgroup.html)
- Need to ensure [fsGroupPolicy in CSI Driver](https://kubernetes-csi.github.io/docs/support-fsgroup.html) spec is set to mode that supports this (File or ReadWriteOnceWithFSType). This will allow kubernetes to change the ownership of the provisioned dir.
- For RWX cases if doing this, users should be extra careful to:
	  - Ensure all pods using same volume run with the same fsGroup, otherwise, one of their pods may be locked out as ownership will change.
	  - In Pod security context, Set [fsGroupChangePolicy as "OnRootMismatch"](https://kubernetes.io/docs/tasks/configure-pod-container/security-context/#configure-volume-permission-and-ownership-change-policy-for-pods) to avoid expensive recursive chowns of provisioned dir. 

We should have some documentation about this in the efs-dir example README, that we can reference here.

pkg/driver/provisioner_dir.go
```diff
@@ -0,0 +1,229 @@
...
+	if err != nil {
+		klog.V(5).Infof("Could not load file info for '%s'", provisionedDirectory)
+	}
+	klog.V(5).Infof("Permissions of folder '%s' are '%s'", provisionedDirectory, actualPerms)
+
+	err = d.mounter.Unmount(target)
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "Could not unmount %q: %v", target, err)
+	}
+	err = d.osClient.RemoveAll(target)
```
dankova22, line 124: We should use Remove here too instead of RemoveAll

consider the case where thread A mounts the root of the filesystem at /var/lib/csi/pv/1234

Thread B executes removeAll /var/lib/csi/pv/1234. We would delete the whole fs. 

The chance of uuid collision is pretty much zero, but better to use Remove to limit blast radius in case of a bug as the target should be empty at this point anyway

pkg/driver/provisioner_dir.go
```diff
@@ -0,0 +1,229 @@
...
+	}
+
+	defer func() {
+		// Try and unmount the directory
+		klog.V(5).Infof("Unmounting directory mounted at '%s'", target)
+		unmountErr := d.mounter.Unmount(target)
+		// If that fails then track the error but don't do anything else
+		if unmountErr != nil {
+			klog.V(5).Infof("Unmount failed at '%s'", target)
+			e = status.Errorf(codes.Internal, "Could not unmount %q: %v", target, err)
```
dankova22, line 184: In this case we never log or return the unmountErr. I think we prob want to do 

e = status.Errorf(codes.Internal, "Could not unmount %q: %v", target, unmountErr)

We should also make sure we log the errors from Mount and RemoveAll, so that we would have visibility in the case when both cleanup and core functionality fails.

test/e2e/e2e.go
```diff
@@ -416,6 +418,93 @@ var _ = ginkgo.Describe("[efs-csi] EFS CSI", func() {
 			}
 		})
 
+		createProvisionedDirectory := func(f *framework.Framework, basePath string, pvcName string) (*v1.PersistentVolumeClaim, *storagev1.StorageClass) {
```
dankova22, line 421: We need to add e2e test coverage of a actually writing to the fs. I think a test similar to "should successfully perform dynamic provisioning" for "efs-dir" is sufficient.

dankova22 (commented):
Thank you for the continued effort on this and addressing the previous comments.

I left a couple more thoughts after reviewing again. Please let me know if you are able to get to that or I will try to find some time to address myself.

mpatlasov:
Thank you for reviewing again, @dankova22 ! Can you please compose a patch for [docs/README.md](https://github.com/kubernetes-sigs/aws-efs-csi-driver/pull/1497#discussion_r1938646754), and I'll work on addressing other comments meanwhile.

mpatlasov:
Hi @dankova22 , all comments (excepting one about  [docs/README.md](https://github.com/kubernetes-sigs/aws-efs-csi-driver/pull/1497#discussion_r1938646754)) are addressed. Let me know if you can fix docs/README.md.

dankova22:
Thanks again! Here is a patch for the documentation update:
[update-readme-directory-provisioning.patch](https://github.com/user-attachments/files/18916700/update-readme-directory-provisioning.patch)

mpatlasov:
Hi @dankova22 , thanks for the doc patch, I've applied it now (sorry for delay - I was afk for last week).

mpatlasov:
Update: last force-push rebased PR (only).

mpatlasov:
`pull-aws-efs-csi-driver-external-test-eks` and `pull-aws-efs-csi-driver-e2e` are perma-failing:
```
ERROR: failed to solve: process "/bin/sh -c clean_install amazon-efs-utils true &&     clean_install crypto-policies true &&     install_binary         /usr/bin/cat         /usr/bin/cd         /usr/bin/df         /usr/bin/env         /usr/bin/find         /usr/bin/grep         /usr/bin/ln         /usr/bin/ls         /usr/bin/mount         /usr/bin/umount         /sbin/mount.nfs4         /usr/bin/openssl         /usr/bin/sed         /usr/bin/stat         /usr/bin/stunnel         /usr/sbin/tcpdump         /usr/bin/which &&     cleanup \"efs-csi\"" did not complete successfully: exit code: 1
make[1]: *** [Makefile:86: .image-30814-linux-amd64-amazon] Error 1
make[1]: Leaving directory '/home/prow/go/src/github.com/kubernetes-sigs/aws-efs-csi-driver'
make: *** [Makefile:119: test-e2e] Error 2
```
This is not relevant to PR.

elmiko:
we are discussing this issue at the sig cloud provider meeting today, and although we are listed as the owning sig we are wondering if this is more appropriate for sig storage?

cc @gnufied @jsafrane @msau42 for more comment, should this be owned by sig storage?

also, @kmala is going to ask around within aws about possible contacts.

PR #1509, opened 2024-11-25
Minor change to align with typical helm chart conventions

rayandabbagh:

Minor change to align with typical helm chart conventions

Minor change to align with typical helm chart conventions

**What testing is done?** 

Confirmed that the additional argument appears when I describe the csi controller pod


charts/aws-efs-csi-driver/templates/controller-deployment.yaml
```diff
@@ -136,11 +136,9 @@ spec:
             {{- if hasKey .Values.controller "leaderElectionLeaseDuration" }}
             - --leader-election-lease-duration={{ .Values.controller.leaderElectionLeaseDuration }}
             {{- end }}
-            {{- if .Values.sidecars.csiProvisioner.additionalArgs }}
```
rayandabbagh, line 139: Minor change here, "csiProvisioner" instead of "provisioner" (fixed)

PR #1550, opened 2024-12-20
Post release PR for v2.1.3

dankova22:
post release for 2.1.3


charts/aws-efs-csi-driver/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: aws-efs-csi-driver
-version: 3.1.3
+version: 3.1.4
 appVersion: 2.1.2
```
mskanth972, line 4: Can you bump this also, like below
`appVersion:2.1.3`

PR #1554, opened 2025-01-06
docs: AAdd information about the incorporation of efs-utils v2.X.X in EFS CSI driver version 2.X.X.

mskanth972:

Add information about the incorporation of efs-utils v2.X.X in EFS CSI driver version 2.X.X. Highlight the replacement of stunnel with efs-proxy for TLS encryption and note the need to re-mount existing file systems for performance benefits.

**What testing is done?**


jrakas-dev:
LGTM

PR #1569, opened 2025-01-10
Upgrade python distro and create symlink to stunnel5

dankova22:
Upgrade python distros. Add symlink for stunnel5 -> stunnel to keep backwards compatibility for upgrade process where binary is expected at /usr/bin/stunnel5.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -28,7 +28,7 @@ ENV EFS_CLIENT_SOURCE=$client_source
 
 RUN OS=${TARGETOS} ARCH=${TARGETARCH} make $TARGETOS/$TARGETARCH
 
-FROM public.ecr.aws/eks-distro-build-tooling/python:3.9.14-gcc-al2 as rpm-provider
+FROM public.ecr.aws/eks-distro-build-tooling/python:3.11-gcc-al23 as rpm-provider
 
 # Install efs-utils from github by default. It can be overriden to `yum` with --build-arg when building the Docker image.
 # If value of `EFSUTILSSOURCE` build arg is overriden with `yum`, docker will install efs-utils from Amazon Linux 2's yum repo.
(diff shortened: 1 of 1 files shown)


andrewjamesbrown:
@dankova22 is this PR/fix somehow related to [this?](https://github.com/aws/efs-utils/blob/master/test/watchdog_test/test_start_tls_tunnel.py#L110-L114)

andrewjamesbrown:
>Upgrade python distros. Add symlink for stunnel5 -> stunnel to keep backwards compatibility for upgrade process where binary is expected at /usr/bin/stunnel5.

For my own info, what upgrade processes depend on `/usr/bin/stunnel5` versus `/usr/bin/stunnel`?

andrewjamesbrown:
@dankova22 https://github.com/kubernetes-sigs/aws-efs-csi-driver/tree/master/docs#important-notes-on-efs-csi-driver-v2-upgrade

mskanth972:
@andrewjamesbrown, when there are existing mounts using `Stunnel5` binary and when we upgrade to EFS CSI Driver version which is using `Stunnel`, the mounts are breaking due to binary configuration change and pods are unable to read/write to the existing application.

mskanth972:
> It looks like the efs-utils code should be using [this function](https://github.com/aws/efs-utils/blob/master/src/mount_efs/__init__.py#L1324-L1329) to find the appropriate stunnel binary (which is different on AL2)

Looks like driver mentions `stunnel` explicitly in DockerFile, Unless we create a symlink (ln -s /usr/bin/stunnel /usr/bin/stunnel5), the mount process will fail.

andrewjamesbrown:
@mskanth972 who knew that upgrading from AL2 to AL2023 would be so challenging :-)

mugdha-adhav:
After this change was merged, we started seeing mount issues again.

This maybe related to https://github.com/kubernetes-sigs/aws-efs-csi-driver/issues/1571.

Why did we have to revert back to installing stunnel instead of stunnel5? Would it not have worked if we kept stunnel5 and created a symlink for stunnel instead?

dankova22:
@mugdha-adhav  Are these the same errors you have seen with v2.1.3?

Could you please open a issue and provide some logs with the errors?

PR #1620, opened 2025-05-02
Return existing access point if one already exists during create workflow

jrakas-dev:
Bug fix. This addresses an issue that arises when access points aren't cleanly deleted when the CSI controller crashes, resulting in `AccessPointAlreadyExists` errors being returned when volume creation is retried and preventing the PVC to be re-created.

This PR modifies the controller to adopt the existing access point in the event that one already happens to exist for the same creation token, allowing the PVC creation to succeed.

**What testing is done?** 
Unit tests were added, verified fix against a reproduction of the issue


pkg/cloud/cloud.go
```diff
@@ -201,10 +201,21 @@ func (c *cloud) CreateAccessPoint(ctx context.Context, clientToken string, acces
 		o.Retryer = c.rm.createAccessPointRetryer
 	})
 	if err != nil {
-		if isAccessDenied(err) {
+		if isAccessPointAlreadyExists(err) {
+			klog.V(4).Infof("Access point already exists for client token %s. Retrieving existing access point details.", clientToken)
+			existingAccessPoint, err := c.FindAccessPointByClientToken(ctx, clientToken, *createAPInput.FileSystemId)
```
dankova22, line 206: Should we validate that the access point returned matches exactly the the one we are trying to create? Just as an additional safeguard against FindAccessPointByClientToken returning an access point that doesnt exactly match the one we are trying to create. We could have a wrapper around FindAccessPointByClientToken?

pkg/cloud/cloud.go
```diff
@@ -201,10 +201,21 @@ func (c *cloud) CreateAccessPoint(ctx context.Context, clientToken string, acces
...
+			if err != nil {
+				return nil, fmt.Errorf("Error attempting to retrieve existing access point: %v", err)
+			}
+			if existingAccessPoint == nil {
+				return nil, fmt.Errorf("No access point for client token %s was returned", clientToken)
+			}
+			return existingAccessPoint, nil
```
wangnyue, line 213: can we also verify the Uid, Gid and permission before re-using the existing access point ? I feel there could be a case that we return an access point with wrong permission.

pkg/cloud/cloud.go
wangnyue: +1

PR #1629, opened 2025-05-12
Feat/support az multi parameter

Mikey032:
New feature

Currently, cross-account mounts with dynamic provisioning and without DNS setup, only support mounting a single availability zone, either by providing a specific AZ or by not specifying any and receiving a random one. The AZ multi parameter returns a map of availability zone IDs with their respective mount target IPs to the PV. When combined with the node CSI driver querying the EC2 instance metadata endpoint for its own availability zone, allows it to automatically selects the mount target IP for a pod in the same availability zone.
When used with a multi-AZ EKS deployment, this reduces cross-AZ traffic costs and increases resilience. This does not disrupt the current logic.

**What testing is done?** 
Unit tests have been updated, the logic has been verified on a cluster and EFS file system each in their own VPC and account attached to a transit gateway.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -63,7 +63,6 @@ COPY --from=rpm-provider /tmp/rpms/* /tmp/download/
 # cd, ls, cat, vim, tcpdump, are for debugging
 RUN clean_install amazon-efs-utils true && \
     clean_install crypto-policies true && \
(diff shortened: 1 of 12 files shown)


Mikey032:
Hi @dankova22, were you able to review this already?

PR #1636, opened 2025-05-22
fix mounting volumes with FIPS enabled

h0tw1r3:

Bug

The documentation suggests setting `useFIPS: true` is sufficient to ensure mounts are made using FIPS crypto, but it is not.

Using EKS cluster addons, passing the `useFIPS` configuration does nothing but ensure AWS API calls are bould by FIPS.

**What testing is done?** 

Working on it. Floating to get feedback.

### DIFF
--- a/charts/aws-efs-csi-driver/templates/controller-deployment.yaml
+++ b/charts/aws-efs-csi-driver/templates/controller-deployment.yaml
@@ -93,6 +93,8 @@ spec:
             {{- if .Values.useFIPS }}
             - name: AWS_USE_FIPS_ENDPOINT
               value: "true"
+            - name: FIPS_ENABLED
+              value: "true"
             {{- end }}
             - name: PORT_RANGE_UPPER_BOUND
               value: "{{ .Values.portRangeUpperBound }}"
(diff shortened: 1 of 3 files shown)


charts/aws-efs-csi-driver/templates/controller-deployment.yaml
```diff
@@ -93,6 +93,8 @@ spec:
             {{- if .Values.useFIPS }}
             - name: AWS_USE_FIPS_ENDPOINT
```
dankova22, line 94: The current implementation should work because efs-utils checks for FIPS in two ways:

-Through the AWS_USE_FIPS_ENDPOINT environment variable (which is currently set by the Helm chart)
 -Through the fips_mode_enabled setting in the config file

https://github.com/aws/efs-utils/blob/master/src/mount_efs/__init__.py#L3474-L3478
The current Helm chart sets AWS_USE_FIPS_ENDPOINT=true when useFIPS is enabled, which should be sufficient

However, this change does add redundancy for both ways

dankova22:
Updated the documentation for proper capitalization: https://github.com/kubernetes-sigs/aws-efs-csi-driver/pull/1644. We are still evaluating if using the same helm value to also update the config file makes sense.

charts/aws-efs-csi-driver/templates/controller-deployment.yaml
wangnyue: +1 . I think the real bug is in Readme. it need `--set useFIPS=true ` instead of `--set useFips=true`

PR #1647, opened 2025-06-16
CI Test

torredil:
This PR demonstrates CI failures in https://github.com/kubernetes-sigs/aws-efs-csi-driver/pull/1641 (for kOps) are caused by driver binary issues (and not test-infra).

As for the EKS Job, the root cause is cluster version, see https://docs.aws.amazon.com/eks/latest/userguide/kubernetes-versions.html. The existing Makefile references 1.25 which is not a supported Amazon EKS release.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -129,7 +129,7 @@ test-e2e:
 .PHONY: test-e2e-external-eks
 test-e2e-external-eks:
 	CLUSTER_TYPE=eksctl \
-	K8S_VERSION="1.25" \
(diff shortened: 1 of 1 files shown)


torredil:
Please note that end of standard support for EKS 1.27 is coming up (July 24). I strongly recommend upgrading to the latest version and implementing a mechanism to periodically upgrade dependencies to stay on top of tech debt. Otherwise, these sorts of issues will be recurring.

PR #1660, opened 2025-07-04
Update encryptInTransit examples

Phaow:
- Doc enhancement

- The `tls` under mountOptions is deprecated with this driver since tls is enabled by default(https://github.com/kubernetes-sigs/aws-efs-csi-driver/pull/205).  Remove the deprecated `tls` mountOption from example use cases.


Phaow:
Hi @dankova22 , could you help take a look when you get a chance? Thank you!
/assign @dankova22

Issue #1661, opened 2025-07-07
2.1.9 doesn't work with FIPS enabled EKS optimized Bottlerocket AMIs


krmichelos:
/kind bug

**What happened?**
After upgrading to 2.1.9 the driver failed to mount EFS PVs

**What you expected to happen?**
EFS PVs would continue to mount as they did with version 2.1.8

**How to reproduce it (as minimally and precisely as possible)?**
Run a pod that uses an EFS PV on a node based on the AMI listed below with version 2.1.9 of the EFS CSI driver plugin

**Anything else we need to know?**:
Pods in this state have events with errors like this:
`FATAL: Startup failure (dev note: apps_startup()) for openssl
003E8F1FD67F0000:error:12800067:DSO support routines:dlfcn_load:could not load the shared library:crypto/dso/dso_dlfcn.c:118:filename(/usr/lib64/ossl-modules/fips.so): /usr/lib64/ossl-modules/fips.so: cannot open shared object file: No such file or directory
003E8F1FD67F0000:error:12800067:DSO support routines:DSO_load:could not load the shared library:crypto/dso/dso_lib.c:147:
003E8F1FD67F0000:error:07880025:common libcrypto routines:provider_init:reason(37):crypto/provider_core.c:949:name=fips
003E8F1FD67F0000:error:0700006D:configuration file routines:module_run:module initialization error:crypto/conf/conf_mod.c:276:module=providers, value=provider_sect retcode=-1`

**Environment**
- Kubernetes version (use `kubectl version`): v1.32.5-eks-5d4a308
- Node AMI: ami-0a78cd6255b3985d6 (bottlerocket-aws-k8s-1.32-fips-x86_64-v1.42.0-5ed15786)
- Driver version: 2.1.9

**Please also attach debug logs to help us better diagnose**

- Instructions to gather debug logs can be found [here](https://github.com/kubernetes-sigs/aws-efs-csi-driver/tree/master/troubleshooting)

ghost:
I tried to upload the tgz but github was returning a 422.  A zip of the files seems to have worked.
[results.zip](https://github.com/user-attachments/files/21110207/results.zip)

dderrickmatheww:
This is happening with my custom RHEL 8 AMI EKS 1.29 as well.

hayden-miller-sp:
[...]

PR #1678, opened 2025-08-04
Revert openssl installation change in Dockerfile

anthotse:

Bug fix for https://github.com/kubernetes-sigs/aws-efs-csi-driver/issues/1661

Fix mount/openssl dependency failure on FIPS AMIs


Kazulin:
Is this going to be merged and when will we see Anthotse's fix, release version timeline wise? Thanks!

Issue #1689, opened 2025-08-25
Tags Names with `:` fail to parse


hojerst:
/kind bug

**What happened?**

when adding tags with a `:` in the key, this fails to parse correctly. Example:

| Tag Name | Value |
|---|---|
| mandatory:test | val |

This adds a tag `mandatory` of value `test` instead of `mandatory:test` of value `val`

**What you expected to happen?**

There should be a (documented) way to escape the `:` as it might be in the value OR the key and there is no easy way to distingush between those two scenarios.

**How to reproduce it (as minimally and precisely as possible)?**

run controller with `--tags=mandatory:test:val`

**Anything else we need to know?**:

**Environment**
- Kubernetes version (use `kubectl version`):
- Driver version: 1.2

**Please also attach debug logs to help us better diagnose**

- Instructions to gather debug logs can be found [here](https://github.com/kubernetes-sigs/aws-efs-csi-driver/tree/master/troubleshooting)

dankova22:
Addressed in #1693

PR #1693, opened 2025-09-02
Allow : in Tags with Escape Character

samuhale:
Adding support for `:` in tag names using `\` as an escape character

Issue #1689 

**What testing is done?** 
Added additional check to ensure tag parsing works with the escape character and without it, ran them with `make test`


samuhale:
Ran gofmt on files I changed, only changes are whitespace and removing an unneeded semicolon

PR #1702, opened 2025-09-11
feat: Allow setting allowVolumeExpansion to SC in helm chart

josh-ferrell:
New feature

Allows the `allowVolumeExpansion` key to be added to storage classes managed by the aws-efs-csi-driver helm chart

**What testing is done?** 
- Verified helm template rendered correctly using [example StorageClass](https://github.com/kubernetes-sigs/aws-efs-csi-driver/blob/f82d0dfb61811677c44f6701212ef3adf88feb3f/charts/aws-efs-csi-driver/values.yaml#L221-L238) found in current values file.
- Added `allowVolumeExpansion: true` to current example and verified helm template rendered correctly.

### DIFF
--- a/charts/aws-efs-csi-driver/templates/storageclass.yaml
+++ b/charts/aws-efs-csi-driver/templates/storageclass.yaml
@@ -8,6 +8,9 @@ metadata:
   {{ toYaml . | indent 4 }}
   {{- end }}
 provisioner: efs.csi.aws.com
+{{- with .allowVolumeExpansion }}
+allowVolumeExpansion: {{ . }}
+{{- end }}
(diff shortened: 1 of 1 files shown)


dankova22:
Hello @josh-ferrell thank you for raising the PR. Volume expansion with the EFS CSI Driver should not be necessary as EFS is elastic in nature and total capacity per volume is never bound. The EFS Driver [does not implement the volume expansion RPCs](https://github.com/kubernetes-sigs/aws-efs-csi-driver/blob/master/pkg/driver/node.go#L303-L305). Is there a different use case you have for this feature?

PR #1705, opened 2025-09-15
Add SA option for automounting tokens

bareksml:
Fix and quality of life imptovement (backport from aws-loadbalancer-controller helmchart)

Since modern k8s releases do not automount tokens one wishing to use IRSA must also create dedicated serviceaccounts outside of this helmchart. Considering the fact that values.yaml has hints covering irsa annotations lack of this parameter makes whole thing more than incoherent.

**What testing is done?** 
- template rendered and sa objects compared with sa generated by the aws loadbalancer controller and object definitions

### DIFF
--- a/charts/aws-efs-csi-driver/templates/controller-serviceaccount.yaml
+++ b/charts/aws-efs-csi-driver/templates/controller-serviceaccount.yaml
@@ -1,6 +1,7 @@
 {{- if .Values.controller.serviceAccount.create }}
 apiVersion: v1
 kind: ServiceAccount
+automountServiceAccountToken: {{ .Values.controller.serviceAccount.automountServiceAccountToken }}
 metadata:
   name: {{ .Values.controller.serviceAccount.name }}
   namespace: {{ .Release.Namespace }}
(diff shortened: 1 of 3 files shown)


charts/aws-efs-csi-driver/templates/controller-serviceaccount.yaml
```diff
@@ -1,6 +1,7 @@
 {{- if .Values.controller.serviceAccount.create }}
 apiVersion: v1
 kind: ServiceAccount
+automountServiceAccountToken: {{ .Values.controller.serviceAccount.automountServiceAccountToken }}
```
wangnyue, line 4: It would be better we do with check first like 

```
{{- if .Values.node.serviceAccount.automountServiceAccountToken }}
automountServiceAccountToken: {{ .Values.node.serviceAccount.automountServiceAccountToken }}
{{- end }}
```

charts/aws-efs-csi-driver/templates/node-serviceaccount.yaml
```diff
@@ -1,6 +1,7 @@
 {{- if .Values.node.serviceAccount.create }}
 apiVersion: v1
 kind: ServiceAccount
+automountServiceAccountToken: {{ .Values.node.serviceAccount.automountServiceAccountToken }}
```
wangnyue, line 4: It would be better we do with check first like 

```
{{- if .Values.node.serviceAccount.automountServiceAccountToken }}
automountServiceAccountToken: {{ .Values.node.serviceAccount.automountServiceAccountToken }}
{{- end }}
```

PR #1716, opened 2025-10-08
Support whitespace in tag keys and values

slackfan:

This PR adds a feature.

The current EFS Driver implementation does not support whitespace in tagging keys and values while the EFS API itself does support that. 

The proposed implementation adds escaping support for the whitespace, allowing to configure the tags accordingly so the Driver implementation can handle it and causes fancy potentially undesirable tag values being written in case this is not known. 

I chose to implement this via escape sequence. Namely a backslash as this was chosen recently via PR #1693 

Existing behavior is mainly kept. The only difference I propose to implement is that empty tag values are supported. Before the changes this raised an error, after the changes a missing colon separator leads to an empty value. 

**What testing is done?** 

Additional unit tests have been added.

### DIFF
--- a/pkg/driver/controller_test.go
+++ b/pkg/driver/controller_test.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"math/rand"
+	"reflect"
 	"regexp"
 	"strconv"
 	"sync"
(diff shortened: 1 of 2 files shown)


slackfan:
Also, for I chose not to modify The Helm charts helper function https://github.com/kubernetes-sigs/aws-efs-csi-driver/blob/master/charts/aws-efs-csi-driver/templates/_helpers.tpl#L50 As changing it would break existing behavior (existing workarounds in place). If it makes sense to add automatic escaping there, let me know.

slackfan:
What would be needed to move this forward? If something is missing, please let me know.

dankova22:
Hi, could you please rebase your changes if you would like the feature to be supported?

slackfan:
@dankova22, thanks for the heads-up, the PR is updated and tests are green.

DavidXU12345:
@slackfan Could you pleaes squash all the commits to a single one before we merge it? thanks

slackfan:
@DavidXU12345, as requested, everything is in a single commit now.

PR #1720, opened 2025-10-11
feat: add configurable health probes for controller and node components

vagharsh:
This PR adds configurable probe support for both controller and node components while maintaining backward compatibility.

## Changes
- Add configurable `livenessProbe`, `readinessProbe`, and `startupProbe` settings for controller
- Add configurable `livenessProbe`, `readinessProbe`, and `startupProbe` settings for node  
- Maintain backward compatibility by using existing hardcoded values as defaults
- Enable optional readiness and startup probes that can be configured independently
- Support all standard Kubernetes probe configuration options

## Backward Compatibility
- Existing deployments continue to work without any changes
- Default probe configurations use the same values that were previously hardcoded
- Readiness and startup probes are optional and only enabled when explicitly configured

## Testing
- Verified templates render correctly with default values
- Tested with custom probe configurations
- Confirmed backward compatibility with existing values


vagharsh:
hi @dankova22 @justinsb any update ?

charts/aws-efs-csi-driver/CHANGELOG.md
```diff
@@ -1,4 +1,10 @@
 # Helm chart
+# v3.2.5
```
samuhale, line 18: Changelog should only be updated during a release, please remove this from the PR

charts/aws-efs-csi-driver/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: aws-efs-csi-driver
-version: 3.2.4
+version: 3.2.5
```
samuhale: Same comment here, version should only be changed during a release. Please remove this diff

charts/aws-efs-csi-driver/templates/controller-deployment.yaml
```diff
@@ -109,14 +109,22 @@ spec:
...
-            httpGet:
-              path: /healthz
-              port: healthz
-            initialDelaySeconds: 10
-            timeoutSeconds: 3
-            periodSeconds: 10
-            failureThreshold: 5
+            {{- toYaml . | nindent 12 }}
+          {{- end }}
+          {{- if .Values.controller.readinessProbe }}
```
samuhale, line 122: Can this be condensed at all? Seems like we could make use of the with keyword, e.g. https://github.com/kubernetes-sigs/aws-efs-csi-driver/blob/master/charts/aws-efs-csi-driver/templates/controller-deployment.yaml#L58-L60

charts/aws-efs-csi-driver/templates/node-daemonset.yaml
```diff
@@ -120,14 +120,22 @@ spec:
...
+            {{- toYaml . | nindent 12 }}
+          {{- end }}
+          {{- if .Values.node.readinessProbe }}
+          {{- if not (empty .Values.node.readinessProbe) }}
+          readinessProbe:
+            {{- toYaml .Values.node.readinessProbe | nindent 12 }}
+          {{- end }}
+          {{- end }}
+          {{- if .Values.node.startupProbe }}
+          {{- if not (empty .Values.node.startupProbe) }}
```
samuhale: Same comment here, seems overly verbose with the if/else

samuhale:
Left a few small comments. Once those are resolved, I think this should be ready to merge

charts/aws-efs-csi-driver/templates/node-daemonset.yaml
vagharsh: fixed

charts/aws-efs-csi-driver/templates/controller-deployment.yaml
vagharsh: thanks for the ref, fixed

charts/aws-efs-csi-driver/Chart.yaml
vagharsh: removed

charts/aws-efs-csi-driver/CHANGELOG.md
vagharsh: removed

samuhale (approved):

PR #1739, opened 2025-11-13
Add opt-in parameter to disable node mutation permissions

dankova22:
feauture

Adds a parameter node.serviceAccount.disableMutation to disable mutating permissions on the node SA. Also updates the node code to detect and gracefully handle this case rather than attempting and failing to patch the node.

**What testing is done?** 

```
helm template test-release . > /tmp/default.yaml
grep -A 15 "kind: ClusterRole" /tmp/default.yaml | grep "efs-csi-node-role"
apiVersion: rbac.authorization.k8s.io/v1
metadata:
  name: efs-csi-node-role
  labels:
    app.kubernetes.io/name: aws-efs-csi-driver
rules:
  - apiGroups: [""]
    resources: ["nodes"]
    verbs: ["get", "list", "watch"]
  - apiGroups: [""]
    resources: ["nodes"]
    verbs: ["patch"]


helm template test-release . --set node.serviceAccount.disableMutation=true > /tmp/secure.yaml
grep -A 15 "kind: ClusterRole" /tmp/secure.yaml | grep  "efs-csi-node-role"
apiVersion: rbac.authorization.k8s.io/v1
metadata:
  name: efs-csi-node-role
  labels:
    app.kubernetes.io/name: aws-efs-csi-driver
rules:
  - apiGroups: [""]
    resources: ["nodes"]
    verbs: ["get", "list", "watch"]
---
```

### DIFF
--- a/charts/aws-efs-csi-driver/values.yaml
+++ b/charts/aws-efs-csi-driver/values.yaml
@@ -204,6 +204,10 @@ node:
     annotations: {}
     ## Enable if EKS IAM for SA is used
     #  eks.amazonaws.com/role-arn: arn:aws:iam::111122223333:role/efs-csi-role
+    # Disable mutating permissions for the node service account.
+    # When disableMutation is true, some features of the EFS CSI Driver node pods will not function, such as taint removal.
+    # Primarily useful in particularly security-sensitive environments, or on multi-tenant clusters that isolate tenants by node.
+    disableMutation: false
   healthPort: 9809
   # securityContext on the node pod
   securityContext:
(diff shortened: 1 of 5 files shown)


charts/aws-efs-csi-driver/values.yaml
```diff
@@ -204,6 +204,10 @@ node:
     annotations: {}
     ## Enable if EKS IAM for SA is used
     #  eks.amazonaws.com/role-arn: arn:aws:iam::111122223333:role/efs-csi-role
+    # Disable mutating permissions for the node service account.
+    # When enabled, some features of the EFS CSI Driver node pods will not function, such as taint removal.
```
DavidXU12345: "When enabled" is a bit confusing, can we say "When disableMutation is true"?

DavidXU12345:
Similar PR for EBS: https://github.com/kubernetes-sigs/aws-ebs-csi-driver/pull/2723

DavidXU12345 (approved):

PR #1750, opened 2025-11-24
Add ability to configure node container securityContext

prateekdegaons1991:
Add ability to pass node container security contexts

Since modern k8s platforms comes with security webhooks that mutates container securityContexts if they are not passed explicitly.
the existing node daemonSet templates consists hardcoded value for container security

```
containers:
        - name: efs-plugin
          securityContext:
            privileged: true
```
This PR allowes users to pass the container security contexts via the values.yaml file.

What testing is done?

template rendered and could able to see the values being reflected in the rendred template

### DIFF
--- a/charts/aws-efs-csi-driver/templates/node-daemonset.yaml
+++ b/charts/aws-efs-csi-driver/templates/node-daemonset.yaml
@@ -70,8 +70,10 @@ spec:
       {{- end }}
       containers:
         - name: efs-plugin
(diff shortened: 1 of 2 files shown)


prateekdegaons1991:
PTAL

DavidXU12345:
@prateekdegaons1991 could you clarify why (or in what specific case) csi-node would set `privileged` to false? csi-node is responsible to mount which requires privileged to true. Thanks

PR #1774, opened 2025-12-12
fix: verify startup taint removal after patch to prevent false positives

aliyajo:
Bug Fix

Adds verification to ensure startup taint removal succeeds. 

Previously, there were some false positives in the logs claiming the startup taint on nodes were removed. This change checks that if the startup taint persists, the existing retry mechanism is triggered. 

**What testing is done?** 
* Integration Testing
* Unit tests

### DIFF
--- a/pkg/driver/node.go
+++ b/pkg/driver/node.go
@@ -551,10 +551,19 @@ func removeNotReadyTaint(k8sClient cloud.KubernetesAPIClient) error {
 		return err
 	}
 
-	_, err = clientset.CoreV1().Nodes().Patch(context.Background(), nodeName, k8stypes.JSONPatchType, patch, metav1.PatchOptions{})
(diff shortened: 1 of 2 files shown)


DavidXU12345:
may I know what's the root cause of false positive? (i.e. no error but taints are still there)

aliyajo:
> may I know what's the root cause of false positive? (i.e. no error but taints are still there)

The false positive occurred because the original code assumed a successful `Patch()` API call (no error returned) meant the taint was removed, without verifying the actual node state. In rare cases during rapid Karpenter-driven scaling events, the taint could persist on the node even after `Patch()` succeeded, preventing the retry mechanism from activating. 
The theory for a root cause is race conditions from concurrent controllers (CSI drivers, Karpenter, node controller), though I was unable to reproduce this in a test environment to confirm the exact mechanism. This PR adds verification after the `Patch()` operation by re-fetching the node object and confirming the taint has been removed. If the taint persists, an error is returned to trigger the retry loop, ensuring the taint is eventually removed.

DavidXU12345 (approved):

PR #1800, opened 2026-02-17
Enable the Pod Disruption Budget by default

mamoit:
Fixes #1799

It's more of a default behaviour change.

Enable the pod disruption budget by default to avoid outages.
The default replica count is 2, so this avoids outages with the default configuration.

**What testing is done?** 
Yes


samuhale:
We think this change has a potential to be disruptive to other customers. Anyone needing this feature can always enable this option in your environment using the helm chart. Closing this as not planned

PR #1801, opened 2026-02-18
Closes #1609: Ipv4 private link in ipv6 network env

### LINKED ISSUE #1609: Idea Suggestion: Enable IPv4 preference in EFS CSI Driver in Ipv6 Environments

Opening this for possible discussions with the maintainers

- An interesting use-case in a dual stack VPC network (with EFS private link - which allowed the IPv4 traffic in scenario) and an IPv6 EKS cluster.
- the EFS volume ENI has an IPv4 and IPv6 address attached
- default connection and mount from EC2 workernode and kubernetes pod are using the EFS IPv6 address instead of the IPv4, which results in a mount timeout. EFS does not use IPv6
```
stunnel log:
2025-03-13T18:45:06.028Z 83833 WARN efs_proxy::controller Failed to establish an initial connection to EFS. Error: Connect attempt failed - Timeout

mount log:
2025-03-13 18:45:06 UTC - ERROR - Mounting fs-xxxxxxx.efs.us-gov-west-1.amazonaws.com to /mnt/efs failed due to timeout after 15 sec, mount attempt 1/3, wait 0 sec before next attempt.
```

- However adding a preference for IPv4 in the worker node, allows use of IPv4 over 6 and successful mount
```
mkdir /etc/gai.conf
echo "precedence ::ffff:0:0/96  100" >> /etc/gai.conf
```

- interestingly adding this same configuration in the linux environment of the EFS Driver csi-node pod allows successful mount into the pod in IPv6 as well

Is it possible to pack the default EFS container with a preference for IPv4 connection in dual-stack environment until IPv6 is also supported in EFS

There isnt an alternative with IPv6 use

abeowlu:
adds a feature

This PR address the issue reported in #1609 
in a dual stack environment with EFS service private link (VPC endpoint_ configured for connection, pod connection to EFS timeout because IPv6 source address is not allowed;
- NAT translation and other workaround fails because the EFS VPC endpoint does not allow IPv6 connection yet. 

This suggestion implements a low level change in the container env with a `gai.conf` file to prioritize IPv4 A responses, without exposing any unnecessary user knobs. As eventually, the expectations is EFS will not only supports IPv4 but IPv6 will be enabled as on all services - avoiding deprecation requirements for any user exposed configurations.

**What testing is done?** 
build and function testing

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM public.ecr.aws/eks-distro-build-tooling/golang:1.25.5 as go-builder
+FROM public.ecr.aws/eks-distro-build-tooling/golang:1.25.6-gcc-al23 as go-builder
 WORKDIR /go/src/github.com/kubernetes-sigs/aws-efs-csi-driver
 
 ARG TARGETOS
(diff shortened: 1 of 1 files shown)


abeowlu:
@DavidXU12345 wanted to request review & comments on this

samuhale:
EFS does support ipv6, which was added after #1609 was opened. https://aws.amazon.com/about-aws/whats-new/2025/06/amazon-efs-internet-protocol-version-6/ Even still, I think setting ipv4 to the default is not the right approach since ipv6 should be generally preferred when possible. At most, we should consider an option to configure this option, but I don't want to set it as the default for all customers.

Since the original issue is closed, I will close this PR as well. If this is still needed, please reopen the github issue and/or update this CR based on the feedback above

PR #1676, opened 2025-08-01
Feature/enhanced health 

oyiz-michael:
#1675 
**Is this a bug fix or adding new feature?**
Adding new feature

**What is this PR about? / Why do we need it?**
This PR implements enhanced EFS mount health monitoring to address critical production issues where pods crash-loop when EFS mounts degrade.

Problem: Current health probes only check if the CSI driver process is running, not whether EFS mounts actually work for I/O operations. This causes false positives where pods appear healthy but can't access storage.

Solution:

Real-time monitoring of EFS mount points with actual I/O testing
Multiple health endpoints (/healthz, /healthz/ready, /healthz/live, /healthz/mounts)
Prometheus-style metrics for observability
Enhanced CSI probe integration
Configurable health check intervals and timeouts
Benefits:

Prevents pod crash-loops by detecting mount issues before applications fail
Better observability with detailed mount health metrics
Kubernetes-native integration with readiness/liveness probes
Addresses issues #336, #1411, #1156.

**What testing is done?**
✅ Comprehensive unit tests for all health monitoring components
✅ Full driver test suite passes (31/31 tests)
✅ HTTP endpoint testing with proper status codes and JSON responses
✅ I/O testing validates actual file operations on mount points
✅ Code formatting (gofmt) and quality (go vet) verified
✅ Binary compilation successful
✅ Zero breaking changes - backward compatible

PR #1816, opened 2026-03-11
feat: CSI volume health monitoring and failover

### LINKED ISSUE #1675: Enhance EFS mount health monitoring with real-time I/O testing and modernized CSI probes
/feature

The AWS EFS CSI Driver currently lacks robust health monitoring capabilities for EFS mount points, leading to several critical production issues:

Pod Crash-Loop Problems: When EFS mounts degrade or become unavailable, pods continue to report as "healthy" through basic liveness probes, but actual I/O operations fail, causing application crashes and restart loops (as reported in issues #336, #1411, #1156).

Insufficient Health Visibility: The current health probes only check if the CSI driver process is running, not whether EFS mounts are actually functional for I/O operations. This leads to false positives where pods appear healthy but cannot access storage.

Reactive Problem Detection: Issues with EFS mounts are typically discovered after applications fail, rather than being proactively detected through proper health monitoring.

Limited Observability: There's no standardized way to monitor the health of individual EFS mounts or get detailed metrics about mount performance and reliability.

CSI Specification Gaps: The driver doesn't fully leverage modern CSI health monitoring capabilities that could provide better integration with Kubernetes health checking mechanisms.

I propose implementing a comprehensive Enhanced EFS Mount Health Monitoring System with the following components:

Core Health Monitoring
Real-time Mount Health Tracking: Monitor each EFS mount point continuously with configurable intervals
Actual I/O Testing: Perform real file read/write operations to verify mount functionality (not just mount status checks)
Timeout Protection: Configurable timeouts to prevent health checks from hanging on unresponsive mounts
Background Processing: Non-blocking health monitoring that doesn't interfere with normal CSI operations
Multiple Health Endpoints
/healthz - Overall driver health status
/healthz/ready - Readiness probe compatible endpoint
/healthz/live - Liveness probe compatible endpoint
/healthz/mounts - Detailed per-mount health information with JSON response
Observability & Metrics
Prometheus-style Metrics: Export mount health, response times, and volume metrics
Structured Logging: Detailed health check results with proper log levels
Health Summary API: Programmatic access to mount health status for monitoring tools
CSI Integration
Enhanced Probe Method: Modernized CSI Identity service Probe implementation
Mount Registration: Automatic registration/unregistration of mounts during volume operations
Graceful Lifecycle Management: Proper startup and shutdown of health monitoring components
Configuration Options
Configurable health check intervals (default: 30 seconds)
Configurable I/O operation timeouts (default: 10 seconds)
Optional health server port configuration
Integration with existing volume metrics settings

Simple Probe Configuration (as attempted in PR #1428): Only making existing probe timeouts configurable, but this doesn't address the root cause of mount health issues.

External Health Monitoring: Using external tools like Prometheus node exporter or custom monitoring scripts, but this adds complexity and doesn't integrate with CSI driver lifecycle.

Application-Level Health Checks: Having applications implement their own EFS health checks, but this duplicates effort and doesn't provide driver-level visibility.

Basic Mount Point Checks: Only checking if mount points exist without I/O testing, but this provides false positives when mounts are stale or unresponsive.

Existing CSI Health Standards: Using only standard CSI health probes, but these don't provide storage-specific health validation.

**Additional context**
Related Issues
Issue #336: Liveness Probe Failure - demonstrates ongoing probe reliability problems
Issue #1411: Health should be configurable - shows community need for better health configuration
Issue #1156: Liveness probe failures after EFS version changes
PR #1428: Make health options configurable - approved but stalled attempt at basic probe configuration

**Implementation Benefits**
Prevents Pod Crash-Loops: Proactive detection of EFS mount issues before applications fail
Production Ready: Comprehensive error handling, timeouts, and resource cleanup
Cloud Native Integration: Full Kubernetes and Prometheus ecosystem compatibility
Zero Downtime: Background health monitoring without blocking normal operations
Enterprise Observability: Detailed metrics and logging for production monitoring

abeowlu:
**Is this a bug fix or adding new feature?** Feature enhancement

This PR should close #1675 #1676 satisfactorily, as well as related request for CSI volume health monitoring, high availability and self-heal

- The CSI should at the minimum be able to monitor health issues on EFS volume mounts and surface volumeCondition status in volume information to kubelet and other operator
- At best, a self recovery of dead volume mount, caused by EFS service AZ failure, network failure, etc. should be possible, and failing that, trigger pod failover (possibly by some separately concerned operator; this behaviour would be out of scope for CSI)

In this PR the minimum use-case, monitor volume mounts an report health and condition along with stat info when they are probed by kubelet, is implemented.
As a stab at the best case scenario, an asyn attempt at volume remount, and failover if EFS server issue or hang is encountered is also implemented.

**What testing is done?** 
unit testing for health check and async recovery attempt flow

### DIFF
--- a/pkg/driver/driver.go
+++ b/pkg/driver/driver.go
@@ -100,6 +100,7 @@ func SetNodeCapOptInFeatures(volMetricsOptIn bool) []csi.NodeServiceCapability_R
 	if volMetricsOptIn {
 		klog.V(4).Infof("Enabling Node Service capability for Get Volume Stats")
 		nCaps = append(nCaps, csi.NodeServiceCapability_RPC_GET_VOLUME_STATS)
+		nCaps = append(nCaps, csi.NodeServiceCapability_RPC_VOLUME_CONDITION)
 	} else {
 		klog.V(4).Infof("Node Service capability for Get Volume Stats Not enabled")
 	}
--- a/pkg/driver/driver.go
+++ b/pkg/driver/driver.go
@@ -67,14 +67,15 @@ func NewDriver(options *Options, efsUtilsCfgPath string) *Driver {
 
 	nodeCaps := SetNodeCapOptInFeatures(*options.VolMetricsOptIn)
 	watchdog := newExecWatchdog(efsUtilsCfgPath, *options.EfsUtilsStaticFilesPath, "amazon-efs-mount-watchdog")
-	return &Driver{
+	vs := NewVolStatter()
+	d := &Driver{
 		endpoint:                 *options.Endpoint,
 		nodeID:                   cloud.GetMetadata().GetInstanceID(),
 		mounter:                  newNodeMounter(),
 		efsWatchdog:              watchdog,
 		cloud:                    cloud,
 		nodeCaps:                 nodeCaps,
-		volStatter:               NewVolStatter(),
+		volStatter:               vs,
 		volMetricsOptIn:          *options.VolMetricsOptIn,
 		volMetricsRefreshPeriod:  *options.VolMetricsRefreshPeriod,
 		volMetricsFsRateLimit:    *options.VolMetricsFsRateLimit,
@@ -88,6 +89,10 @@ func NewDriver(options *Options, efsUtilsCfgPath string) *Driver {
 		forceUnmountAfterTimeout: *options.ForceUnmountAfterTimeout,
 		unmountTimeout:           *options.UnmountTimeout,
 	}
+	if imp, ok := vs.(*VolStatterImpl); ok {
+		imp.SethealthDriver(d)
+	}
+	return d
 }
 
 func SetNodeCapOptInFeatures(volMetricsOptIn bool) []csi.NodeServiceCapability_RPC_Type {
--- a/pkg/driver/mount_health_ha.go
+++ b/pkg/driver/mount_health_ha.go
@@ -1,0 +1,88 @@
+package driver
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"path/filepath"
+	"time"
+
+	"k8s.io/klog/v2"
+)
+
+type MountParams struct {
+	source       string
+	target       string
+	mountOptions []string
+}
+
+func checkMountHealth(ctx context.Context, target string) (bool, error) {
+	osCtx, cancel := context.WithTimeout(ctx, 2*time.Second)
+	defer cancel()
+	done := make(chan error, 1)
+	mount_health := filepath.Join(target, ".efs_mount_health")
+	writeBytes := []byte(fmt.Sprintf("%s: mount I/O health ping at %v", target, time.Now().UTC()))
+
+	go func() {
+		defer os.Remove(mount_health)
+		err := os.WriteFile(mount_health, writeBytes, 0644)
+		done <- err
+	}()
+
+	select {
+	case err := <-done:
+		if err != nil {
+			if os.IsNotExist(err) {
+				return false, nil
+			}
+			return false, err
+		}
+		return true, nil
+	case <-osCtx.Done():
+		return false, fmt.Errorf("mount point %s health check timed out", target)
+	}
+}
+
+// High availability and self-heal attempt implementations for unhealthy mount points
+func (d *Driver) AsyncMountHealthRecovery(volId, volPath string) {
+	if !d.lockManager.lockMutex(volId, 1*time.Millisecond) {
+		klog.V(4).Infof("Mount health recovery already in progress for path %s, skipping duplicate attempt", volPath)
+		return
+	}
+	go func() {
+		defer d.lockManager.unlockMutex(volId)
+		params, ok := volumeMountOptions[volId]
+		if !ok {
+			klog.Errorf("Mount parameters not found for volume ID %s during health recovery attempt", volId)
+			return
+		}
+		source := params.source
+		target := params.target
+		mountOptions := params.mountOptions
+
+		klog.Infof("Attempting mount point health recovery for path: %s ", volPath)
+
+		if err := d.mounter.UnmountWithForce(volPath, d.unmountTimeout); err != nil {
+			klog.Errorf("Failed to unmount target %s during health recovery: %v", volPath, err)
+			return
+		}
+
+		// Attempt to remount the target path
+		if err := d.mounter.Mount(source, target, "efs", mountOptions); err != nil {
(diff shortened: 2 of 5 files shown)


pkg/driver/node.go
```diff
@@ -316,13 +323,17 @@ func (d *Driver) NodeGetVolumeStats(ctx context.Context, req *csi.NodeGetVolumeS
...
 	if err != nil {
 		return nil, status.Errorf(codes.Internal, "Could not get metrics: %v ", err)
 	}
 
+	klog.V(5).Infof("Compute volume metrics complete for Vol ID: %v, Vol Health: %v", volId, volMetrics.mountHealthy)
 	return &csi.NodeGetVolumeStatsResponse{
 		Usage: volMetrics.volUsage,
+		VolumeCondition: &csi.VolumeCondition{
+			Abnormal: !volMetrics.mountHealthy,
+			Message:  fmt.Sprintf("volume %v mount is healthy: %v", volId, !volMetrics.mountHealthy),
```
DavidXU12345, line 335: The message negates health. I think it should be `fmt.Sprintf("volume %v mount is healthy: %v", volId, volMetrics.mountHealthy)`

pkg/driver/mount_health_ha.go
```diff
@@ -0,0 +1,88 @@
...
+
+// High availability and self-heal attempt implementations for unhealthy mount points
+func (d *Driver) AsyncMountHealthRecovery(volId, volPath string) {
+	if !d.lockManager.lockMutex(volId, 1*time.Millisecond) {
+		klog.V(4).Infof("Mount health recovery already in progress for path %s, skipping duplicate attempt", volPath)
+		return
+	}
+	go func() {
+		defer d.lockManager.unlockMutex(volId)
+		params, ok := volumeMountOptions[volId]
```
DavidXU12345, line 54: `volumeMountOptions` is written in `NodePublishVolume/NodeUnpublishVolume` and read in `AsyncMountHealthRecovery` — the production read/writes are not under `mu` (only the tests take `mu`). Go will fatal error: concurrent map read and map write under load.

pkg/driver/vol_statter.go
```diff
@@ -52,6 +66,10 @@ func NewVolStatter() VolStatter {
 
 func (v VolStatterImpl) computeVolumeMetrics(volId, volPath string, refreshRate float64, fsRateLimit int) (*volMetrics, error) {
 	if value, ok := v.retrieveFromCache(volId); ok {
+		if _, ok := volStatterJobTracker[volId]; ok {
```
DavidXU12345, line 69: Read `volStatterJobTracker` requires `mu` 

because there might be multiple goroutines touching it.
- `computeVolumeMetrics` runs on the kubelet RPC handler goroutine (from `NodeGetVolumeStats`).
- Meanwhile a previously-spawned computeDiskUsage goroutine for some volume is running its deferred `mu.Lock(); delete(volStatterJobTracker, volId); mu.Unlock()`.
- And another kubelet RPC (kubelet polls volumes concurrently) can be inside launchVolStatsRoutine doing `mu.Lock(); volStatterJobTracker[volId] = true; mu.Unlock().`

pkg/driver/vol_statter.go
```diff
@@ -114,20 +133,71 @@ func (v VolStatterImpl) launchVolStatsRoutine(volId, volPath string, fsRateLimit
...
+	klog.V(5).Infof("Compute volume metrics invoked for Vol ID: %v, Sleeping for %v before execution", volId, waitTime)
+
+	volMetrics := &volMetrics{
+		mountHealthy: health,
+		volPath:      volPath,
+		timeStamp:    update_time,
+		volUsage:     []*csi.VolumeUsage{{Unit: csi.VolumeUsage_UNKNOWN}},
+	}
+	if _, ok := v.retrieveFromCache(volId); !ok {
+		volUsageCache[volId] = volMetrics
```
DavidXU12345, line 151: Write into `volUsageCache` requires `mu` since multiple goroutines may be spawned in `launchVolStatsRoutine` for `computeDiskUsage`

pkg/driver/mount_health_ha.go
```diff
@@ -0,0 +1,88 @@
...
+			klog.Errorf("Mount parameters not found for volume ID %s during health recovery attempt", volId)
+			return
+		}
+		source := params.source
+		target := params.target
+		mountOptions := params.mountOptions
+
+		klog.Infof("Attempting mount point health recovery for path: %s ", volPath)
+
+		if err := d.mounter.UnmountWithForce(volPath, d.unmountTimeout); err != nil {
```
DavidXU12345, line 65: I am concerned about forcing unmount and remount automatically just based on 5-second I/O timeout. Can we remove the health recovery mechanism (or at least gate it behind a specific boolean toggle) and only report health status.

pkg/driver/driver.go
```diff
@@ -88,13 +89,18 @@ func NewDriver(options *Options, efsUtilsCfgPath string) *Driver {
...
+	}
+	return d
 }
 
 func SetNodeCapOptInFeatures(volMetricsOptIn bool) []csi.NodeServiceCapability_RPC_Type {
 	var nCaps = []csi.NodeServiceCapability_RPC_Type{}
 	if volMetricsOptIn {
 		klog.V(4).Infof("Enabling Node Service capability for Get Volume Stats")
 		nCaps = append(nCaps, csi.NodeServiceCapability_RPC_GET_VOLUME_STATS)
+		nCaps = append(nCaps, csi.NodeServiceCapability_RPC_VOLUME_CONDITION)
```
DavidXU12345, line 103: We should guard it behind another specific flag.

pkg/driver/mount_health_ha.go
```diff
@@ -0,0 +1,88 @@
...
+	target       string
+	mountOptions []string
+}
+
+func checkMountHealth(ctx context.Context, target string) (bool, error) {
+	osCtx, cancel := context.WithTimeout(ctx, 5*time.Second)
+	defer cancel()
+	done := make(chan error, 1)
+	mount_health := filepath.Join(target, ".efs_mount_health")
+	writeBytes := []byte(fmt.Sprintf("%s: mount I/O health ping at %v", target, time.Now().UTC()))
```
DavidXU12345, line 24: We shouldn't write files for checking health.

DavidXU12345 (changes requested):
- have you done any e2e tests? like show the volume metrics? 
- Please guard the entire feature under a separate flag to minimize the risk
- I would suggest not do auto recovery for now or at least guard auto recovery under a separate flag if you have a strong reason to do so

PR #1836, opened 2026-04-08
feat(test): Allow alternative helm charts in e2e tests

JoeNorth:
New feature
The existing e2e test suite assumes that for each run the user wants to build and upload a new image for the driver and use it for the testing. In some cases we want to run this test suite against an existing image in a different registry or a separately maintained Helm chart altogether.

This PR introduces a number of options for overriding the default behavior and new options to support alternative images and helm charts. These changes ensure that the existing behavior is not altered if no new options are provided.

Changes include:
* Skipping image build and upload
* Specifying Kops state bucket. Bucket is created automatically if it does not exist.
* Helm chart repository and version tag options
* Helm chart registry login
* K8s image pull secret creation and cleanup

**What testing is done?** 
1. Running the e2e tests on kops when specifying only the cluster name and a kops state bucket to test that existing behavior has not changed. 
2. Running the e2e tests on kops with the chart and images hosted in a separate registry.


DavidXU12345:
@JoeNorth  Thanks for submitting the PR. Could you clarify more about the use case and give examples that how you would run E2E tests with these new options you added.

> In some cases we want to run this test suite against an existing image in a different registry or a separately maintained Helm chart altogether.

Also, what is the motivation / reason to test against an existing image rather than building a new image? is it just to save time? Thanks.

JoeNorth:
> @JoeNorth Thanks for submitting the PR. Could you clarify more about the use case and give examples that how you would run E2E tests with these new options you added.
> 
> > In some cases we want to run this test suite against an existing image in a different registry or a separately maintained Helm chart altogether.

For context, I work at Chainguard and we build FIPS and non-FIPS versions of the EFS CSI driver chart and images and currently have a number of EKS Addons in the AWS Marketplace for our FIPS versions.

After the build phase of our pipeline we'd like to be able to run the E2E test suite against the artifacts in our internal staging registry however given the current assumptions of the test suite that is not possible. Only after the testing phase is completed and passing can those artifacts in staging be promoted to production and be made available to our customers.

> Also, what is the motivation / reason to test against an existing image rather than building a new image? is it just to save time? Thanks.

As we're an AWS partner and have this available as an EKS Addon in the AWS Marketplace we want to make sure we can do our due diligence and test our images against the upstream test suite to minimize the chances that we falsely attribute any issues to upstream and waste your resources.

DavidXU12345:
@JoeNorth Thanks for clarification. Is it possible to deploy whatever CSI driver image you would like to test against and run e2e test with `--deploy-driver=false`?

PR #1845, opened 2026-04-10
Add note about s3files: prefix requirement for static provisioning

DavidXU12345:
docs update

Remind user to add `s3files` prefix in volumeHandle for static provisioning.

**What testing is done?** 
NA

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -27,8 +27,9 @@ The following CSI interfaces are implemented:
 * Node Service: NodePublishVolume, NodeUnpublishVolume, NodeGetCapabilities, NodeGetInfo, NodeGetId, NodeGetVolumeStats
 * Identity Service: GetPluginInfo, GetPluginCapabilities, Probe
 
-**Note**  
-Since Amazon EFS and Amazon S3 file systems can scale elastically, it doesn't really enforce any file system capacity. The actual storage capacity value in persistent volume and persistent volume claim is not used when creating the file system. However, since the storage capacity is a required field by Kubernetes, you must specify the value and you can use any valid value for the capacity.
+> **Note** When using static provisioning with an Amazon S3 file system, the `volumeHandle` in your PersistentVolume must include the `s3files:` prefix (e.g., `s3files:fs-01234567890abcdef0`).
+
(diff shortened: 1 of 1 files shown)


docs/README.md
```diff
@@ -27,8 +27,9 @@ The following CSI interfaces are implemented:
...
+> **Note** When using static provisioning with an Amazon S3 file system, the `volumeHandle` in your PersistentVolume must include the `s3files:` prefix (e.g., `s3files:fs-01234567890abcdef0`). Without this prefix, the driver will treat the file system as Amazon EFS, and the mount will fail. For dynamic provisioning, this is not required as the driver determines the file system type from the `provisioningMode` parameter in the StorageClass.
+
+> **Note** Since Amazon EFS and Amazon S3 file systems can scale elastically, it doesn't really enforce any file system capacity. The actual storage capacity value in persistent volume and persistent volume claim is not used when creating the file system. However, since the storage capacity is a required field by Kubernetes, you must specify the value and you can use any valid value for the capacity.
```
DanielRubinstein: nit: Amazon EFS and S3 Files are fully elastic and scalable file systems, and so we do not have to worry about the capacity as it automatically scales up or down based on usage. The actual storage capacity value in persistent volume and persistent volume claim is a placeholder value that must be specified (required by Kubernetes) but is not actually used. You can specify any valid value for the capacity.

docs/README.md
```diff
@@ -27,8 +27,9 @@ The following CSI interfaces are implemented:
 * Node Service: NodePublishVolume, NodeUnpublishVolume, NodeGetCapabilities, NodeGetInfo, NodeGetId, NodeGetVolumeStats
 * Identity Service: GetPluginInfo, GetPluginCapabilities, Probe
 
-**Note**  
-Since Amazon EFS and Amazon S3 file systems can scale elastically, it doesn't really enforce any file system capacity. The actual storage capacity value in persistent volume and persistent volume claim is not used when creating the file system. However, since the storage capacity is a required field by Kubernetes, you must specify the value and you can use any valid value for the capacity.
+> **Note** When using static provisioning with an Amazon S3 file system, the `volumeHandle` in your PersistentVolume must include the `s3files:` prefix (e.g., `s3files:fs-01234567890abcdef0`). Without this prefix, the driver will treat the file system as Amazon EFS, and the mount will fail. For dynamic provisioning, this is not required as the driver determines the file system type from the `provisioningMode` parameter in the StorageClass.
```
DanielRubinstein: I would remove the second and third sentences. I'm not sure what happens if you specify the prefix for dynamic provisioning (whether it works). Customers might do that if we say that it's "not required".

DanielRubinstein:
LGTM

PR #1846, opened 2026-04-13
feat(helm): add updateStartegy to deamonset and strategy to deployment

camaeel:

This PR adds update strategy cotnrolls that were introduced 3 years ago but disappeared in the mean time.

Allow control for update strategy for both controller and deamonset. It is especially important for Deamonset for bigger clusters. Replacement of 100+ nodes takes really long time.

**What testing is done?** 

`helm template` with different inputs

### DIFF
--- a/charts/aws-efs-csi-driver/values.yaml
+++ b/charts/aws-efs-csi-driver/values.yaml
@@ -146,6 +146,10 @@ controller:
   # Enable reading filesystem IDs from configmap/secret
   fileSystemIdRefs:
     enabled: false
+  # rollingUpdate for controller deployment strategy
+  rollingUpdate: {}
+    #  maxUnavailable: 1
+    #  maxSurge: 1
 
 ## Node daemonset variables
 
@@ -222,6 +226,11 @@ node:
   volumes: []
   volumeMounts: []
   kubeletPath: /var/lib/kubelet
+
+  # rollingUpdate for node deamonset updateStrategy.
+  rollingUpdate: {}
+  #  maxSurge: 0
+  #  maxUnavailable: 20%
 
 storageClasses: []
 # Add StorageClass resources like:
--- a/charts/aws-efs-csi-driver/templates/controller-deployment.yaml
+++ b/charts/aws-efs-csi-driver/templates/controller-deployment.yaml
@@ -15,6 +15,12 @@ spec:
   {{- if or (kindIs "float64" .Values.controller.revisionHistoryLimit) (kindIs "int64" .Values.controller.revisionHistoryLimit) }}
   revisionHistoryLimit: {{ .Values.controller.revisionHistoryLimit }}
   {{- end }}
+  {{- if .Values.controller.rollingUpdate }}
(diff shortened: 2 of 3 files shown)


DavidXU12345:
@camaeel Could you clarify the use case for changing `updateStrategy` to `onDelete`? We intentionally removed the `updateStrategy` configuration because we found that customers sometimes misconfigure it. The original motivation for adding it was to avoid helm upgrade timing out with the default 300-second timeout, but
customers can work around this by specifying a larger value for the `--timeout` option in `helm upgrade`. Thanks

camaeel:
@DavidXU12345 I don't want to switch to `onDelete` updateStrategy. 

But I would really like to have ability to set on Deamonset `.spec.updateStrategy.rollingUpdate.maxUnavailable` = 20%.
I have multiple clusters with 100+ and 200+ nodes. Upgrading daemonsets is always painful experience there. 

I don't want to extend default helm install time to anything longer - this seems to be a waste of time to wait 20 mintues (or retry a failed helm install) because only 1 pod is rotated at a time, when they are completely independent and could be rotated much faster. 

Even if people are missusing something doesn't mean it should be completely blocked.

If needed I can narrow down the scope of modifiable values (only contents of `.spec.updateStrategy.rollingUpdate`).

DavidXU12345:
>If needed I can narrow down the scope of modifiable values (only contents of .spec.updateStrategy.rollingUpdate).

@camaeel Thanks for the clarification. Yes, could you narrow it down and push an updated commit?

camaeel:
@DavidXU12345 

I changed to to expose only rollingUpdate setting.

charts/aws-efs-csi-driver/values.yaml
```diff
@@ -223,6 +227,12 @@ node:
   volumeMounts: []
   kubeletPath: /var/lib/kubelet
 
+  # rollingUpdate for node deamonset updateStrategy.
+  rollingUpdate: null
+  #  maxSurge: 0
+  #  maxUnavailable: 20%
+
```
DavidXU12345, line 234: nit: redundant white line

charts/aws-efs-csi-driver/values.yaml
```diff
@@ -146,6 +146,10 @@ controller:
   # Enable reading filesystem IDs from configmap/secret
   fileSystemIdRefs:
     enabled: false
+  # rollingUpdate for controller deployment strategy
+  rollingUpdate: null
```
DavidXU12345: Can we make it empty object `{}` rather than `null` to be consistent with other fields?

DavidXU12345 (changes requested):

charts/aws-efs-csi-driver/values.yaml
camaeel: fixed

DavidXU12345 (approved):

mrbobbytables:
/unassign
I’m not an owner and don’t do anything with this repo^^;;

PR #1861, opened 2026-04-20
fix: Avoid random mounttargetip for cross-account EFS mounts

vmishra22:
  Bug fix

  When a cross-account EFS PVC is dynamically provisioned with `awsRoleArn` in the secret but without `crossaccount=true`, the controller calls `DescribeMountTargets` and picks a single random mount target IP, baking it into the PV's volumeAttributes as `mounttargetip`. This single IP is used by every node in every AZ. If that AZ goes down, all pods lose EFS connectivity — not just the ones in the affected AZ.

  **Implementation:**

  `Cloud interface`: Add `DescribeAvailableMountTargets` method that returns all available mount targets for a file system (instead of picking a random one).

  `CreateVolume`: The cross-account mount target selection logic now has three branches:
  1. `crossaccount=true` → DNS-based resolution (no mount target IP in PV)
  2. `crossaccount` not set or `false`, with `az` parameter → use `DescribeMountTargets` with the specified AZ (deterministic, no random selection)
  3. `crossaccount` not set or `false`, no `az` → fetch all mount targets via `DescribeAvailableMountTargets`, build an AZ→IP JSON map, and store it as `mounttargetipmap` in the PV's volumeAttributes. Each CSI node parses this map and selects the mount target IP matching its own AZ, with fallback to any available mount target.

  `DeleteVolume`: Replace the random mount target selection with `selectMountTargetIP` helper that prefers the controller's own AZ mount target for the cleanup mount, falling back to any available target.

  The random mount target selection that caused the AZ failover issue is removed entirely.

  **What testing is done?**

  Unit tests added:
  - `TestSelectMountTargetIP`: 5 cases covering preferred AZ found, not found (fallback), empty targets, single target match/no-match
  - `TestBuildMountTargetIPMap`: 3 cases covering multiple targets, single target, API error
  - `TestGetCloudCrossAccountDefaults`: 5 cases covering crossaccount not set, with roleArn, explicit true/false, invalid value
  - `TestNodePublishVolumeMountTargetIpMap`: 3 cases covering node AZ in map, node AZ not in map (fallback), invalid JSON


pkg/driver/controller.go
```diff
@@ -429,13 +430,18 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
 				// the same AZ-ID as the client instance); mounttargetip should
 				// not be used as a mount option in this case.
 				volContext[CrossAccount] = strconv.FormatBool(true)
-			} else {
+			} else if azName != "" {
```
DavidXU12345, line 433: Shall we pass `azName` down to efs-utils via mount options? 

https://github.com/aws/efs-utils/blob/76f065c75f557f15ce83e942508e385c773ead58/src/mount_efs/dns_resolver.py#L40

pkg/driver/controller.go
```diff
@@ -562,12 +568,9 @@ func (d *Driver) DeleteVolume(ctx context.Context, req *csi.DeleteVolumeRequest)
 					// Connect via dns rather than mounttargetip
 					mountOptions = append(mountOptions, CrossAccount)
 				} else {
-					mountTarget, err := localCloud.DescribeMountTargets(ctx, fileSystemId, "", fsType)
-					if err == nil {
-						mountOptions = append(mountOptions, MountTargetIp+"="+mountTarget.IPAddress)
-					} else {
-						klog.Warningf("Failed to describe mount targets for file system %v. Skip using `mounttargetip` mount option: %v", fileSystemId, err)
-					}
+					klog.Warningf("Cross-account mount without crossaccount=true: no mounttargetip will be set. "+
```
DavidXU12345, line 571: Is the warning message correct? I think `efs-utils` will fetch the mount target based on the AZ where csi-node is. If there is no available mount target in the AZ where csi-node is, it will randomly pick one available mount target for it.

Could you also do more comprehensive testing for different scenarios? thanks

DavidXU12345 (changes requested):

pkg/driver/controller.go
vmishra22: Updated with the both cases when az is present and absent.

pkg/driver/controller.go
vmishra22: This code is updated now. Test results are updated in the ticket.

pkg/driver/controller.go
```diff
@@ -305,9 +307,10 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
 			accessPointsOptions.DirectoryPerms = value
 		}
 
-		// Storage class parameter `az` will be used to fetch preferred mount target for cross account mount.
-		// If the `az` storage class parameter is not provided, a random mount target will be picked for mounting.
-		// This storage class parameter different from `az` mount option provided by efs-utils https://github.com/aws/efs-utils/blob/v1.31.1/src/mount_efs/__init__.py#L195
+		// Storage class parameter `az` can be used to select a specific AZ's mount target for cross account mount.
+		// Previously, if `az` was not provided, a random mount target was selected — this caused AZ failover issues.
```
DavidXU12345: no need to mention previous behavior.

pkg/driver/controller.go
```diff
@@ -305,9 +307,10 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
...
+		// Now, cross-account mounts default to DNS-based resolution. The `az` parameter is retained for backward
```
DavidXU12345: I think user still needs to set `crossaccount=true` to make it DNS-based resolution?

pkg/driver/node.go
```diff
@@ -117,6 +117,17 @@ func (d *Driver) NodePublishVolume(ctx context.Context, req *csi.NodePublishVolu
...
+		case MountTargetIpMap:
+			var ipMap map[string]string
+			if err := json.Unmarshal([]byte(v), &ipMap); err != nil {
+				return nil, status.Errorf(codes.InvalidArgument, "Failed to parse %s: %v", MountTargetIpMap, err)
+			}
+			nodeAZ := d.cloud.GetMetadata().GetAvailabilityZone()
+			if ip, ok := ipMap[nodeAZ]; ok {
+				mountOptions = append(mountOptions, MountTargetIp+"="+ip)
+			} else {
+				klog.Warningf("No mount target IP for node AZ %s in map %v, mount may fail", nodeAZ, ipMap)
```
DavidXU12345: shouldn't we still pick a random available mount target IP here? Could you add some comments here and add unit tests for the mount target IP selection logic?

pkg/driver/controller.go
```diff
@@ -429,14 +432,34 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
...
+					klog.Warningf("Failed to describe mount targets for file system %v: %v", accessPointsOptions.FileSystemId, err)
+				} else {
+					ipMap := make(map[string]string, len(mountTargets))
+					for _, mt := range mountTargets {
+						ipMap[mt.AZName] = mt.IPAddress
+					}
+					jsonBytes, err := json.Marshal(ipMap)
+					if err != nil {
+						klog.Warningf("Failed to marshal mount target IP map: %v", err)
+					} else {
```
DavidXU12345: nit: could you try to refactor it to avoid nested code so that reader can follow it easier, thanks

pkg/driver/controller.go
```diff
@@ -562,11 +585,22 @@ func (d *Driver) DeleteVolume(ctx context.Context, req *csi.DeleteVolumeRequest)
...
 				} else {
-					mountTarget, err := localCloud.DescribeMountTargets(ctx, fileSystemId, "", fsType)
-					if err == nil {
-						mountOptions = append(mountOptions, MountTargetIp+"="+mountTarget.IPAddress)
-					} else {
-						klog.Warningf("Failed to describe mount targets for file system %v. Skip using `mounttargetip` mount option: %v", fileSystemId, err)
+					// No crossaccount DNS: resolve a mount target IP via the cross-account role
+					// so the controller can mount the filesystem for access point cleanup.
+					// Prefer the controller's own AZ for lower latency, fall back to any available.
+					mountTargets, err := localCloud.DescribeAvailableMountTargets(ctx, fileSystemId)
```
DavidXU12345, line 573: can we add unit tests of this MountTargetIp selection logic?

DavidXU12345 (changes requested):

pkg/driver/controller.go
vmishra22: Update the tests

pkg/driver/controller.go
vmishra22: Updated.

pkg/driver/controller.go
vmishra22: Yes, changed the comment.

pkg/driver/node.go
vmishra22: Updated the logic with additional tests.

pkg/driver/controller.go
vmishra22: Thanks, refactored the code into two methods

docs/parameters.md
```diff
@@ -13,15 +13,15 @@
...
 | ensureUniqueDirectory |        | true            | true     | **NOTE: Only set this to false if you're sure this is the behaviour you want**.<br/> Used when dynamic provisioning is enabled, if set to true, appends the a UID to the pattern specified in `subPathPattern` to ensure that access points will not accidentally point at the same directory.                                                                                                |
-| az                    |        | ""              | true     | Used for cross-account mount. `az` under storage class parameter is optional. If specified, mount target associated with the az will be used for cross-account mount. If not specified, a random mount target will be picked for cross account mount                                                                                                                                          |
+| az                    |        | ""              | true     | Used for cross-account mount. If specified, the mount target in the given AZ is used. If not specified, the driver passes all available mount target IPs to each node, which selects the mount target matching its own AZ.  |
```
DavidXU12345: We should mention "If no available mount targets in the csi node AZ, a random available mount target will be selected"

pkg/driver/controller.go
```diff
@@ -429,14 +427,24 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
 				// the same AZ-ID as the client instance); mounttargetip should
 				// not be used as a mount option in this case.
 				volContext[CrossAccount] = strconv.FormatBool(true)
-			} else {
+			} else if azName != "" {
+				// Use the specified AZ's mount target rather than picking randomly.
```
DavidXU12345: nit: no need to mention "rather than picking randomly"

pkg/driver/controller.go
```diff
@@ -429,14 +427,24 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
...
 				mountTarget, err := localCloud.DescribeMountTargets(ctx, accessPointsOptions.FileSystemId, azName, fsType)
 				if err != nil {
-					klog.Warningf("Failed to describe mount targets for file system %v. Skip using `mounttargetip` mount option: %v", accessPointsOptions.FileSystemId, err)
+					klog.Warningf("Failed to describe mount targets for file system %v. Consider using crossaccount=true for DNS-based resolution: %v", accessPointsOptions.FileSystemId, err)
```
DavidXU12345: so if there is `err`, controller won't pass `MountTargetIp` down to node and the mount will ultimately fail due to not able to resolve DNS since FS is in another account?

If that is the case, shall we just throw fatal error here to provide clearer error message for troubleshooting?

pkg/driver/node.go
```diff
@@ -117,6 +117,25 @@ func (d *Driver) NodePublishVolume(ctx context.Context, req *csi.NodePublishVolu
...
+			var ipMap map[string]string
+			if err := json.Unmarshal([]byte(v), &ipMap); err != nil {
+				return nil, status.Errorf(codes.InvalidArgument, "Failed to parse %s: %v", MountTargetIpMap, err)
+			}
+			nodeAZ := d.cloud.GetMetadata().GetAvailabilityZone()
+			if ip, ok := ipMap[nodeAZ]; ok {
+				mountOptions = append(mountOptions, MountTargetIp+"="+ip)
+			} else {
+				// No mount target in this node's AZ; pick any available one as fallback.
+				klog.Warningf("No mount target IP for node AZ %s in map %v, falling back to another AZ", nodeAZ, ipMap)
```
DavidXU12345: Shall we move the warning message inside the range over ipMap so that we know which AZ we fallback to.

pkg/driver/controller_test.go
```diff
@@ -5500,3 +5501,204 @@ func randStringBytes(n int) string {
...
+			mountTargets: []*cloud.MountTarget{
+				{AZName: "us-west-2a", IPAddress: "10.0.1.1"},
+				{AZName: "us-west-2b", IPAddress: "10.0.2.1"},
+				{AZName: "us-west-2c", IPAddress: "10.0.3.1"},
+			},
+			preferredAZ: "us-west-2b",
+			expectedIP:  "10.0.2.1",
+		},
+		{
+			name: "preferred AZ not found falls back to first",
```
DavidXU12345: nit: I think Go's range over a map iterates in a random order, so the description should be falls back to a random available AZ

DavidXU12345 (changes requested):

docs/parameters.md
vmishra22: Thanks, added this statement.

pkg/driver/controller.go
vmishra22: Updated

pkg/driver/controller.go
vmishra22: Thanks, that's a good suggestion. Changed the warning into error.

pkg/driver/node.go
vmishra22: Yes, updated.

pkg/driver/controller_test.go
vmishra22: Thanks, updated

pkg/driver/controller.go
```diff
@@ -562,11 +569,12 @@ func (d *Driver) DeleteVolume(ctx context.Context, req *csi.DeleteVolumeRequest)
...
-					if err == nil {
-						mountOptions = append(mountOptions, MountTargetIp+"="+mountTarget.IPAddress)
-					} else {
-						klog.Warningf("Failed to describe mount targets for file system %v. Skip using `mounttargetip` mount option: %v", fileSystemId, err)
+					// Resolve mount target IP for cross-account cleanup mount.
+					mountTargets, err := localCloud.DescribeAvailableMountTargets(ctx, fileSystemId)
+					if err != nil {
+						klog.Warningf("DeleteVolume: failed to describe mount targets for %v: %v", fileSystemId, err)
+					} else if ip := selectMountTargetIP(mountTargets, d.cloud.GetMetadata().GetAvailabilityZone()); ip != "" {
+						mountOptions = append(mountOptions, MountTargetIp+"="+ip)
```
DavidXU12345, line 577: Shall we also return fatal error here if ip cannot be found or `err != nil` ^?

vmishra22: I think it won't be appropriate to return error here, If we can't get a mount target IP, we should still try to delete the access point itself. Throwing an error here would block the entire delete operation, leaving orphaned access points.

DavidXU12345: Ack, thanks

DavidXU12345 (approved):

PR #1867, opened 2026-04-23
add e2e test for cross account setup

vmishra22:
new test

Adding e2e test to cover crossaccount setup scenarios

**What testing is done?** 
Locally tests were run successfully. 

 ## │ Test                                                                                           │ Time │ Result                               │
  
  │ 1   │ crossaccount not set, no `az` → `mounttargetipmap`        │ ~15s │ ✅ PASS                              │
  │ 2   │ `crossaccount=false` + `az` → single mount target           │ ~15s │ ✅ PASS                              │
  │ 3   │ crossaccount not set + `az` → single mount target            │ ~15s │ ✅ PASS                              │
  │ 4   │ `crossaccount=false`, no `az` → `mounttargetipmap`      │ ~15s │ ✅ PASS                              │
  │ 5   │ Multi-AZ simultaneous mount                                               │ ~15s │ ✅ PASS                              │
  │ 6   │ Pod rescheduling across AZs                                               │ ~15s │ ✅ PASS                              │
  │ 7   │ `mounttargetipmap` multi-AZ + per-node mount               │ ~15s │ ✅ PASS                              │
  │ 8   │ PVC deletion cleanup                                                            │ ~20s │ ✅ PASS                              │

### DIFF
--- a/test/e2e/e2e.go
+++ b/test/e2e/e2e.go
@@ -45,6 +45,16 @@ var (
 	S3FilesFileSystemName string
 	s3FilesResources      *S3FilesResources // Store complete S3Files resources for cleanup
 
+	// Cross-account parameters — when set, StorageClasses include the provisioner
+	// secret reference and iam mount option for cross-account EFS access.
+	// CrossAccountMountTargetIP is the mount target IP for static PV tests —
+	// static provisioning bypasses the controller, and AZ-specific EFS DNS isn't
+	// resolvable cross-VPC unless extra Route 53 setup is done; the customer
+	// pattern in the AWS docs is to pass mounttargetip directly.
+	CrossAccountSecretName    string
+	CrossAccountAZ            string
+	CrossAccountMountTargetIP string
+
 	// CreateFileSystem if set true will create a file system before tests.
 	// Alternatively, provide an existing file system via FileSystemId. If this
 	// is true, ClusterName and Region must be set. For CI it should be true
@@ -158,6 +168,8 @@ func (e *efsDriver) GetDynamicProvisionStorageClass(ctx context.Context, config 
 		"directoryPerms":   "777",
 	}
 
+	mountOptions := applyCrossAccountConfig(parameters)
+
 	generateName := fmt.Sprintf("efs-csi-dynamic-sc-test1234-")
 
 	defaultBindingMode := storagev1.VolumeBindingImmediate
@@ -167,6 +179,7 @@ func (e *efsDriver) GetDynamicProvisionStorageClass(ctx context.Context, config 
 		},
 		Provisioner:       "efs.csi.aws.com",
 		Parameters:        parameters,
+		MountOptions:      mountOptions,
 		VolumeBindingMode: &defaultBindingMode,
 	}
 }
@@ -635,8 +648,23 @@ func makeEFSPVCDynamicProvisioning(namespace, name string, storageClassName stri
 	}
 }
 
+// applyCrossAccountConfig adds cross-account secret reference, mount options,
+// and az parameter to StorageClass parameters when cross-account testing is enabled.
+func applyCrossAccountConfig(parameters map[string]string) []string {
+	if CrossAccountSecretName == "" {
+		return nil
+	}
+	parameters["csi.storage.k8s.io/provisioner-secret-name"] = CrossAccountSecretName
+	parameters["csi.storage.k8s.io/provisioner-secret-namespace"] = "kube-system"
+	if CrossAccountAZ != "" {
+		parameters["az"] = CrossAccountAZ
+	}
+	return []string{"tls", "iam"}
+}
+
 func GetStorageClass(params map[string]string) *storagev1.StorageClass {
 	parameters := params
+	mountOptions := applyCrossAccountConfig(parameters)
 
 	generateName := fmt.Sprintf("efs-csi-dynamic-sc-test1234-")
 
@@ -647,6 +675,7 @@ func GetStorageClass(params map[string]string) *storagev1.StorageClass {
 		},
 		Provisioner:       "efs.csi.aws.com",
 		Parameters:        parameters,
+		MountOptions:      mountOptions,
 		VolumeBindingMode: &defaultBindingMode,
 	}
 }
@@ -699,6 +728,17 @@ func makeEFSPV(name, path string, volumeAttributes map[string]string, config Fil
 	if path != "" {
 		volumeHandle += ":" + path
 	}
+
+	// Cross-account static PV tests: inject the mount target IP directly via
+	// the mounttargetip mount option. Static provisioning bypasses the CSI
+	// controller, and EFS DNS isn't resolvable cross-VPC without additional
+	// Route 53 setup
+	var mountOptions []string
+	if CrossAccountSecretName != "" && CrossAccountMountTargetIP != "" {
+		mountOptions = []string{"tls", "iam"}
+		volumeAttributes["mounttargetip"] = CrossAccountMountTargetIP
+	}
+
 	return &v1.PersistentVolume{
 		ObjectMeta: metav1.ObjectMeta{
 			Name: name,
@@ -715,7 +755,8 @@ func makeEFSPV(name, path string, volumeAttributes map[string]string, config Fil
 					VolumeAttributes: volumeAttributes,
 				},
 			},
-			AccessModes: []v1.PersistentVolumeAccessMode{v1.ReadWriteMany},
+			MountOptions: mountOptions,
+			AccessModes:  []v1.PersistentVolumeAccessMode{v1.ReadWriteMany},
 		},
 	}
 }
--- a/test/e2e/e2e_test.go
+++ b/test/e2e/e2e_test.go
@@ -72,7 +72,9 @@ func init() {
 	flag.StringVar(&combinedMountTargetSubnetIds, "mount-target-subnet-ids", "", "comma-separated list of subnet IDs to use for mount targets of provisioned EFS file system, only used if -file-system-id is not set")
 	flag.StringVar(&EfsDriverNamespace, "efs-driver-namespace", "kube-system", "namespace of EFS driver pods")
 	flag.StringVar(&combinedEfsDriverLabelSelectors, "efs-driver-label-selectors", "app=efs-csi-node", "comma-separated label selectors for EFS driver pods, follows the form key1=value1,key2=value2")
-
+	flag.StringVar(&CrossAccountSecretName, "cross-account-secret-name", "", "name of a pre-existing K8s secret in kube-system with awsRoleArn for cross-account EFS access")
+	flag.StringVar(&CrossAccountAZ, "cross-account-az", "", "availability zone for cross-account mount target selection (used with crossaccount=false)")
+	flag.StringVar(&CrossAccountMountTargetIP, "cross-account-mount-target-ip", "", "mount target IP for static PV tests; used when testing cross-account static provisioning")
 	flag.Parse()
 
 	var err error
--- a/test/e2e/README.md
+++ b/test/e2e/README.md
@@ -28,6 +28,10 @@ go test -v -timeout 0 ./... -report-dir=$ARTIFACTS -ginkgo.focus="\[efs-csi\]" -
 The E2E flags that you can pass to `go test` are defined in [e2e_test.go](https://github.com/kubernetes-sigs/aws-efs-csi-driver/blob/master/test/e2e/e2e_test.go#L66-L75).
 
 
+### Running Cross-Account E2E Tests
+Cross-account tests verify EFS behaviors when the filesystem is in a different AWS account than the EKS cluster. The same standard e2e test suite runs against a cluster pre-configured with cross-account Secret and StorageClass.
+
+
 ### Running Upgrade Test
 In order to test upgrades from previous releases to the current development version of the driver, the following steps can be followed:
 


test/e2e/e2e.go
```diff
@@ -606,6 +612,382 @@ var _ = ginkgo.Describe("[efs-csi]", func() {
...
+			)
+			pvc, err := f.ClientSet.CoreV1().PersistentVolumeClaims(namespace).Get(context.TODO(), pvcName, metav1.GetOptions{})
+			framework.ExpectNoError(err, "getting PVC")
+			pv, err := f.ClientSet.CoreV1().PersistentVolumes().Get(context.TODO(), pvc.Spec.VolumeName, metav1.GetOptions{})
+			framework.ExpectNoError(err, "getting PV")
+			return pv.Spec.CSI.VolumeAttributes
+		}
+
+		ginkgo.It("should dynamically provision with crossaccount=true and mount via DNS", func() {
+			ginkgo.Skip("Requires Route 53 Resolver for cross-account DNS resolution - skipping in cross-account e2e")
```
DavidXU12345: will this test be always skipped?

test/e2e/e2e.go
```diff
@@ -606,6 +612,382 @@ var _ = ginkgo.Describe("[efs-csi]", func() {
...
+				"directoryPerms":                                  "700",
+				"basePath":                                        "/e2e-cross-account",
+				"csi.storage.k8s.io/provisioner-secret-name":      secretName,
+				"csi.storage.k8s.io/provisioner-secret-namespace": "kube-system",
+			}
+			for k, v := range extraParams {
+				params[k] = v
+			}
+			sc := GetStorageClass(params)
+			sc.MountOptions = []string{"tls", "iam"}
```
DavidXU12345: I don't think we need to add them, just don't specify any mount options

test/e2e/e2e.go
```diff
@@ -606,6 +612,382 @@ var _ = ginkgo.Describe("[efs-csi]", func() {
...
+
+			ginkgo.By("Verifying PV has mounttargetipmap and no single mounttargetip")
+			attrs := getPVVolumeAttributes(pvc.Name, f.Namespace.Name)
+			if _, hasIpMap := attrs["mounttargetipmap"]; !hasIpMap {
+				ginkgo.Fail("PV should have mounttargetipmap when crossaccount is not set")
+			}
+			if _, hasMtIp := attrs["mounttargetip"]; hasMtIp {
+				ginkgo.Fail("PV should not have single mounttargetip")
+			}
+
```
DavidXU12345: shall we print out `attrs["mounttargetipmap"]` for debugging purpose?

test/e2e/e2e.go
```diff
@@ -606,6 +612,382 @@ var _ = ginkgo.Describe("[efs-csi]", func() {
...
+		ginkgo.It("should dynamically provision with az parameter when crossaccount=false", func() {
+			secretName := "e2e-xacct-az-" + rand.String(4)
+			secret := createCrossAccountSecret(secretName, map[string]string{"crossaccount": "false"})
+			defer f.ClientSet.CoreV1().Secrets("kube-system").Delete(context.TODO(), secret.Name, metav1.DeleteOptions{})
+
+			// Get an AZ from the cluster nodes
+			nodes, err := f.ClientSet.CoreV1().Nodes().List(context.TODO(), metav1.ListOptions{})
+			framework.ExpectNoError(err, "listing nodes")
+			nodeAZ := nodes.Items[0].Labels["topology.kubernetes.io/zone"]
```
DavidXU12345: I don't think we should use nodeAZ here? since even if we don't specify az, it would use nodeAZ for mount target  by default?

test/e2e/e2e.go
```diff
@@ -606,6 +612,382 @@ var _ = ginkgo.Describe("[efs-csi]", func() {
...
+			sc := createCrossAccountSC(secretName, map[string]string{"az": nodeAZ})
+			defer f.ClientSet.StorageV1().StorageClasses().Delete(context.TODO(), sc.Name, metav1.DeleteOptions{})
+
+			pvc, err := createEFSPVCPVDynamicProvisioning(f.ClientSet, f.Namespace.Name, f.Namespace.Name, sc.Name)
+			framework.ExpectNoError(err, "creating PVC")
+			defer f.ClientSet.CoreV1().PersistentVolumeClaims(f.Namespace.Name).Delete(context.TODO(), pvc.Name, metav1.DeleteOptions{})
+
+			ginkgo.By("Verifying PV has mounttargetip for the specified AZ")
+			attrs := getPVVolumeAttributes(pvc.Name, f.Namespace.Name)
+			if _, hasMtIp := attrs["mounttargetip"]; !hasMtIp {
```
DavidXU12345: nit: is there a way to check `mounttargetip` is in the AZ we configured?

test/e2e/e2e.go
```diff
@@ -606,6 +612,382 @@ var _ = ginkgo.Describe("[efs-csi]", func() {
...
+			}
+
+			ginkgo.By("Deploying pod and verifying read/write")
+			pod := e2epod.MakePod(f.Namespace.Name, nil, []*v1.PersistentVolumeClaim{pvc}, admissionapi.LevelBaseline, "echo cross-account-az-test > /mnt/volume1/out && cat /mnt/volume1/out")
+			pod, err = f.ClientSet.CoreV1().Pods(f.Namespace.Name).Create(context.TODO(), pod, metav1.CreateOptions{})
+			framework.ExpectNoError(err, "creating pod")
+			framework.ExpectNoError(e2epod.WaitForPodSuccessInNamespace(context.TODO(), f.ClientSet, pod.Name, f.Namespace.Name), "waiting for pod success")
+		})
+
+		ginkgo.It("should dynamically provision with az parameter when crossaccount is not set", func() {
```
DavidXU12345: I think it is sort of duplicated with previous test. It is fine just to test crossaccount is not set

test/e2e/e2e.go
```diff
@@ -606,6 +612,382 @@ var _ = ginkgo.Describe("[efs-csi]", func() {
...
+			}
+
+			ginkgo.By("Deploying pod and verifying read/write")
+			pod := e2epod.MakePod(f.Namespace.Name, nil, []*v1.PersistentVolumeClaim{pvc}, admissionapi.LevelBaseline, "echo cross-account-nocc-az-test > /mnt/volume1/out && cat /mnt/volume1/out")
+			pod, err = f.ClientSet.CoreV1().Pods(f.Namespace.Name).Create(context.TODO(), pod, metav1.CreateOptions{})
+			framework.ExpectNoError(err, "creating pod")
+			framework.ExpectNoError(e2epod.WaitForPodSuccessInNamespace(context.TODO(), f.ClientSet, pod.Name, f.Namespace.Name), "waiting for pod success")
+		})
+
+		ginkgo.It("should dynamically provision with mounttargetipmap when crossaccount=false and no az", func() {
```
DavidXU12345: is it same as "should dynamically provision with mounttargetipmap when crossaccount is not set"?

test/e2e/e2e.go
```diff
@@ -606,6 +612,382 @@ var _ = ginkgo.Describe("[efs-csi]", func() {
...
+			}
+
+			ginkgo.By("Deploying pod and verifying read/write")
+			pod := e2epod.MakePod(f.Namespace.Name, nil, []*v1.PersistentVolumeClaim{pvc}, admissionapi.LevelBaseline, "echo cross-account-false-noaz-test > /mnt/volume1/out && cat /mnt/volume1/out")
+			pod, err = f.ClientSet.CoreV1().Pods(f.Namespace.Name).Create(context.TODO(), pod, metav1.CreateOptions{})
+			framework.ExpectNoError(err, "creating pod")
+			framework.ExpectNoError(e2epod.WaitForPodSuccessInNamespace(context.TODO(), f.ClientSet, pod.Name, f.Namespace.Name), "waiting for pod success")
+		})
+
+		ginkgo.It("should mount from pods across all AZs simultaneously using mounttargetipmap", func() {
```
DavidXU12345: Why do we care they mount "simultaneously"? I think we want to check whether csi-nodes are using mount target in their own AZ using mounttargetipmap

test/e2e/e2e.go
```diff
@@ -606,6 +612,382 @@ var _ = ginkgo.Describe("[efs-csi]", func() {
...
+			targetNode := nodes.Items[0]
+			pod := e2epod.MakePod(f.Namespace.Name, nil, []*v1.PersistentVolumeClaim{pvc}, admissionapi.LevelBaseline, "echo per-node-selection-test > /mnt/volume1/out && cat /mnt/volume1/out")
+			pod.Spec.NodeSelector = map[string]string{"kubernetes.io/hostname": targetNode.Name}
+			pod, err = f.ClientSet.CoreV1().Pods(f.Namespace.Name).Create(context.TODO(), pod, metav1.CreateOptions{})
+			framework.ExpectNoError(err, "creating pod")
+			framework.ExpectNoError(e2epod.WaitForPodSuccessInNamespace(context.TODO(), f.ClientSet, pod.Name, f.Namespace.Name),
+				fmt.Sprintf("pod on node %s (AZ %s) should mount successfully using its AZ's mount target", targetNode.Name, targetNode.Labels["topology.kubernetes.io/zone"]))
+		})
+
+		ginkgo.It("should clean up access points on PVC deletion", func() {
```
DavidXU12345: why do we need this test for cross-account specifically?

test/e2e/e2e.go
```diff
@@ -606,6 +612,382 @@ var _ = ginkgo.Describe("[efs-csi]", func() {
...
+
+			ginkgo.By("Deploying pod on second node and reading data written by first pod")
+			pod2 := e2epod.MakePod(f.Namespace.Name, nil, []*v1.PersistentVolumeClaim{pvc}, admissionapi.LevelBaseline, "cat /mnt/volume1/resched-out | grep resched-data")
+			pod2.Spec.NodeSelector = map[string]string{"kubernetes.io/hostname": node2}
+			pod2, err = f.ClientSet.CoreV1().Pods(f.Namespace.Name).Create(context.TODO(), pod2, metav1.CreateOptions{})
+			framework.ExpectNoError(err, "creating second pod")
+			framework.ExpectNoError(e2epod.WaitForPodSuccessInNamespace(context.TODO(), f.ClientSet, pod2.Name, f.Namespace.Name), "second pod should read data written by first pod from different AZ")
+		})
+
+		ginkgo.It("should have mounttargetipmap with entries for multiple AZs and mount from specific node", func() {
```
DavidXU12345: What are we trying to verify here? seems to me just another duplicated test with cross-account not set and no az set

test/e2e/e2e.go
```diff
@@ -606,6 +612,382 @@ var _ = ginkgo.Describe("[efs-csi]", func() {
...
+
+			ginkgo.By("Waiting for all pods to succeed")
+			for _, pod := range pods {
+				framework.ExpectNoError(e2epod.WaitForPodSuccessInNamespace(context.TODO(), f.ClientSet, pod.Name, f.Namespace.Name),
+					fmt.Sprintf("waiting for pod %s to succeed", pod.Name))
+			}
+			framework.Logf("All %d pods across %d AZs mounted and wrote successfully", len(pods), len(azNodes))
+		})
+
+		ginkgo.It("should allow pod to remount after rescheduling to a different AZ", func() {
```
DavidXU12345: Why do we need to add this test case specifically for cross-account?

test/e2e/e2e.go
```diff
@@ -606,6 +612,382 @@ var _ = ginkgo.Describe("[efs-csi]", func() {
...
+				ObjectMeta: metav1.ObjectMeta{Name: name, Namespace: "kube-system"},
+				Data:       data,
+			}
+			secret, err := f.ClientSet.CoreV1().Secrets("kube-system").Create(context.TODO(), secret, metav1.CreateOptions{})
+			framework.ExpectNoError(err, "creating cross-account secret")
+			return secret
+		}
+
+		// Helper: create StorageClass with cross-account secret reference
+		createCrossAccountSC := func(secretName string, extraParams map[string]string) *storagev1.StorageClass {
```
DavidXU12345: we should do the setup in `SynchronizedBeforeSuite`

DavidXU12345 (changes requested):

test/e2e/e2e.go
vmishra22: This code is removed from e2e.go. DNS resolution tests will be added in pipeline separately

test/e2e/e2e.go
vmishra22: What I have found that for e2e testing:
  `tls` — Required because the mount goes through VPC peering to a mount target in a different VPC. EFS requires TLS for cross-account connections. Without it, the mount fails.

  `iam` — Required because the node needs to authenticate to the cross-account EFS filesystem. The EFS file system policy checks the caller's IAM identity to authorize ClientMount. Without iam, efs-utils doesn't include the IAM credentials in the TLS certificate, and EFS returns access denied by server.
  
  We hit this exact issue during testing — the mount command without iam produced:

  mount.nfs4: access denied by server while mounting 127.0.0.1:/

  Adding iam to mount options fixed it because efs-utils then uses the node's Pod Identity credentials for IAM-authenticated mounting.

  For same-account mounts, tls is added by default by the CSI driver and iam isn't needed because the EFS filesystem trusts the same account implicitly. That's why we only add these mount options when CrossAccountSecretName is set.

test/e2e/e2e.go
vmishra22: This test is removed now.

test/e2e/e2e.go
vmishra22: This test is removed now from e2e.go

test/e2e/e2e.go
vmishra22: This test is removed now from e2e.go

test/e2e/e2e.go
vmishra22: This test is removed now from e2e.go

test/e2e/e2e.go
vmishra22: This test is removed now from e2e.go

test/e2e/e2e.go
vmishra22: This test is removed now from e2e.go

test/e2e/e2e.go
vmishra22: This test is removed now from e2e.go

test/e2e/e2e.go
vmishra22: This test is removed now from e2e.go

test/e2e/e2e.go
vmishra22: This test is removed now from e2e.go

test/e2e/e2e.go
vmishra22: Yes, the tests are moved to pipeline now.

test/e2e/e2e_test.go
```diff
@@ -72,6 +72,9 @@ func init() {
 	flag.StringVar(&combinedMountTargetSubnetIds, "mount-target-subnet-ids", "", "comma-separated list of subnet IDs to use for mount targets of provisioned EFS file system, only used if -file-system-id is not set")
 	flag.StringVar(&EfsDriverNamespace, "efs-driver-namespace", "kube-system", "namespace of EFS driver pods")
 	flag.StringVar(&combinedEfsDriverLabelSelectors, "efs-driver-label-selectors", "app=efs-csi-node", "comma-separated label selectors for EFS driver pods, follows the form key1=value1,key2=value2")
+	flag.StringVar(&CrossAccountSecretName, "cross-account-secret-name", "", "name of a pre-existing K8s secret in kube-system with awsRoleArn for cross-account EFS access")
+	flag.StringVar(&CrossAccountAZ, "cross-account-az", "", "availability zone for cross-account mount target selection (used with crossaccount=false)")
+	flag.StringVar(&CrossAccountMountTargetIPMap, "cross-account-mount-target-ip-map", "", "JSON map of AZ to mount target IP for cross-account static provisioning, e.g. '{\"us-east-2a\":\"10.0.1.1\",\"us-east-2b\":\"10.0.2.1\"}'")
```
DavidXU12345: why do we need to expose mount target ip map?

test/e2e/e2e.go
```diff
@@ -699,6 +723,17 @@ func makeEFSPV(name, path string, volumeAttributes map[string]string, config Fil
...
 		volumeHandle += ":" + path
 	}
+
+	// Cross-account static PV tests use the crossaccount mount option. This
+	// requires Route 53 private hosted zones in the client VPC that resolve the
+	// AZ-specific EFS DNS names (e.g. use2-az1.fs-xxx.efs.region.amazonaws.com)
+	// to the cross-account mount target IPs, per efs-utils docs.
+	var mountOptions []string
+	if CrossAccountSecretName != "" {
+		mountOptions = []string{"tls", "iam"}
```
DavidXU12345, line 738: just a reminder: https://aws.amazon.com/blogs/storage/mount-amazon-efs-file-systems-cross-account-from-amazon-eks/ => it also uses `tls` and `iam` as mount options. we should highlight it in github docs as well

test/e2e/e2e_test.go
vmishra22: Not needed anymore

DavidXU12345 (approved):

test/e2e/e2e.go
vmishra22: Yes, I will add this note in documentation update ticket.

PR #1871, opened 2026-04-29
feat(e2e): Add crossaccount mode flag for cross-account e2e tests

vmishra22:
e2e tests

Add --cross-account-secret-crossaccount-mode flag to control static PV behavior in cross-account tests:
- 'true': sets crossaccount=true in PV volumeAttributes (DNS-based)
- 'false': sets crossaccount=false + mounttargetip (explicit)
- unset: uses mounttargetip only (default behavior)

This enables testing all three cross-account provisioning modes: crossaccount=true, explicit crossaccount=false, and default (unset).

**What testing is done?** 
Tested locally for all scenarios

### DIFF
--- a/test/e2e/e2e.go
+++ b/test/e2e/e2e.go
@@ -47,13 +47,21 @@ var (
 
 	// Cross-account parameters — when set, StorageClasses include the provisioner
 	// secret reference and iam mount option for cross-account EFS access.
-	// CrossAccountMountTargetIP is the mount target IP for static PV tests —
-	// static provisioning bypasses the controller, and AZ-specific EFS DNS isn't
-	// resolvable cross-VPC unless extra Route 53 setup is done; the customer
-	// pattern in the AWS docs is to pass mounttargetip directly.
-	CrossAccountSecretName    string
-	CrossAccountAZ            string
-	CrossAccountMountTargetIP string
+	// CrossAccountSecretCrossaccountMode indicates the crossaccount value in the
+	// pre-configured cross-account Secret. When set to "true", the test secret
+	// has crossaccount=true (efs-utils resolves AZ-specific DNS — requires
+	// Route 53 AZ-specific zones in the client VPC). When empty/unset, the
+	// secret omits the crossaccount field (defaults to false), and the
+	// controller injects mount target IP(s) into the PV volumeAttributes.
+	// This flag also controls static PV test behavior:
+	//   - "true": PV has crossaccount=true volumeAttribute (+ tls, iam mount opts)
+	//   - else:   PV has mounttargetip volumeAttribute (+ tls, iam mount opts)
+	// CrossAccountMountTargetIP is the mount target IP used for static PV tests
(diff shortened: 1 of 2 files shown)


test/e2e/e2e.go
```diff
@@ -729,14 +737,24 @@ func makeEFSPV(name, path string, volumeAttributes map[string]string, config Fil
...
+	// (b) inject the mount target IP directly via mounttargetip
 	var mountOptions []string
-	if CrossAccountSecretName != "" && CrossAccountMountTargetIP != "" {
+	if CrossAccountSecretName != "" {
 		mountOptions = []string{"tls", "iam"}
-		volumeAttributes["mounttargetip"] = CrossAccountMountTargetIP
+		if CrossAccountSecretCrossaccountMode == "true" {
+			volumeAttributes["crossaccount"] = "true"
+		} else {
+			if CrossAccountSecretCrossaccountMode == "false" {
```
DavidXU12345: I don't think we need to check `CrossAccountSecretCrossaccountMode == "false"` again here.

test/e2e/e2e.go
```diff
@@ -729,14 +737,24 @@ func makeEFSPV(name, path string, volumeAttributes map[string]string, config Fil
...
+				volumeAttributes["crossaccount"] = "false"
+			}
+			if CrossAccountMountTargetIP != "" {
```
DavidXU12345: We require `CrossAccountMountTargetIP` not to be empty when `CrossAccountSecretCrossaccountMode` is not true. so I think we should throw fatal error here to crash the e2e test early. Or even better, we should have some `validate` function being called to check parameters before running e2e tests

DavidXU12345 (changes requested):

test/e2e/e2e.go
vmishra22: Added a parameter validation

test/e2e/e2e.go
vmishra22: Updated.

test/e2e/e2e_test.go
```diff
@@ -84,6 +85,11 @@ func init() {
 	}
 	MountTargetSecurityGroupIds = parseCommaSeparatedStrings(combinedMountTargetSecurityGroupIds)
 	MountTargetSubnetIds = parseCommaSeparatedStrings(combinedMountTargetSubnetIds)
+
+	// Validate cross-account parameters
+	if CrossAccountSecretName != "" && CrossAccountSecretCrossaccountMode != "true" && CrossAccountMountTargetIP == "" {
+		log.Fatalln("--cross-account-mount-target-ip is required when --cross-account-secret-name is set and --cross-account-secret-crossaccount-mode is not 'true'")
```
DavidXU12345: nit: can you also describe the reason in the log like "because static provisioning requires mounttargetip when cross-account DNS is not enabled". Also could you refactor this into a `validate()` function?

vmishra22: Updated

DavidXU12345 (approved):

PR #1875, opened 2026-04-30
Add e2e test for reuseAccessPoint StorageClass parameter for EFS

YangjinanHu:
new feature
Adds an e2e test for the reuseAccessPoint StorageClass parameter. This feature allows users to reconnect to the same access point from a different cluster by using a deterministic client token derived from the PVC name, but it had no end-to-end test coverage.

Adds a new Ginkgo test that validates the reuseAccessPoint feature:
1. Creates a StorageClass with reuseAccessPoint=true
2. Creates a PVC, writes data, then deletes the PVC/PV
3. Creates a new PVC with the same name
4. Verifies data persists (proving the access point was reused)

Note: reuseAccessPoint is not supported for S3files

**What testing is done?** 
Tested locally

### DIFF
--- a/test/e2e/e2e.go
+++ b/test/e2e/e2e.go
@@ -18,9 +18,11 @@ import (
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/apimachinery/pkg/util/wait"
 	clientset "k8s.io/client-go/kubernetes"
 	"k8s.io/kubernetes/test/e2e/framework"
 	e2epod "k8s.io/kubernetes/test/e2e/framework/pod"
+	e2epv "k8s.io/kubernetes/test/e2e/framework/pv"
 	storageframework "k8s.io/kubernetes/test/e2e/storage/framework"
 	"k8s.io/kubernetes/test/e2e/storage/testsuites"
 	"k8s.io/kubernetes/test/e2e/storage/utils"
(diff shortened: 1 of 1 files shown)


test/e2e/e2e.go
```diff
@@ -495,6 +497,88 @@ var _ = ginkgo.Describe("[efs-csi]", func() {
...
+					framework.ExpectNoError(e2epv.WaitForPersistentVolumeDeleted(context.TODO(), f.ClientSet, firstPVName, 5*time.Second, 2*time.Minute), "waiting for first pv deletion")
+
+					ginkgo.By("Creating second PVC with the same name to trigger access point reuse")
+					pvc2, err := createEFSPVCPVDynamicProvisioning(f.ClientSet, f.Namespace.Name, pvcName, sc.Name)
+					framework.ExpectNoError(err, "creating second pvc")
+					defer func() {
+						_ = f.ClientSet.CoreV1().PersistentVolumeClaims(f.Namespace.Name).Delete(context.TODO(), pvc2.Name, metav1.DeleteOptions{})
+					}()
+
+					ginkgo.By("Reading data from second PVC to verify access point was reused")
```
DavidXU12345, line 585: Can we verify the same access point was used in second PVC?

test/e2e/e2e.go
```diff
@@ -495,6 +497,88 @@ var _ = ginkgo.Describe("[efs-csi]", func() {
...
+					defer f.ClientSet.CoreV1().Pods(f.Namespace.Name).Delete(context.TODO(), pod.Name, metav1.DeleteOptions{})
+
+					output := kubectl.RunKubectlOrDie(f.Namespace.Name, "exec", pod.Name, "--", "/bin/sh", "-c", readCommand)
+					output = strings.TrimSuffix(output, "\n")
+					framework.Logf("Read output: %s", output)
+
+					if output == "" {
+						ginkgo.Fail("Read data is empty — access point was not reused.")
+					}
+					if output != testData {
```
DavidXU12345, line 600: We should cleanup second PV

DavidXU12345 (changes requested):

test/e2e/e2e.go
YangjinanHu: Updated

test/e2e/e2e.go
YangjinanHu: Updated

DavidXU12345 (approved):

PR #1876, opened 2026-05-01
Use pagniation for EFS DescribeAccessPoint

DavidXU12345:
New feature

Use pagniation for `efs.DescribeAccessPoint`, same as `s3files.ListAccessPointsInput`

**What testing is done?** 
Unit tests

### DIFF
--- a/pkg/cloud/cloud.go
+++ b/pkg/cloud/cloud.go
@@ -47,8 +47,8 @@ const (
 	AccessPointAlreadyExists        = "AccessPointAlreadyExists"
 	ConflictException               = "ConflictException"
 	PvcNameTagKey                   = "pvcName"
-	AccessPointPerFsLimit           = 10000
 	s3FilesListAccessPointsPageSize = 1000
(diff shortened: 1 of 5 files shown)


pkg/cloud/cloud.go
```diff
@@ -434,36 +434,42 @@ func (c *cloud) FindAccessPointByClientToken(ctx context.Context, clientToken, f
 
 	switch fsType {
 	case util.FileSystemTypeEFS:
-		// TODO: Use pagniation implementation to reduce the value of MaxResults
-		describeAPInput := &efs.DescribeAccessPointsInput{
-			FileSystemId: &fileSystemId,
-			MaxResults:   aws.Int32(AccessPointPerFsLimit), // 10000 is the uppper bound for MaxResults on efs DescribeAccessPoints
```
samuhale, line 440: Should we use a maxResults of 1000 on these APIs to match the s3files implementation?

samuhale (approved):

PR #1878, opened 2026-05-01
fix(e2e): validate encryptInTransit via efs-proxy --tls flag

YangjinanHu:
bug fix
With efs-utils v2, both TLS and non-TLS mounts route through efs-proxy on 127.0.0.1, so the mount-table check in testEncryptInTransit cannot distinguish encrypted from unencrypted mounts.

Replace mount table inspection with direct efs-proxy process argument checking: exec into the efs-csi-node driver pod, find the efs-proxy process for the target filesystem, and assert --tls flag presence/absence.

 
**What testing is done?** 
locally

### DIFF
--- a/test/e2e/e2e.go
+++ b/test/e2e/e2e.go
@@ -4,12 +4,14 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"regexp"
 	"strconv"
 	"strings"
 	"time"
 
 	ginkgo "github.com/onsi/ginkgo/v2"
 	"k8s.io/apimachinery/pkg/util/rand"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/kubernetes/test/e2e/framework/kubectl"
 
 	v1 "k8s.io/api/core/v1"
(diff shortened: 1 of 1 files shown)


DavidXU12345 (commented):
## Code Review Observations (non-blocking)

Overall this looks good — the approach of checking efs-proxy process args is the right way to validate TLS behavior with efs-utils v2. CI is green. A few minor suggestions:

### 1. Python one-liner readability

The inline Python script passed to `kubectl exec` is dense and hard to read/maintain as a Go string literal. Consider a simpler shell approach:

```
cat /proc/*/cmdline 2>/dev/null | tr '\\0' ' ' | grep efs-proxy
```

This avoids the Python dependency, though since the `efs-plugin` container is guaranteed to have Python (efs-utils depends on it), the current approach works fine.

### 2. Redundant context timeout

```go
waitCtx, waitCancel := context.WithTimeout(context.Background(), 60*time.Second)
defer waitCancel()
waitErr := wait.PollUntilContextTimeout(waitCtx, 5*time.Second, 60*time.Second, true, ...)
```

`PollUntilContextTimeout` already manages its own deadline internally from the timeout parameter. The outer `context.WithTimeout` is redundant. You could simplify to:

```go
waitErr := wait.PollUntilContextTimeout(context.Background(), 5*time.Second, 60*time.Second, true, ...)
```

### 3. PVC name matching assumption

```go
if strings.Contains(line, fsID) && strings.Contains(line, pvcName) {
```

This assumes efs-proxy's command-line arguments include the PVC name. If efs-proxy uses the PV name or volume handle instead, this match could fail silently (timeout after 60s with no clear error about *why* no line matched). Worth confirming this assumption holds — though since E2E passed, it's empirically validated.

### 4. Positive notes
- Adding `defer` cleanup for PVC (previously only PV was cleaned up) is a correctness improvement.
- The `expectTLS` logic correctly handles all three cases (nil → true, true → true, false → false).
- The retry loop handles the race between pod Running and efs-proxy startup.
- The regex `(^|\s)--tls(\s|$)` correctly avoids false matches on `--tls-port` or similar flags.

DavidXU12345 (approved):

PR #1879, opened 2026-05-05
test(e2e): Add upgrade E2E test for driver version upgrades

DavidXU12345:
Add a Ginkgo E2E test that verifies existing mounted EFS volumes survive a driver upgrade and new volumes work after upgrade. The test:

1. Deploys an old driver version via Helm
2. Creates pods with EFS volumes and writes test data
3. Upgrades to the new driver version via Helm
4. Verifies existing mounts still work (read/write)
5. Creates new pods with new EFS volumes post-upgrade
6. Verifies new mounts work correctly

This replaces the shell-based upgrade test that only checked deployment success without functional verification.

**What testing is done?**


test/e2e/e2e.go
```diff
@@ -602,7 +602,7 @@ var _ = ginkgo.Describe("[efs-csi]", func() {
 					}
 				})
 
-				ginkgo.It("should continue reading/writing after the driver pod is restarted", func() {
+				ginkgo.It("should continue reading/writing after the driver pod is restarted", ginkgo.Serial, func() {
```
DavidXU12345, line 605: It will force s3files and EFS test case run in sequence. Otherwise, it may have race condition (e.g. driver pod dies when running s3files test which may cause EFS test fails)

test/e2e/upgrade_test.go
```diff
@@ -0,0 +1,267 @@
...
+	framework.Logf("helm %s: %s", strings.Join(args[:2], " "), string(output))
+	if err != nil {
+		framework.Failf("helm command failed: %v\nOutput: %s", err, string(output))
+	}
+}
+
+// waitForDriverReady waits for the EFS CSI driver DaemonSet pods to be ready.
+func waitForDriverReady(f *framework.Framework) {
+	err := wait.PollUntilContextTimeout(context.TODO(), 5*time.Second, 3*time.Minute, true, func(ctx context.Context) (bool, error) {
+		ds, err := f.ClientSet.AppsV1().DaemonSets(helmNamespace).Get(ctx, "efs-csi-node", metav1.GetOptions{})
```
vmishra22, line 261: I may be wrong here but should we also wait for `efs-csi-controller` to be up as well (which maybe needed in dynamic provisioning)?

test/e2e/upgrade_test.go
```diff
@@ -0,0 +1,267 @@
...
+	"k8s.io/kubernetes/test/e2e/framework/kubectl"
+	e2epod "k8s.io/kubernetes/test/e2e/framework/pod"
+	admissionapi "k8s.io/pod-security-admission/api"
+)
+
+const (
+	helmReleaseName     = "aws-efs-csi-driver"
+	helmNamespace       = "kube-system"
+	publicEFSDriverRepo = "public.ecr.aws/efs-csi-driver/amazon/aws-efs-csi-driver"
+	localHelmChartDir   = "../../charts/aws-efs-csi-driver"
```
vmishra22: nit: maybe there is a better alternative rather than hardcoded path to not always run the test from `test/e2e`

test/e2e/e2e.go
```diff
@@ -602,7 +602,7 @@ var _ = ginkgo.Describe("[efs-csi]", func() {
...
 					const FilePath = "/mnt/testfile.txt"
 					const TestDuration = 30 * time.Second
```
vmishra22, line 607: Maybe needs to be deleted

DavidXU12345: I added `ginkgo.Serial` here since S3Files and EFS tests will run the same test case in parallel without it. Since this test would delete csi-node pods, it will cause race condition and fail the test if running in parallel.

test/e2e/upgrade_test.go
DavidXU12345: I think the alternative is to ask user to input the specific directory. We can expose it if we need to override it later.

test/e2e/upgrade_test.go
DavidXU12345: Fixed it. also wait for `efs-csi-controller`.

vmishra22 (approved):

PR #1880, opened 2026-05-06
Add --efs-utils-conf-overrides and --s3files-utils-conf-overrides CLI flags

### LINKED ISSUE #1833: Add support for configuring efs-utils.conf settings via environment variables or Helm values

Yes. The EFS CSI driver currently bundles efs-utils.conf in the container image with no way to override specific settings without replacing the entire configuration file. This creates a maintenance burden when operators need to tune specific parameters for production reliability.

We encountered a race condition on Bottlerocket nodes where TLS certificate rotation (every 60 minutes via `tls_cert_renewal_interval_min = 60`) kills stunnel (or `efs-proxy`) during mass pod evictions. When this coincides with 60+ pods being terminated, the watchdog's health check interval of 5 minutes (`stunnel_health_check_interval_min = 5`) is too slow to detect and restart the dead stunnel process. This causes mount operations to fill all `--max-inflight-mount-calls` slots with stuck goroutines, deadlocking the node.

Add support for overriding specific efs-utils.conf settings via environment variables or Helm values, so operators can tune parameters without owning the entire config file.

Option 1: Environment Variables

Option 2: Helm Values

The driver's entrypoint would merge these overrides with the bundled efs-utils.conf, preserving upstream defaults for settings not explicitly overridden. This allows critical production tuning while automatically receiving updates from new efs-utils versions (new regions, security patches, feature flags).

Own the entire efs-utils.conf via ConfigMap - This works but means we miss automatic updates when new AWS regions are added, security patches are applied, or new configuration options are introduced in efs-utils updates.

Init container to copy and modify config - Similar to option 1, requires maintaining a full copy of the config.

Fork and maintain a custom image - Too much overhead for what should be a simple configuration override.

All alternatives require tracking upstream changes manually and rebasing our config file with each efs-utils release.

**Additional context**

- Our specific fix includes `stunnel_health_check_interval_min = 1` (down from default 5) to detect dead stunnel within 60 seconds instead of 5 minutes
- Combined with `--force-unmount-after-timeout=true` and `--max-inflight-mount-calls=30`, this reduces the chance the node will deadlock during TLS cert rotation
- This issue is more prevalent on Bottlerocket (read-only root FS) than AL2, making configuration flexibility even more important
- Related to #1821 about efs-utils.conf not updating during Helm upgrades

DavidXU12345:
## Summary

  Adds two new CLI flags that allow users to override arbitrary values in the generated efs-utils.conf and s3files-utils.conf files without building custom images.

  Fixes #1833

  ## Changes

  - New flags: `--efs-utils-conf-overrides` and `--s3files-utils-conf-overrides`
  - Format: comma-separated `section:key=value` pairs
  - Example: `--efs-utils-conf-overrides="mount-watchdog:stunnel_health_check_interval_min=1,mount-watchdog:tls_cert_renewal_interval_min=30"`
  - Overrides are parsed and validated at startup in `Options.Validate()` — malformed strings cause immediate exit with a clear error
  - Strict mode: driver fatals if a section or key doesn't exist in the config, preventing silent misconfiguration
  - Commented-out keys are supported — they are uncommented and set to the override value
  - Pre-parsed `[]ConfOverride` slices are passed to the watchdog (no re-parsing at config-write time)
  - Overrides are applied after template rendering, so they take final precedence over other flags (e.g., `--efs-cloudwatch-log-enabled`)
  - Helm chart values exposed: `node.efsUtilsConfOverrides` and `node.s3filesUtilsConfOverrides`

  ## Testing

  - Unit tests for parsing, applying overrides, and startup validation of malformed input
  - `make test` and `make verify` pass

  ## Manual E2E Validation

  | Test | Result |
  |------|--------|
  | Driver deployed and pods healthy (2/2 Running, 0 restarts) | PASS |
  | Override flags present in daemonset args | PASS |
  | efs-utils.conf: `stunnel_health_check_interval_min=1` under `[mount-watchdog]` | PASS |
  | efs-utils.conf: `tls_cert_renewal_interval_min=30` under `[mount-watchdog]` | PASS |
  | s3files-utils.conf: `retry_nfs_mount_command_count=5` under `[proxy]` | PASS |
  | Invalid section causes startup crash with clear error message | PASS |
  | Invalid key causes startup crash with clear error message | PASS |
  | Logs show "Applied config override" messages for all keys | PASS |
  | Dynamic provisioning (EFS access points) works with overridden config | PASS |


pkg/driver/overrides.go
```diff
@@ -0,0 +1,135 @@
...
+		if !replaced {
+			newLine := o.Key + " = " + o.Value
+			lines = append(lines[:nextSectionIdx], append([]string{newLine}, lines[nextSectionIdx:]...)...)
+		}
+	}
+	return strings.Join(lines, "\n")
+}
+
+// ValidateOverrides parses the override string and checks for conflicts with managed keys.
+func ValidateOverrides(raw string, managedKeys map[string]string) error {
```
DavidXU12345: Let's simplify the logic to make overrides always win so we don't need to `ValidateOverrides` here

pkg/driver/conf_overrides.go
```diff
@@ -0,0 +1,97 @@
+package driver
```
DavidXU12345, line 1: rename the file to `conf_overrides` and the type name to `ConfOverride` to be clearer

pkg/driver/options.go
vmishra22, line 82: maybe this method should have more validation for the options to catch the error early during startup. For now, maybe new ones added in this PR can be validated.

charts/aws-efs-csi-driver/values.yaml
```diff
@@ -232,6 +232,9 @@ node:
   #  maxSurge: 0
   #  maxUnavailable: 20%
 
+  efsUtilsConfOverrides: ""
+  s3filesUtilsConfOverrides: ""
+
```
vmishra22, line 241: nit: maybe some comment here about these values

vmishra22 (commented):
minor comments

pkg/driver/overrides.go
DavidXU12345: Done. Removed `ValidateOverrides` and the managed keys maps. Overrides always win now.

pkg/driver/conf_overrides.go
DavidXU12345: Done. File renamed to `conf_overrides.go` and type renamed to `ConfOverride`.

pkg/driver/options.go
DavidXU12345: Done. `Validate()` now calls `parseConfOverrides()` on both override flags at startup. Malformed strings return an error immediately, and the parsed `[]ConfOverride` slices are stored and passed directly to the watchdog (no re-parsing at config-write time).

charts/aws-efs-csi-driver/values.yaml
DavidXU12345: Done. Added comments with format description and examples above both values.

vmishra22 (approved):

vmishra22 (approved):

PR #1900, opened 2026-05-24
feat(helm): make updateStrategy configurable for node DaemonSet

### LINKED ISSUE #1898: [helm] Why is OnDelete updateStrategy not supported for the node DaemonSet?
## Current State

The [FAQ](https://github.com/kubernetes-sigs/aws-efs-csi-driver/blob/master/docs/faq.md#update-strategy) states that changing the update strategy (e.g. to `OnDelete`) is not supported. The Helm chart's [node-daemonset.yaml template](https://github.com/kubernetes-sigs/aws-efs-csi-driver/blob/master/charts/aws-efs-csi-driver/templates/node-daemonset.yaml) hardcodes `type: RollingUpdate` when `node.rollingUpdate` values are provided, with no option to configure the strategy type.

## Problem

When the `efs-csi-node` DaemonSet pod restarts for any reason (rolling update, OOM kill, node pressure, etc.), active EFS mount connections on that node break. This causes disruption to all workloads with mounted EFS volumes on the affected node.

With `RollingUpdate` as the only option, users have no way to control when the node pods restart during an upgrade — the rollout happens automatically and causes connection loss across nodes.

## Questions

1. **Why is `OnDelete` not supported as a configurable updateStrategy?** Is there a technical reason it's incompatible with the driver, or is this simply not yet implemented?

2. **What is the recommended approach for users to prevent EFS connection loss when the node DaemonSet pods restart?** The FAQ documents the limitation but doesn't provide guidance on mitigation.

3. **Would a PR to make `updateStrategy.type` configurable via Helm values be accepted?** This could default to `RollingUpdate` for backwards compatibility while allowing users who need zero-disruption upgrades to opt into `OnDelete`.

## Use Case

In clusters with long-running workloads that maintain active EFS mounts (e.g. Jupyter notebooks, batch jobs), any interruption to EFS connectivity causes user-visible failures. The `OnDelete` strategy would allow operators to coordinate node pod restarts with maintenance windows or node drains, avoiding unexpected connection loss.

DavidXU12345:
Fixes #1898

## Summary

Makes the DaemonSet `updateStrategy` and Deployment `strategy` fully configurable via Helm values, allowing users to set `OnDelete` strategy for the node DaemonSet to control when pods restart during upgrades.

## Changes

- **node-daemonset.yaml**: Replace hardcoded `RollingUpdate` with configurable `node.updateStrategy` object
- **controller-deployment.yaml**: Replace hardcoded `RollingUpdate` with configurable `controller.strategy` object  
- **values.yaml**: New `node.updateStrategy` (replaces `node.rollingUpdate`) and `controller.strategy` (replaces `controller.rollingUpdate`) with `type: RollingUpdate` as default

## Usage

To use `OnDelete` strategy for the node DaemonSet:
```yaml
node:
  updateStrategy:
    type: OnDelete
```

To use `RollingUpdate` with custom parameters:
```yaml
node:
  updateStrategy:
    type: RollingUpdate
    rollingUpdate:
      maxUnavailable: 20%
```

## Breaking Change

`node.rollingUpdate` is now `node.updateStrategy` and `controller.rollingUpdate` is now `controller.strategy`. Users who previously configured these values must migrate to the new structure.

## Testing

- `helm template` renders correctly for all scenarios (default, OnDelete, RollingUpdate with params)
- `make test` passes
- `make verify` passes

### DIFF
--- a/charts/aws-efs-csi-driver/templates/controller-deployment.yaml
+++ b/charts/aws-efs-csi-driver/templates/controller-deployment.yaml
@@ -15,11 +15,9 @@ spec:
   {{- if or (kindIs "float64" .Values.controller.revisionHistoryLimit) (kindIs "int64" .Values.controller.revisionHistoryLimit) }}
   revisionHistoryLimit: {{ .Values.controller.revisionHistoryLimit }}
   {{- end }}
-  {{- if .Values.controller.rollingUpdate }}
+  {{- with .Values.controller.strategy }}
   strategy:
-    type: RollingUpdate
-    rollingUpdate:
-      {{- toYaml .Values.controller.rollingUpdate | nindent 6 }}
+    {{- toYaml . | nindent 4 }}
   {{- end }}
   selector:
     matchLabels:
--- a/charts/aws-efs-csi-driver/templates/node-daemonset.yaml
+++ b/charts/aws-efs-csi-driver/templates/node-daemonset.yaml
@@ -13,11 +13,9 @@ spec:
   {{- if or (kindIs "float64" .Values.node.revisionHistoryLimit) (kindIs "int64" .Values.node.revisionHistoryLimit) }}
   revisionHistoryLimit: {{ .Values.node.revisionHistoryLimit }}
   {{- end }}
-  {{- if .Values.node.rollingUpdate }}
-  updateStrategy: 
-    type: RollingUpdate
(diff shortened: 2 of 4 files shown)


DavidXU12345:
Breaking change, need to wait for major version bump

PR #1915, opened 2026-06-29
fix: Resolve S3 Files mount target AZ-ID for DeleteVolume

### LINKED ISSUE #1913: `deleteAccessPointRootDir` fails for `s3files-ap` when IMDS is inaccessible from pods
/kind bug

**What happened?**

When using `s3files-ap` provisioning mode with `deleteAccessPointRootDir: true`, deleting a PV with `reclaimPolicy: Delete` leaves the PV permanently stuck in `Released` state. The `DeleteVolume` RPC fails because the controller attempts to mount the S3 Files filesystem via NFS to delete the root directory — but this mount requires querying IMDS to retrieve the AZ-ID, which is inaccessible from pods on clusters configured with `http_put_response_hop_limit: 1`.

As a result, `deleteAccessPointRootDir: true` actively breaks `reclaimPolicy: Delete` for `s3files-ap`: neither the root directory nor the access point is deleted, and the PV is never cleaned up.

The PV is deleted, the S3 Files access point is deleted, and the root directory (S3 objects under the access point prefix) is deleted.

1. Configure EKS nodes with `http_put_response_hop_limit: 1` (standard IMDSv2 hardening that blocks pods from accessing IMDS).

2. Enable `deleteAccessPointRootDir: true` on the controller (via Helm or EKS add-on `configuration_values`):
```json
{
  "controller": {
    "deleteAccessPointRootDir": true
  }
}
```

3. Create a StorageClass with `provisioningMode: s3files-ap` and `reclaimPolicy: Delete`:
```yaml
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  name: efs-s3
parameters:
  basePath: /csi
  directoryPerms: "777"
  ensureUniqueDirectory: "false"
  fileSystemId: fs-0123456789abcdef0
  gidRangeEnd: "200000"
  gidRangeStart: "1000"
  provisioningMode: s3files-ap
  subPathPattern: ${.PV.name}
provisioner: efs.csi.aws.com
reclaimPolicy: Delete
volumeBindingMode: Immediate
```

4. Create a PVC, and wait that it is provisioned

5. Delete the PVC.

6. Observe the PV stuck in `Released` state:
```
NAME                                       CAPACITY   ACCESS MODES   RECLAIM POLICY   STATUS     CLAIM                  STORAGECLASS
pvc-xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx   5Gi        RWX            Delete           Released   default/s3files-test   efs-s3
```

**Anything else we need to know?**

This problem is **specific to `s3files-ap`** and does not affect `efs-ap`. EFS uses a regional DNS endpoint (`<fs-id>.efs.<region>.amazonaws.com`) that does not require the AZ-ID. S3 Files uses an AZ-specific DNS format (`{az_id}` in `dns_name_format`) which requires IMDS to resolve the current instance's AZ-ID at mount time.

`http_put_response_hop_limit: 1` is a widely-recommended EKS security hardening (see [AWS docs](https://docs.aws.amazon.com/eks/latest/userguide/best-practices-security.html)). This makes the limitation broadly impactful for production clusters.

**Proposed fix:** Instead of mounting the S3 Files filesystem (NFS) to delete the root directory, the controller should use the S3 API directly (`s3:ListObjectsV2` + `s3:DeleteObjects`) on the objects under the access point prefix. This would:
- Not require IMDS or an NFS mount from the controller pod
- Work regardless of `http_put_response_hop_limit` configuration
- Be consistent with how S3 Files data is actually stored (S3 objects)

**Environment**

- Kubernetes version: v1.34
- Driver version: v3.2.0 (EKS managed add-on)
- AWS region: eu-west-3 (Paris)
- Credential mechanism: EKS Pod Identity
- Node IMDS: `http_put_response_hop_limit: 1`

**Debug logs**

Events on the PV after PVC deletion:

```
Warning  VolumeFailedDelete  20s (x7 over 116s)  efs.csi.aws.com_efs-csi-controller-xxxxxxxxxx-xxxxx_xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx  rpc error: code = Internal desc = Could not mount "fs-0123456789abcdef0" at "/var/lib/csi/pv/fsap-0123456789abcdef0": mount failed: exit status 1
Mounting command: mount
Mounting arguments: -t s3files -o tls,iam fs-0123456789abcdef0 /var/lib/csi/pv/fsap-0123456789abcdef0
Output: Cannot retrieve AZ-ID from instance metadata service (IMDS). Ensure IMDS is reachable and that the instance has the correct IAM permissions. This is required for S3Files mounts using {az_id} in dns_name_format. You can also specify the AZ ID directly using the mount option: -o azid=<az-id>.
```

YangjinanHu:

Bug fix

**What testing is done?** 
Fixes #1913

### DIFF
--- a/pkg/cloud/cloud.go
+++ b/pkg/cloud/cloud.go
@@ -112,6 +112,7 @@ type S3Files interface {
 	DeleteAccessPoint(context.Context, *s3files.DeleteAccessPointInput, ...func(*s3files.Options)) (*s3files.DeleteAccessPointOutput, error)
 	ListAccessPoints(context.Context, *s3files.ListAccessPointsInput, ...func(*s3files.Options)) (*s3files.ListAccessPointsOutput, error)
 	ListFileSystems(context.Context, *s3files.ListFileSystemsInput, ...func(*s3files.Options)) (*s3files.ListFileSystemsOutput, error)
+	ListMountTargets(context.Context, *s3files.ListMountTargetsInput, ...func(*s3files.Options)) (*s3files.ListMountTargetsOutput, error)
 }
 
 type Cloud interface {
@@ -701,7 +702,41 @@ func (c *cloud) DescribeMountTargets(ctx context.Context, fileSystemId, azName s
 			IPAddress:     *mountTarget.IpAddress,
 		}, nil
 	case util.FileSystemTypeS3Files:
-		return nil, fmt.Errorf("DescribeMountTargets is not implemented for s3files in CSI Driver")
+		var nextToken *string
+		for {
+			listMtInput := &s3files.ListMountTargetsInput{
+				FileSystemId: &fileSystemId,
+				NextToken:    nextToken,
+			}
+			klog.V(5).Infof("Calling ListMountTargets with input: %+v", *listMtInput)
+			res, err := c.s3files.ListMountTargets(ctx, listMtInput, func(o *s3files.Options) {
+				o.Retryer = c.rm.describeMountTargetsRetryer
+			})
+			if err != nil {
+				if isAccessDenied(err) {
+					return nil, ErrAccessDenied
+				}
+				if isS3FilesFileSystemNotFound(err) {
+					return nil, ErrNotFound
+				}
+				return nil, fmt.Errorf("List S3 Files Mount Targets failed: %v", err)
+			}
+			for _, mt := range res.MountTargets {
+				if mt.Status != s3filestypes.LifeCycleStateAvailable || mt.AvailabilityZoneId == nil {
+					continue
+				}
+				target := &MountTarget{AZId: *mt.AvailabilityZoneId}
+				if mt.MountTargetId != nil {
+					target.MountTargetId = *mt.MountTargetId
+				}
+				return target, nil
+			}
+			if res.NextToken == nil {
+				break
+			}
+			nextToken = res.NextToken
+		}
+		return nil, fmt.Errorf("No available mount targets for S3 Files file system %v", fileSystemId)
 	default:
 		return nil, fmt.Errorf("Unsupported fsType: %v", err)
 	}
--- a/pkg/cloud/cloud_test.go
+++ b/pkg/cloud/cloud_test.go
@@ -1792,18 +1792,59 @@ func TestDescribeFileSystemS3Files(t *testing.T) {
 
 func TestDescribeMountTargetsS3Files(t *testing.T) {
 	var (
-		fileSystemId = "fs-abcd1234"
-		az           = "us-east-1a"
+		fileSystemId  = "fs-abcd1234"
+		az            = "us-east-1a"
+		azId          = "use1-az1"
+		mountTargetId = "fsmt-abcd1234"
 	)
 
 	testCases := []struct {
-		name        string
-		expectError errtyp
+		name         string
+		listOutput   *s3files.ListMountTargetsOutput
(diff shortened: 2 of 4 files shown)


pkg/cloud/cloud.go
```diff
@@ -701,7 +702,46 @@ func (c *cloud) DescribeMountTargets(ctx context.Context, fileSystemId, azName s
...
+				if isAccessDenied(err) {
+					return nil, ErrAccessDenied
+				}
+				if isS3FilesFileSystemNotFound(err) {
+					return nil, ErrNotFound
+				}
+				return nil, fmt.Errorf("List S3 Files Mount Targets failed: %v", err)
+			}
+			for _, mt := range res.MountTargets {
+				if mt.Status != s3filestypes.LifeCycleStateAvailable || mt.AvailabilityZoneId == nil {
```
DavidXU12345, line 725: Could you explain why we need to check `mt.Status == s3filestypes.LifeCycleStateAvailable`?

pkg/cloud/cloud.go
```diff
@@ -701,7 +702,46 @@ func (c *cloud) DescribeMountTargets(ctx context.Context, fileSystemId, azName s
...
+					continue
+				}
+				target := &MountTarget{AZId: *mt.AvailabilityZoneId}
+				if mt.MountTargetId != nil {
+					target.MountTargetId = *mt.MountTargetId
+				}
+				if mt.Ipv4Address != nil {
```
DavidXU12345: Are we sure we prefer Ipv4 over Ipv6? Ideally we should save both and let downstream decide (probably by connecting using IPv6 first and fallback to IPv4 if IPv6 not working)

pkg/cloud/cloud.go
```diff
@@ -701,7 +702,46 @@ func (c *cloud) DescribeMountTargets(ctx context.Context, fileSystemId, azName s
...
+					target.IPAddress = *mt.Ipv4Address
+				} else if mt.Ipv6Address != nil {
+					target.IPAddress = *mt.Ipv6Address
+				}
+				return target, nil
```
DavidXU12345, line 732: looks like we would return the first available mount target. Its `azid` may be different from what IMDS will return, which may have performance implications. Shall we try IMDS first (which is also the current behavior) for `azid` and fallback to this method only if IMDS fails?

pkg/cloud/cloud.go
YangjinanHu: DeleteVolume mounts the filesystem to delete the AP root dir, so the AZ-ID needs to belong to a mount target that can actually serve NFS, and only available ones can.

pkg/cloud/cloud.go
YangjinanHu: IMDS would return the controller node's AZ, which may have no S3 Files mount target and IMDS could be unreachable. IMDS-first-then-fallback flow could force a failing mount attempt on every delete in those two cases I think would be a longer wait than potential cross-az mount latency.

pkg/cloud/cloud.go
YangjinanHu: IP isn't used by this path (only AZ-ID is); I can drop the IP population entirely

YangjinanHu: Updated

PR #1924, opened 2026-07-20
feat: allow setting environment variables for csi-provisioner

jandersen-plaid:

new feature

We sometimes come across issues with the csi-provisioner failing to connect to the upstream API server due to load balancer timeouts. To tune this, we use the HTTP2 ping settings that are exposed via environment variables in order to make sure streams refresh appopriately.

This updates the chart to allow the csi-provisioner to set environment variables which, in turn, allows us to get to 0 load balancer timeouts.

**What testing is done?** 

I have verified that the helm chart renders with the new values set for the env and without (in the absent case, the default values take over and will template out nothing).


DavidXU12345:
hi @jandersen-plaid please do not bump version in your PR. After merging your PR, we will bump version when doing release.

jandersen-plaid:
> hi @jandersen-plaid please do not bump version in your PR. After merging your PR, we will bump version when doing release.

Done! Thank you for taking a look at the changes.

DavidXU12345 (approved):

PR #1926, opened 2026-07-21
fix: Validate access point filesystem ownership in DeleteVolume

DavidXU12345:

**What testing is done?**

### DIFF
--- a/pkg/cloud/cloud.go
+++ b/pkg/cloud/cloud.go
@@ -388,6 +388,12 @@ func (c *cloud) DescribeAccessPoint(ctx context.Context, accessPointId string, f
 			return nil, fmt.Errorf("DescribeAccessPoint failed. Expected exactly 1 access point in DescribeAccessPoint result. However, received %d access points", len(accessPoints))
 		}
 
+		// Verify the access point belongs to the expected filesystem.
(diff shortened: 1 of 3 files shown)


wangnyue (approved):

pkg/cloud/cloud.go
```diff
@@ -388,6 +388,12 @@ func (c *cloud) DescribeAccessPoint(ctx context.Context, accessPointId string, f
 			return nil, fmt.Errorf("DescribeAccessPoint failed. Expected exactly 1 access point in DescribeAccessPoint result. However, received %d access points", len(accessPoints))
 		}
 
+		// Verify the access point belongs to the expected filesystem.
+		if *accessPoints[0].FileSystemId != fileSystemId {
+			return nil, fmt.Errorf("access point %s belongs to filesystem %s, not %s",
+				accessPointId, *accessPoints[0].FileSystemId, fileSystemId)
+		}
```
ABBennett, line 395: Could we avoid exposing the actual owning filesystem ID here? Saying it doesn’t belong to the expected filesystem should be enough.

DavidXU12345: Done

PR #1927, opened 2026-07-21
fix(controller): Reject empty PVC name when reuseAccessPoint is set

DavidXU12345:
When reuseAccessPoint is enabled the reuse client-token is derived from the PVC name only. If the external-provisioner runs without --extra-create-metadata the PVC name is not passed to the driver, so hashing it produces sha256("") for every reuse-enabled volume, collapsing them all onto a single shared access point.

Reject an empty PVC name in the reuse path with InvalidArgument instead of silently colliding. The token formula is unchanged (name only), so this is non-breaking for standard installs, which enable --extra-create-metadata by default.

Also document the cross-namespace data-sharing behavior of reuseAccessPoint and warn against enabling it on a StorageClass shared across mutually-distrusting tenants.

**What testing is done?**


DavidXU12345:
Approved from @aliyajo , going to merge it manually

PR #1934, opened 2026-07-24
docs: document IAM requirement for controller custom tags

### LINKED ISSUE #1921: Missing IAM policy documentation for controller custom tags
**What happened?**

The EFS CSI Driver Helm chart allows configuring custom tags on the controller via the `--tags` flag:

```yaml
controller:
  tags:
    environment: prod
    cluster-name: my-cluster
```

However, when these tags are enabled, dynamic access point creation fails. PVCs remain stuck in pending state, and controller logs show EFS permission errors.

**Root Cause:**

The default AWS Managed Policy for EFS CSI Driver (`AmazonEFSCSIDriverPolicy`) documented in [install.md](https://github.com/kubernetes-sigs/aws-efs-csi-driver/blob/master/docs/install.md#set-up-driver-permission) includes an `AllowCreateAccessPoint` statement with:

```json
"Condition": {
  "ForAllValues:StringEquals": {
    "aws:TagKeys": "efs.csi.aws.com/cluster"
  }
}
```

This condition restricts allowed tag keys to **only** `efs.csi.aws.com/cluster`. Custom tags cause the AP creation to fail with a permission denied error.

**Impact:**

- Users enabling `controller.tags` cannot create dynamic access points without AWS IAM policy modifications
- The Helm chart's `controller.tags` configuration parameter is unusable with default IAM policies

**Suggested Solutions**

1. **Documentation:** Update `docs/install.md` to explicitly warn that custom tags require a custom IAM policy, with an example of the required condition
2. **Helm values documentation:** Add a comment in `charts/aws-efs-csi-driver/values.yaml` under `controller.tags` explaining the IAM requirement

DavidXU12345:
Custom tags configured via the controller --tags flag (controller.tags in the Helm chart) fail dynamic access point creation when the controller uses the AWS-managed AmazonEFSCSIDriverPolicy. That policy restricts allowed tag keys to efs.csi.aws.com/cluster via a ForAllValues:StringEquals aws:TagKeys condition, so any custom key causes CreateAccessPoint to be denied and PVCs to remain Pending.

Document the requirement and a working policy alternative in the install guide, add a note to the tags parameter reference, and flag it in the Helm values comment under controller.tags.

Fix https://github.com/kubernetes-sigs/aws-efs-csi-driver/issues/1921

**What testing is done?**


DavidXU12345:
Approved from @aliyajo

Issue #1937, opened 2026-08-13
Data race in `volumeIdCounter` map causes `fatal error: concurrent map read and map write` panic when `vol-metrics-opt-in=true`


smtribal:
/kind bug

### What happened?

The `efs-csi-node` pod panics and crashes with:

```
fatal error: concurrent map read and map write
```

The crash happens during normal `NodePublishVolume` / `NodeUnpublishVolume` activity — in our case immediately after a `NodeUnpublishVolume` call that was evicting a volume from the metrics cache:

```
I0724 18:53:38.367855       1 mount_linux.go:402] Unmounting /var/lib/kubelet/pods/<pod-uid>/volumes/kubernetes.io~csi/<vol>/mount
I0724 18:53:38.380578       1 node.go:319] NodeUnpublishVolume: /var/lib/kubelet/pods/<pod-uid>/volumes/kubernetes.io~csi/<vol>/mount unmounted
I0724 18:53:38.380606       1 node.go:327] Evicting vol ID: <fs-id>, vol path : /var/lib/kubelet/pods/<pod-uid>/volumes/kubernetes.io~csi/<vol>/mount from cache
I0724 18:53:38.380614       1 node.go:319] NodeUnpublishVolume: /var/lib/kubelet/pods/<pod-uid>/volumes/kubernetes.io~csi/<vol>/mount unmounted
fatal error: concurrent map read and map write
```

The daemonset then restarts (CrashLoopBackOff / repeated restarts) whenever multiple pods on the same node mount/unmount EFS volumes at roughly the same time.

Full panic stack trace (identifiers redacted):

```
fatal error: concurrent map read and map write
goroutine 270346 [running]:
internal/runtime/maps.fatal({0x22bbd27?, 0x584fcd?})
        /go/pkg/mod/golang.org/toolchain@v0.0.1-go1.26.4.linux-amd64/src/runtime/panic.go:1181 +0x18
github.com/kubernetes-sigs/aws-efs-csi-driver/pkg/driver.(*Driver).NodeUnpublishVolume(0x2c6134038600, {0x1?, 0x0?}, 0x2c6136111b40)
        /go/src/github.com/kubernetes-sigs/aws-efs-csi-driver/pkg/driver/node.go:324 +0x699
github.com/container-storage-interface/spec/lib/go/csi._Node_NodeUnpublishVolume_Handler.func1(...)
[...]

PR #1938, opened 2026-08-14
fix: synchronize access to volumeIdCounter map in node.go

wangnyue:
volumeIdCounter is a package-level map read, written, and deleted from NodePublishVolume and NodeUnpublishVolume, both of which can run concurrently as independent goroutines per gRPC call when vol-metrics-opt-in is enabled. Unsynchronized concurrent access trips Go's runtime map-race detector and crashes efs-csi-node with a fatal error.

Add a dedicated sync.Mutex (volumeIdCounterMu) and lock it around the read-modify-write critical sections in both handlers.

Add a unit test that exercises concurrent NodePublishVolume and NodeUnpublishVolume calls across goroutines (with volMetricsOptIn enabled) and verified it reproduces the race without the fix and passes cleanly under go test -race with the fix applied.

bug fix 

  Fixes a `fatal error: concurrent map read and map write` panic in `efs-csi-node`, reported in #1937.

  `volumeIdCounter` in `pkg/driver/node.go` is a package-level `map[string]int` used to
  track per-volume publish/unpublish reference counts when volume metrics are enabled
  (`--vol-metrics-opt-in=true`). It was read/written/deleted directly from both
  `NodePublishVolume` and `NodeUnpublishVolume` with no synchronization. Since kubelet
  issues these RPCs concurrently (as independent goroutines) whenever multiple pods on a
  node mount/unmount EFS volumes around the same time, this triggers Go's runtime map-race
  detector and crashes the process, causing the daemonset to CrashLoopBackOff.

  This only affects deployments with `node.volMetricsOptIn: true` (default is `false`,
  where `volumeIdCounter` is never touched).

  The fix adds a dedicated `sync.Mutex` (`volumeIdCounterMu`) guarding every read, write,
  and delete of `volumeIdCounter` in `NodePublishVolume` and `NodeUnpublishVolume`. This
  mirrors the existing pattern in `vol_statter.go`, which already protects its own cache
  maps with a `sync.RWMutex`. The lock is held across the call into
  `d.volStatter.removeFromCache(...)` in the eviction path, which is safe since
  `VolStatter` uses its own independent mutex and never re-enters `volumeIdCounterMu`.

  Relates to #1937.

**What testing is done?** 
  - Added `TestNodePublishUnpublishVolumeConcurrent` in `pkg/driver/node_test.go`, which
    spins up 50 goroutines calling `NodePublishVolume`/`NodeUnpublishVolume` concurrently
    (with `volMetricsOptIn=true`) across a mix of shared and distinct volume IDs.
  - Verified the test reproduces the reported data race on `volumeIdCounter` when run with
    `go test -race` against the code *before* this fix, and passes cleanly *after* the fix.
  - `go build ./...` — passes.
  - `go vet ./...` — no issues.
  - `go test -race -count=1 ./pkg/...` — all packages pass, including all pre-existing
    `TestNodePublishVolume`/`TestNodeUnpublishVolume` subtests plus the new concurrency test.
  - `gofmt -s -l` on changed files — no issues.

### DIFF
--- a/pkg/driver/node.go
+++ b/pkg/driver/node.go
@@ -26,6 +26,7 @@ import (
 	"regexp"
 	"strconv"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/container-storage-interface/spec/lib/go/csi"
@@ -44,9 +45,14 @@ var (
 		csi.VolumeCapability_AccessMode_SINGLE_NODE_WRITER,
 		csi.VolumeCapability_AccessMode_MULTI_NODE_MULTI_WRITER,
 	}
-	volumeIdCounter  = make(map[string]int)
-	supportedFSTypes = []string{util.FileSystemTypeEFS.String(), util.FileSystemTypeS3Files.String(), ""}
-	hexSuffixRegex   = regexp.MustCompile(`^[0-9a-f]{8,40}$`)
+	volumeIdCounter = make(map[string]int)
+	// volumeIdCounterMu guards concurrent access to volumeIdCounter, which is
+	// read/written/deleted from NodePublishVolume and NodeUnpublishVolume.
+	// Those handlers run concurrently as independent goroutines per gRPC
+	// call, so the map must be protected against concurrent access.
+	volumeIdCounterMu sync.Mutex
+	supportedFSTypes  = []string{util.FileSystemTypeEFS.String(), util.FileSystemTypeS3Files.String(), ""}
+	hexSuffixRegex    = regexp.MustCompile(`^[0-9a-f]{8,40}$`)
 )
 
 const (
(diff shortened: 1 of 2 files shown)


pkg/driver/node.go
```diff
@@ -267,11 +273,13 @@ func (d *Driver) NodePublishVolume(ctx context.Context, req *csi.NodePublishVolu
 
 	//Increment volume Id counter
 	if d.volMetricsOptIn {
+		volumeIdCounterMu.Lock()
```
samuhale, line 276: Is there a reason we want to use a mutex rather than making volumeIdCounter a sync.Map? Seems to me like sync.Map is better optimized

wangnyue: I think sync.Map is more for read heavy map. It can work on read + increment path thread safe. But it cannot make decrement/evict/delete path thread safe as we need to do two steps: d.volStatter.removeFromCache(req.VolumeId) and volumeIdCounter.delete/decrement

samuhale (approved):

Issue #1936, opened 2026-08-11
Startup taint is removed before the driver is registered with the kubelet


fabn:
/kind bug

**What happened?**

With `efs.csi.aws.com/agent-not-ready:NoExecute` configured as a Karpenter `startupTaint`, pods that mount EFS volumes on a freshly provisioned node still fail with:

```
MountVolume.MountDevice failed for volume "pvc-..." : kubernetes.io/csi:
attacher.MountDevice failed to create newCsiDriverClient:
driver name efs.csi.aws.com not found in the list of registered CSI drivers
```

Reproduced on **7 out of 7** consecutive node provisionings.

The taint itself works at the scheduling layer — for a few seconds the scheduler correctly reports `node(s) had untolerated taint(s)`. The problem is that the driver removes the taint *before* CSI registration with the kubelet has completed, so the pod is admitted into a window where the kubelet still does not know about the driver. The taint does not close the race, it just shifts it by a few seconds.

Observed sequence, timestamps taken from Kubernetes events (relative to node Ready):

```
T+0s    node Ready
T+5s    FailedScheduling: node(s) had untolerated taint(s)   <- taint present, doing its job
T+10s   Scheduled                                            <- taint removed
T+12s   FailedMount: driver name efs.csi.aws.com not found in the list of registered CSI drivers
T+23s   container Started                                    <- registration finally complete
```

**What you expected to happen?**

The startup taint should not be removed until the driver is registered with the kubelet, so that a pod scheduled after removal can mount immediately.

**How to reproduce it (as minimally and precisely as possible)?**

1. Create a Karpenter `NodePool` with `spec.template.spec.startupTaints: [{key: efs.csi.aws.com/agent-not-ready, effect: NoExecute}]`.
2. Add a persistent taint to the same pool so only the test workload can land there.
3. Deploy a single-replica pod mounting an EFS-backed PVC, tolerating only the persistent taint (deliberately **not** the startup taint).
4. Delete the `NodeClaim` to force a fresh node, and watch the pod's events.

Each fresh node reproduces the `FailedMount`. Note that deleting the *pod* is not sufficient — it reschedules onto the existing node, which is already past startup. The `NodeClaim` has to be deleted so a genuinely new node is provisioned.

**Anything else we need to know?**

[...]

Issue #1613, opened 2025-03-28
Startup taint not actually getting removed


gnadaban:
/kind bug

**What happened?**

When some deployments scale up dozens of new pods, and the cluster is grown rapidly to accomodate, EFS CSI node pods don't remove the startup taint on some nodes.

According to logs in the Pod, the taint should have been removed.

This essentially keeps nodes in a Ready but unusable state, which makes Pods that would get scheduled onto them stay stuck in a Pending state.

**What you expected to happen?**

Startup taint removal works reliably, and is not only tried once, but periodically.

**How to reproduce it (as minimally and precisely as possible)?**

Use Karpenter on EKS, grow a deployment by dozens of replicas to be scheduled on separate nodes to trigger quick growth.

**Anything else we need to know?**:

Deleting EFS pods solves the taint removal, which tells me that taint removal is only done once on startup.

**Environment**
- Kubernetes version (use `kubectl version`): 
Client Version: v1.31.7
Kustomize Version: v5.4.2
Server Version: v1.32.2-eks-bc803b4

- Driver version: v2.1.7

treyhyde:
FWIW we see the taint fail to clear on new nodes every month or so despite the log claiming it cleared it.

Issue #1491, opened 2024-10-28
Taint removal race condition


ricardov1:
/kind bug

**What happened?**
When spinning up an instance in a cluster where both the ebs-csi-driver and the efs-csi-driver are deployed, occasionally (about 10% of the time) either of the two taints are not removed, and the node is unable to accept any pods due to the lingering taint.

**What you expected to happen?**
After successful daemonset startup, the taint should be removed always.

**How to reproduce it (as minimally and precisely as possible)?**
This appears to be caused by a race condition between the efs and ebs daemonsets (described below) so it is difficult to reproduce. The best way to increase the likelihood of reproducing the issue without launching and relauching a single instance is to spin up many (50+) instances where both daemonsets run simultaneously on instance startup. If you encounter the race condition on one of those nodes, the node will have either the efs or ebs taint on it and the pod for the corresponding lingering taint will be missing the `Removed taint(s) from local node` log.

**Anything else we need to know?**:
This issue appears to be caused by a race condition where the daemonset incorrectly determines that the `efs.csi.aws.com/agent-not-ready` taint has been removed. [This block](https://github.com/kubernetes-sigs/aws-efs-csi-driver/blob/master/pkg/driver/node.go#L489-L501) fetches the taints on the node and uses the length of the list of taints on the node and the length of `taintsToRemove` to determine whether the step of removing the taint is needed. When we run into this issue, we suspect that the **ebs-csi-driver** "beats" the efs-csi-driver to the taint removal and when the lengths of the slices are compared, they are equal even though it was the `ebs.csi.aws.com/agent-not-ready` taint that was removed, not the `efs.csi.aws.com/agent-not-ready` taint.

related issue on ebs-csi-driver repo: https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/2199

**Environment**
- Kubernetes version (use `kubectl version`): 1.29.8
- Driver version: 1.7.5

**Please also attach debug logs to help us better diagnose**

- Instructions to gather debug logs can be found [here](https://github.com/kubernetes-sigs/aws-efs-csi-driver/tree/master/troubleshooting)

mskanth972:
[...]

Issue #1320, opened 2024-04-19
Startup Taint is sometimes not removed


dpiddock:
/kind bug

**What happened?**
We use Karpenter for the cluster node scaler. We recently implemented the startup taint for the efs driver. Since then we have occasionally seen nodes stuck with unscheduleable pods. Upon investigation it is caused by the `efs.csi.aws.com/agent-not-ready:NoExecute` taint still being present on the node despite the `efs-csi-node` pod running correctly.

The `efs-plugin` container has this log line:
```
E0419 06:03:32.149207       1 driver.go:134] "Unexpected failure when attempting to remove node taint(s)" err="the server rejected our request due to an error in our request"
```

**What you expected to happen?**
`efs-csi-node` pod to successfully remove the node taint so that other pods can be scheduled.

**How to reproduce it (as minimally and precisely as possible)?**
We don't currently know. It's a rare and intermittent problem.

**Anything else we need to know?**:
I managed to find the failed request in the audit logs. At the same time there was a request being processed to add the taint `node.kubernetes.io/not-ready:NoExecute` by node-controller. This could be a classic race condition with other parts of the system? Although this is a sample size of 1.

I attach the two entries:
[efs-csi.json](https://github.com/kubernetes-sigs/aws-efs-csi-driver/files/15040596/efs-csi.json)
[node-controller.json](https://github.com/kubernetes-sigs/aws-efs-csi-driver/files/15040597/node-controller.json)

**Environment**
- Kubernetes version (use `kubectl version`): `v1.29.1-eks-b9c9ed7`
- EKS 1.29
- Driver version: `v1.7.6`
- EKS add-on: `v1.7.6-eksbuild.2`

**Please also attach debug logs to help us better diagnose**

- Instructions to gather debug logs can be found [here](https://github.com/kubernetes-sigs/aws-efs-csi-driver/tree/master/troubleshooting)

[results.tgz](https://github.com/kubernetes-sigs/aws-efs-csi-driver/files/15038180/results.tgz)

seanzatzdev-amazon:
Hi @dpiddock , thank you for bringing this to our attention. We are working together with the author of the following PR to address this issue: https://github.com/kubernetes-sigs/aws-efs-csi-driver/pull/1287

Please let us know if you have any further questions or concerns.

mteodori:
is this same as https://github.com/kubernetes-sigs/aws-efs-csi-driver/issues/1273 ?

[...]

PR #1939, opened 2026-08-14
feat: gate startup taint removal on CSINode driver registration

wangnyue:
Add checkDriverRegistration function that verifies the driver is registered in the CSINode object with Allocatable set before removing the efs.csi.aws.com/agent-not-ready taint. This prevents FailedMount errors on freshly provisioned nodes where the taint was previously removed before csi-node-driver-registrar completed.

Changes:
- Add checkDriverRegistration() in pkg/driver/node.go
- Modify tryRemoveNotReadyTaintUntilSucceed() to accept a registration check function with exponential backoff
- Update Driver.Run() call site with registration gate
- Add RBAC warning logs for forbidden/unauthorized errors
- Add csinodes get permission to node ClusterRole (helm + kustomize)
- Add StorageV1/CSINode mock for testing
- Add TestCheckDriverRegistration with table-driven test cases
- Update TestTryRemoveNotReadyTaintUntilSucceed for new signature

Bug fix

When `efs.csi.aws.com/agent-not-ready:NoExecute` is configured as a Karpenter startup taint, the taint is removed before the driver has actually registered with kubelet's CSINode object. This creates a window where pods are scheduled to the node but kubelet cannot mount their EFS volumes because it doesn't yet know about the `efs.csi.aws.com` driver:

```
FailedMount: driver name efs.csi.aws.com not found in the list of registered CSI drivers
```

Reproduced on 7/7 consecutive node provisionings. The taint shifts the race by a few seconds but doesn't close it. The root cause is that `removeNotReadyTaint` in `pkg/driver/node.go` patches out the taint as soon as the driver process starts — it only checks that the K8s API is reachable, not that `csi-node-driver-registrar` has completed registration with kubelet. 

`aws-ebs-csi-driver` had the same bug (kubernetes-sigs/aws-ebs-csi-driver#1945) and fixed it in kubernetes-sigs/aws-ebs-csi-driver#1949 by gating on the CSINode object. That fix was never ported to this driver.

This PR ports the same pattern:

1. **`checkDriverRegistration()`** — Queries the `CSINode` object via `StorageV1().CSINodes().Get()`, confirms the driver name (`efs.csi.aws.com`) is listed in `Spec.Drivers` with `Allocatable` set. Only then does it allow taint removal to proceed.

2. **Exponential backoff** — `tryRemoveNotReadyTaintUntilSucceed` now accepts an initial delay (2s), a starting interval (1s), and a max interval (10s). It calls the registration check first; only when that passes does it attempt taint removal. This avoids API spam on fresh nodes while registration is still in progress.

3. **RBAC warning** — If the CSINode `Get` fails with Forbidden/Unauthorized, a prominent `klog.Warning` tells the operator exactly what permission is missing and that the taint will NOT be removed until it's granted. This avoids silently creating a "taint never removed" scenario (the opposite failure mode tracked in #1320/#1491/#1613) when upgrading without updating RBAC.

4. **RBAC manifests** — Added `get` on `csinodes.storage.k8s.io` to the node ClusterRole in both the Helm chart and kustomize base manifests.

This is NOT a duplicate of #1320, #1491, or #1613 (which describe the taint *never* being removed). Here the taint is removed reliably — just too early to be useful.

Relates to #1936.

**What testing is done?** 
- Added `TestCheckDriverRegistration` with 7 table-driven cases:
  - `CSI_NODE_NAME` not set → error
  - K8s client creation fails → error
  - CSINode Get returns NotFound → error
  - CSINode Get returns Forbidden → error + warning log
  - CSINode exists but driver not listed → error
  - CSINode exists, driver listed, `Allocatable` nil → error
  - CSINode exists, driver listed, `Allocatable` set → success
- Extended `TestTryRemoveNotReadyTaintUntilSucceed` with 4 scenarios:
  - Registration fails N times then succeeds → taint removal called only after gate passes
  - Registration succeeds immediately, taint removal fails then succeeds on retry
  - Registration fails permanently (capped iteration test)
  - Registration + removal both succeed on first try
- All existing `TestRemoveNotReadyTaint` cases pass unchanged.
- `go build ./...` — passes.
- `go vet ./...` — no issues.
- `go test -race -count=1 ./pkg/...` — all packages pass (cloud, driver, util).


pkg/driver/node.go
```diff
@@ -608,6 +609,38 @@ type JSONPatch struct {
...
+// by checking that the CSINode exists, lists our driver name, and has Allocatable set.
+// Returns nil when registered, error otherwise.
+func checkDriverRegistration(k8sClient cloud.KubernetesAPIClient, driverName string) error {
+	nodeName := os.Getenv("CSI_NODE_NAME")
+	if nodeName == "" {
+		return fmt.Errorf("CSI_NODE_NAME not set")
+	}
+	clientset, err := k8sClient()
+	if err != nil {
+		return fmt.Errorf("failed to create kubernetes client: %w", err)
```
samuhale: nit: Capitalization

samuhale (approved):

sma-nextdoor:
@wangnyue heads up — #1966 reports a regression from this change: the registration check requires `CSINode.spec.drivers[].allocatable` to be set, but kubelet only populates that field for drivers that report a `max_volumes_per_node` limit in `NodeGetInfo`, which the EFS driver never does. So the check can never pass, every `efs-csi-node` pod polls `GET csinodes` every 10s indefinitely, and the `agent-not-ready` startup taint would never be removed if used. Details and repro in the issue.

PR #1946, opened 2026-08-21
fix: restore DNS-name volume handles in static PVs

wangnyue:
Bug fix (regression introduced in v3.0.0).

v3.4.1 hardened `NodePublishVolume` against mount-option injection by strictly
validating the file system ID in a volume handle against `fs-[0-9a-f]{8,40}`.
That strict check also rejected the mount-target **DNS-name** form of the ID
that customers use in static PV `volumeHandle`s, e.g.
`fs-9919e11b.efs.us-east-1.amazonaws.com`. As a result, static PVs that worked
before v3.4.1 began failing at `NodePublishVolume` after upgrading.

This PR relaxes `isValidFileSystemId` (`pkg/driver/node.go`) to accept **either**:

- the bare id — still strictly validated as `fs-[0-9a-f]{8,40}`, or
- the DNS-name form — validated by a **structural** regex (`dnsFileSystemIdRegex`)
  that checks the shape of the name instead of enumerating known service domains:

  ```
  ^([a-z0-9][a-z0-9-]*\.)?fs-[0-9a-f]{8,40}\.[a-z0-9][a-z0-9.-]*[a-z0-9]$
  ```

An optional leading AZ label is allowed, the `fs-<hex>` id label stays strictly
validated, and the rest of the domain is constrained to the DNS charset
`[a-z0-9.-]` only. Because commas, spaces, slashes, and `=` cannot appear, a
DNS-name `volumeHandle` still cannot be used to inject additional mount options
into the source passed to efs-utils — the injection-safety property added in
v3.4.1 is preserved.

Covered forms:

- EFS bare: `fs-9919e11b.efs.us-east-1.amazonaws.com`
- EFS AZ-prefixed: `us-east-1a.fs-9919e11b.efs.us-east-1.amazonaws.com`
- S3 Files: `use1-az1.fs-0123456abcdef0189.s3files.us-east-1.on.aws`
- China / FIPS variants

The fsid is returned unchanged, so the efs-utils mount source is byte-for-byte
identical to pre-v3.4.1 behavior. This is a strict widening of accepted input;
there is no behavior change for bare-id volume handles and no API change.

**What testing is done?** 
Unit tests in `pkg/driver/node_test.go`:

- `TestIsValidFileSystemId` — positive cases (bare id, EFS bare, EFS
  AZ-prefixed, S3 Files, China, FIPS) and negative cases (injected mount
  options, too-short id label, non-hex id label, empty, prefix-only).
- `TestParseVolumeIdDnsName` — a DNS-name `volumeHandle` parses successfully
  and the fsid round-trips unchanged so the mount source is unaltered.

### DIFF
--- a/pkg/driver/node.go
+++ b/pkg/driver/node.go
@@ -57,6 +57,21 @@ var (
 	volumeIdCounterMu sync.Mutex
 	supportedFSTypes  = []string{util.FileSystemTypeEFS.String(), util.FileSystemTypeS3Files.String(), ""}
 	hexSuffixRegex    = regexp.MustCompile(`^[0-9a-f]{8,40}$`)
+	// dnsFileSystemIdRegex matches the DNS-name form of a file system ID used in
+	// static PV volumeHandles. It validates the structure of the name rather than
+	// enumerating known domains, covering:
+	//   EFS bare:        fs-9919e11b.efs.us-east-1.amazonaws.com
+	//   EFS AZ-prefixed: us-east-1a.fs-9919e11b.efs.us-east-1.amazonaws.com
+	//   S3 Files:        use1-az1.fs-0123456abcdef0189.s3files.us-east-1.on.aws
+	// An optional single leading AZ label may precede the strictly-validated
+	// fs-[0-9a-f]{8,40} id label. The surrounding DNS labels are matched
+	// case-insensitively (DNS names are case-insensitive), but the fs-<hex> id
+	// label itself is kept strictly lowercase hex so that the DNS branch does not
+	// accept ids (e.g. fs-ABCDEF12) that the bare-id branch would reject. The
+	// remaining domain is constrained to the DNS charset ([a-zA-Z0-9.-]) only (no
+	// commas, spaces, slashes, or '='), so a DNS-name volumeHandle cannot be used
+	// to inject additional mount options into the mount source passed to efs-utils.
+	dnsFileSystemIdRegex = regexp.MustCompile(`^([a-zA-Z0-9][a-zA-Z0-9-]*\.)?fs-[0-9a-f]{8,40}\.[a-zA-Z0-9][a-zA-Z0-9.-]*[a-zA-Z0-9]$`)
 )
 
 const (
@@ -563,7 +578,9 @@ func (d *Driver) validateFStype(volCaps []*csi.VolumeCapability) error {
 //   - Examples: `fs-abcd1234::`, `fs-abcd1234:`, `fs-abcd1234`, `fs-abcd1234:/path:fsap-xyz123`
 //
 // COMMON RULES:
-//   - The `{fileSystemID}` is required, and expected to be of the form `fs-...`
+//   - The `{fileSystemID}` is required. It may be either the bare id `fs-[0-9a-f]{8,40}`,
+//     or the mount-target DNS name form (e.g. `fs-abcd1234.efs.<region>.amazonaws.com`,
+//     optionally with a leading AZ label such as `us-east-1a.fs-abcd1234.efs.<region>.amazonaws.com`)
 //   - The `{mountPath}` and `{accessPointID}` are optional -- they may be empty or omitted entirely
 //   - The `{mountPath}`, if specified, is not required to be absolute
 //   - The `{accessPointID}` is expected to be of the form `fsap-...`
@@ -595,7 +612,7 @@ func parseVolumeId(volumeId string) (fsid, subpath, apid string, fsType util.Fil
 		// Validate and extract fsid
 		fsid = tokens[1]
 		if !isValidFileSystemId(fsid) {
-			err = status.Errorf(codes.InvalidArgument, "volume ID '%s' is invalid: Expected a file system ID of the form 'fs-[0-9a-f]{8,40}'", volumeId)
+			err = status.Errorf(codes.InvalidArgument, "volume ID '%s' is invalid: Expected a file system ID of the form 'fs-[0-9a-f]{8,40}' or a mount-target DNS name (e.g. 'fs-abcd1234.efs.<region>.amazonaws.com')", volumeId)
 			return
 		}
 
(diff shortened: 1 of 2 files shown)


vmishra22:
Could you check if regression predates 3.4.1? isValidFileSystemId became strict in commit fd112cef, the single commit of #1796, which replaced `return strings.HasPrefix(filesystemId, "fs-")` with the hexSuffixRegex check that rejects DNS-name handles. CHANGELOG-3.x.md:8 records #1796 under "# V3.0.0", and `git show <tag>:pkg/driver/node.go` shows the strict body unchanged from v3.0.0 through v3.4.2. `git diff v3.4.0..v3.4.1 -- pkg/driver/node.go` is empty, so v3.4.1 did not touch this path.

pkg/driver/node.go
vmishra22, line 616: This message no longer describes what the code accepts, so a malformed DNS handle reads as "the DNS form is unsupported". Please name both forms here and at node.go:574. Three tests assert the string verbatim: node_test.go:547,
node_test.go:646 and node_test.go:717.

pkg/driver/node_test.go
```diff
@@ -1508,6 +1521,62 @@ func TestIsValidFileSystemId(t *testing.T) {
...
+			}
+			if fsType != tc.expectedFsType {
+				t.Errorf("parseVolumeId(%q) fsType = %q, expected %q", tc.volumeId, fsType, tc.expectedFsType)
+			}
+
+			// Mirror the mount source construction in NodePublishVolume (node.go:167).
+			if subpath == "" {
+				subpath = "/"
+			}
+			source := fmt.Sprintf("%s:%s", fsid, subpath)
```
vmishra22: This source assertion cannot fail: it rebuilds node.go:172's own expression, so it
holds even if node.go:172 becomes wrong. TestNodePublishVolume asserts the real source through gomock at node_test.go:823, so add two rows there patterned on node_test.go:124 and node_test.go:681 with a DNS-name fsid

pkg/driver/node.go
vmishra22: Nit: this still says the {fileSystemID} is "expected to be of the form 'fs-...'",
which is wrong for the AZ-prefixed form this PR accepts.
us-east-1a.fs-9919e11b.efs.us-east-1.amazonaws.com does not start with "fs-".
Please list the accepted forms.

pkg/driver/node_test.go
```diff
@@ -1508,6 +1521,62 @@ func TestIsValidFileSystemId(t *testing.T) {
...
+				t.Errorf("parseVolumeId(%q) fsid = %q, expected %q (must be unchanged)", tc.volumeId, fsid, tc.expectedFsid)
+			}
+			if apid != "" {
+				t.Errorf("parseVolumeId(%q) apid = %q, expected empty", tc.volumeId, apid)
+			}
+			if fsType != tc.expectedFsType {
+				t.Errorf("parseVolumeId(%q) fsType = %q, expected %q", tc.volumeId, fsType, tc.expectedFsType)
+			}
+
+			// Mirror the mount source construction in NodePublishVolume (node.go:167).
```
vmishra22: Nit: this references node.go:167, but the mount source is built at node.go:172.
Line numbers go stale on the next edit; naming NodePublishVolume is enough.

pkg/driver/node.go
```diff
@@ -47,6 +47,18 @@ var (
...
+	// enumerating known domains, covering:
+	//   EFS bare:        fs-9919e11b.efs.us-east-1.amazonaws.com
+	//   EFS AZ-prefixed: us-east-1a.fs-9919e11b.efs.us-east-1.amazonaws.com
+	//   S3 Files:        use1-az1.fs-0123456abcdef0189.s3files.us-east-1.on.aws
+	// An optional single leading AZ label may precede the strictly-validated
+	// fs-[0-9a-f]{8,40} id label. The remaining domain is constrained to the DNS
+	// charset ([a-z0-9.-]) only (no commas, spaces, slashes, or '='), so a
+	// DNS-name volumeHandle cannot be used to inject additional mount options
+	// into the mount source passed to efs-utils.
+	dnsFileSystemIdRegex = regexp.MustCompile(`^([a-z0-9][a-z0-9-]*\.)?fs-[0-9a-f]{8,40}\.[a-z0-9][a-z0-9.-]*[a-z0-9]$`)
```
vmishra22: Nit: the regex is lowercase only, so fs-9919e11b.EFS.us-east-1.amazonaws.com is
rejected, though v2.0.6 accepted it and DNS names are case-insensitive. If you
change it, keep the fs-<hex> label lowercase rather than prefixing a blanket (?i),
which accepts fs-ABCDEF12 in the DNS branch while the bare-id branch rejects it.

pkg/driver/node.go
wangnyue: ah, yes. Will fix this

wangnyue: yep, will change it

pkg/driver/node_test.go
wangnyue: good catch. Will fix it

wangnyue:
> Could you check if regression predates 3.4.1? isValidFileSystemId became strict in commit [fd112ce](https://github.com/kubernetes-sigs/aws-efs-csi-driver/commit/fd112cef8abf7770593909df53da5bb0dded1eb2), the single commit of #1796, which replaced `return strings.HasPrefix(filesystemId, "fs-")` with the hexSuffixRegex check that rejects DNS-name handles. CHANGELOG-3.x.md:8 records #1796 under "# V3.0.0", and `git show <tag>:pkg/driver/node.go` shows the strict body unchanged from v3.0.0 through v3.4.2. `git diff v3.4.0..v3.4.1 -- pkg/driver/node.go` is empty, so v3.4.1 did not touch this path.

yes. you are right. regression should be starting from 3.0.0

pkg/driver/node_test.go
wangnyue: fixed it

aliyajo:
Can we squash commits?

vmishra22 (approved):
LGTM

DavidXU12345 (approved):

PR #1948, opened 2026-08-25
feat: write per-volume efs_meta.json during NodePublishVolume

tzneal:
Write a per-mount metadata file under /csi/mounts, keyed by
sha256(cleaned target path), after successful mounts. Each file
records the volume handle, resolved context, mount flags, read-only
state, and actual IAM usage.

**What testing is done?** 

Unit tests + deployed to my cluster and validated metadata file creation/removal.

### DIFF
--- a/pkg/driver/node.go
+++ b/pkg/driver/node.go
@@ -18,11 +18,14 @@ package driver
 
 import (
 	"context"
+	"crypto/sha256"
+	"encoding/hex"
 	"encoding/json"
 	"fmt"
 	"net"
 	"os"
 	"path"
+	"path/filepath"
 	"regexp"
 	"strconv"
 	"strings"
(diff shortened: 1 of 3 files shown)


pkg/driver/node.go
```diff
@@ -265,6 +298,25 @@ func (d *Driver) NodePublishVolume(ctx context.Context, req *csi.NodePublishVolu
...
+		},
+		VolumeContext: efsVolumeContextMeta{
+			EncryptInTransit: encryptInTransit,
+			MountTargetIP:    resolvedMountTargetIP,
+			CrossAccount:     crossAccountDNSEnabled,
+		},
+		MountFlags: mountOptions,
+		ReadOnly:   req.GetReadonly(),
+		Iam: hasOption(mountOptions, "iam") ||
+			fsType == util.FileSystemTypeS3Files,
```
YangjinanHu, line 322: CR description also mentions:
> ture ... when accessPointId is set

But here it does not have accessPointId logic

Note for EFS, accesspoint does not require iam. https://docs.aws.amazon.com/efs/latest/ug/mounting-access-points.html

pkg/driver/node.go
```diff
@@ -318,6 +371,8 @@ func (d *Driver) NodeUnpublishVolume(ctx context.Context, req *csi.NodeUnpublish
 
 	klog.V(5).Infof("NodeUnpublishVolume: %s unmounted", target)
 
+	removeEfsMeta(target)
```
YangjinanHu: Kubelet's `removeMountDir` deletes only `vol_data.json` then rmdir's the PV dir. Will this file causes ENOTEMPTY and fails teardown if for some reason this metadata file isn't cleaned up?

For example, if this driver version is rolled back, the old NodeUnpublish won't remove `efs_meta.json`or if for some reason driver pod is absent at pod deletion (node reboot, DaemonSet update) and removeEfsMeta(target) isn't called but `vol_data.json` is already deleted.

We may need a mechanism to recover from this situation.

tzneal: I can put them in `/var/lib/kubelet/plugins/efs.csi.aws.com/mounts/<hash of req.TargetPath>.json` which won't have this issue. If driver rolls back, files get left but doesn't block any kubelet cleanup as the CSI owns that dir.

pkg/driver/node.go
tzneal: Updated description

YangjinanHu (approved):

PR #1953, opened 2026-09-03
fix: stop NodeUnpublishVolume leaking a thread per retry on a dead mount

vmishra22:

Bug fix.

`NodeUnpublishVolume` could leak one OS thread per kubelet retry, without limit, against a mount whose `efs-proxy` had died. In one reported incident a node accumulated 9,780 permanently blocked threads and stayed unusable for 12 days. The Go runtime aborts the process at 10,000 threads (`runtime/proc.go`, `sched.maxmcount`).

Two causes, both in the non-force path:

1. The pre-check called `NodeMounter.GetDeviceName`, which delegated to `mount_utils.GetDeviceNameFromMount`. That resolves the target with `filepath.EvalSymlinks`, so it stats the mountpoint itself. On an unresponsive `hard` NFS mount the stat blocks in the kernel uninterruptibly, and the handler does not consult the request context, so no handler reached the unmount.
2. Nothing bounded concurrency. kubelet abandons the call at its RPC deadline and retries, but the abandoned handler keeps running, so every retry added another permanently blocked thread. The existing in-flight tracker covers only `NodePublishVolume`.

This PR reimplements `GetDeviceName` to read `/proc/mounts` and resolve symlinks on the target's parent only, never the target, and adds a per-target in-flight guard that rejects a duplicate unpublish with `ABORTED`. kubelet retries that like any other failure.

Scope, stated plainly. This contains the failure rather than repairing it. The unmount itself is still unbounded, so one handler per wedged target can remain stuck and that volume stays unpublishable until the mount responds or the driver restarts. The guard is what holds the cost at one handler instead of one per retry. Escalating to `umount -f` stays behind the existing `--force-unmount-after-timeout` opt-in, which defaults to off, so operators choose force rather than receiving it by default.

Two paths with the same unbounded shape are not addressed here: `NodePublishVolume` stats the target through `MakeDir`, and `NodeGetVolumeStats` calls `os.Stat` on it directly.

**What testing is done?**

`make test` and `make verify` pass.

- `TestGetDeviceName` covers the mounted, unmounted, non-canonical and symlinked-parent cases. One subtest pins the property that matters: the target is itself a symlink whose destination is also in the mount listing under a different device, so the test fails if anything resolves the target again.
- `TestNodeUnpublishVolumeDeduplicatesConcurrentCalls` runs against both the default and the `forceUnmountAfterTimeout` branch. It asserts the duplicate is rejected with `ABORTED` without reaching the mounter, that a trailing-slash spelling is also rejected, and that the guard is released once the first call returns. Both channel waits have deadlines so a regression fails rather than hanging the package.
- Checked by mutation. Re-introducing a stat of the mountpoint, deleting the not-mounted early return, guarding only one branch, or reverting the unmount call each turn a test red.

### DIFF
--- a/pkg/driver/driver.go
+++ b/pkg/driver/driver.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"net"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/container-storage-interface/spec/lib/go/csi"
@@ -57,6 +58,9 @@ type Driver struct {
 	volumeAttachLimit        int64
 	forceUnmountAfterTimeout bool
 	unmountTimeout           time.Duration
+	// inFlightUnpublishTargets holds target paths with a NodeUnpublishVolume in
+	// flight, so a kubelet retry cannot stack a handler behind a blocked one.
+	inFlightUnpublishTargets sync.Map
 	// metaDir is bind-mounted to /var/lib/kubelet/plugins/efs.csi.aws.com/mounts on the host.
 	metaDir string
 }
--- a/pkg/driver/mounter.go
+++ b/pkg/driver/mounter.go
@@ -15,6 +15,7 @@ package driver
 
 import (
 	"os"
+	"path/filepath"
 
 	mount_utils "k8s.io/mount-utils"
 )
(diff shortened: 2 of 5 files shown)


DavidXU12345 (dismissed):

DavidXU12345 (commented):
Revoke approval

PR #1955, opened 2026-09-09
fix: Deny transport-security keys in conf overrides

wangnyue:
## What is this PR about? / Why do we need it?

The `--efs-utils-conf-overrides` and `--s3files-utils-conf-overrides` flags let an operator override arbitrary `section:key=value` pairs in `efs-utils.conf` / `s3files-utils.conf`. A small set of these keys control the mount transport itself — the DNS suffix used to resolve the endpoint and whether stunnel verifies the server certificate. Allowing those to be set through the conf-override interface means the transport target and TLS verification can be changed via configuration.

This PR hardens the flags by rejecting three transport-security keys at driver startup validation, regardless of section:

- `dns_name_suffix`
- `stunnel_check_cert_hostname`
- `stunnel_check_cert_validity`

All other override keys (e.g. `fips_mode_enabled`, `retry_nfs_mount_command_timeout_sec`, `stunnel_health_check_command_timeout_sec`, `unmount_count_for_consistency`, `readahead_cache_enabled`) continue to work unchanged, so existing legitimate use of the flags is unaffected.

## How was this change tested?

- `go build ./...`, `go vet ./pkg/driver/`, and `gofmt` are clean.
- Added `TestValidateConfOverridesDenylist` covering: each safe key and combinations (allowed), the empty value (allowed), each denied key individually, a denied key in a non-default section, a denied key mixed with safe keys, and a multi-pair string of all denied keys (all rejected).
- `go test ./pkg/driver/` passes, including the existing `TestValidate` override cases.

## Does this PR introduce a user-facing change?
```release-note
The efs-utils-conf-overrides and s3files-utils-conf-overrides flags now reject the transport-security keys dns_name_suffix, stunnel_check_cert_hostname, and stunnel_check_cert_validity.
```


pkg/driver/conf_overrides.go
```diff
@@ -12,6 +12,36 @@ type ConfOverride struct {
...
+// settable via --efs-utils-conf-overrides or --s3files-utils-conf-overrides.
+// These keys control the mount transport DNS target and TLS certificate
+// verification; allowing them to be set through the conf-override interface lets
+// a caller redirect the transport or disable certificate checks. This is a
+// defense-in-depth layer that backs the EKS add-on schema denylist: it applies
+// on every install path (EKS add-on, self-managed Helm, and manual), not just
+// the add-on API. Keys are matched by name regardless of section, mirroring the
+// schema denylist semantics.
+var confOverridesDenylist = map[string]bool{
+	"dns_name_suffix":             true,
```
DavidXU12345, line 24: - Shall we separate s3files and efs conf denylist 
- Shall we add `dns_name_format` and `stunnel_cafile` into denylist?

pkg/driver/options.go
```diff
@@ -90,12 +90,18 @@ func (o *Options) Validate() error {
 	if err != nil {
 		return fmt.Errorf("invalid efs-utils-conf-overrides: %w", err)
 	}
+	if err := validateConfOverridesDenylist(parsed, "efs-utils-conf-overrides"); err != nil {
```
DavidXU12345: We should add unit test in `options_test.go` as well

pkg/driver/conf_overrides.go
```diff
@@ -12,6 +12,36 @@ type ConfOverride struct {
 	Value   string
 }
 
+// confOverridesDenylist is the set of security-sensitive keys that must never be
```
DavidXU12345: We should update `docs/parameters.md` to reflect the denylist

DavidXU12345 (changes requested):

pkg/driver/conf_overrides.go
wangnyue: Done. Split into `efsUtilsConfOverridesDenylist` and `s3filesUtilsConfOverridesDenylist` (each flag validates against its own map, so the two conf files can evolve independently), and added `dns_name_format` and `stunnel_cafile` to both. `validateConfOverridesDenylist` now takes the denylist as a parameter.

pkg/driver/options.go
wangnyue: Done. Added `TestValidate` cases in `options_test.go` that exercise the denylist through the full `Validate()` path — denied key in efs-utils overrides, denied key in s3files overrides, denied `stunnel_cafile`, and an uppercase variant (case-insensitive rejection).

pkg/driver/conf_overrides.go
wangnyue: Done. Updated `docs/parameters.md`: both the `efs-utils-conf-overrides` and `s3files-utils-conf-overrides` rows now note that the transport-security keys (`dns_name_suffix`, `dns_name_format`, `stunnel_cafile`, `stunnel_check_cert_hostname`, `stunnel_check_cert_validity`) are denied (case-insensitive) and cause a startup error.

wangnyue:
@DavidXU12345 thanks for the review — all four points addressed in the latest push:

1. **Separate denylists** — split into `efsUtilsConfOverridesDenylist` and `s3filesUtilsConfOverridesDenylist`; each flag validates against its own map (`validateConfOverridesDenylist` now takes the denylist as a parameter), so the two conf files can evolve independently.
2. **Added keys** — `dns_name_format` and `stunnel_cafile` are now denied in both lists (alongside `dns_name_suffix`, `stunnel_check_cert_hostname`, `stunnel_check_cert_validity`).
3. **options_test.go** — added `TestValidate` cases exercising the denylist through the full `Validate()` path (denied key in each flag, `stunnel_cafile`, and an uppercase/case-insensitive variant).
4. **docs/parameters.md** — both conf-overrides rows now document the denied transport-security keys and that they cause a startup error.

`go build`, `go vet`, `gofmt`, and `go test ./pkg/driver/...` all pass.

DavidXU12345 (approved):

PR #1964, opened 2026-09-16
Update BasePath Parsing for reuseAccessPoint

samuhale:
Bug fix

Prevent false positives when matching basePath to the AccessPointRootDir

**What testing is done?** 
Unit testing


pkg/driver/controller.go
```diff
@@ -897,10 +897,16 @@ func (d *Driver) buildAccessPointTags() map[string]string {
...
 func validateExistingAccessPoint(existingAccessPoint *cloud.AccessPoint, basePath string, gid int64, gidSpecified bool, uid int64, uidSpecified bool, gidMin int64, gidMax int64) error {
-	normalizedBasePath := strings.TrimPrefix(basePath, "/")
-	normalizedAccessPointPath := strings.TrimPrefix(existingAccessPoint.AccessPointRootDir, "/")
-	if !strings.HasPrefix(normalizedAccessPointPath, normalizedBasePath) {
-		return fmt.Errorf("Access point found but has different base path than what's specified in storage class")
+	normalizedBasePath := strings.Trim(basePath, "/")
+	normalizedAccessPointPath := strings.Trim(existingAccessPoint.AccessPointRootDir, "/")
+	if normalizedBasePath == "" {
+		// basePath was omitted: the access point root directory must be created directly under the file system root
+		if strings.Contains(normalizedAccessPointPath, "/") {
```
DavidXU12345: why are we checking whether it contains slash or not? do you mean `tenant-b/shared-claim` is not allowed?

DavidXU12345 (changes requested):

pkg/driver/controller.go
```diff
@@ -897,10 +897,12 @@ func (d *Driver) buildAccessPointTags() map[string]string {
...
 
 func validateExistingAccessPoint(existingAccessPoint *cloud.AccessPoint, basePath string, gid int64, gidSpecified bool, uid int64, uidSpecified bool, gidMin int64, gidMax int64) error {
-	normalizedBasePath := strings.TrimPrefix(basePath, "/")
-	normalizedAccessPointPath := strings.TrimPrefix(existingAccessPoint.AccessPointRootDir, "/")
-	if !strings.HasPrefix(normalizedAccessPointPath, normalizedBasePath) {
-		return fmt.Errorf("Access point found but has different base path than what's specified in storage class")
+	normalizedBasePath := strings.Trim(basePath, "/")
+	normalizedAccessPointPath := strings.Trim(existingAccessPoint.AccessPointRootDir, "/")
+	if normalizedBasePath != "" &&
+		normalizedAccessPointPath != normalizedBasePath &&
```
DavidXU12345: Do we need this check? if `normalizedAccessPointPath == normalizedBasePath`, I think it will also pass the check `strings.HasPrefix(normalizedAccessPointPath, normalizedBasePath+"/")`, so the error branch won't be hit

DavidXU12345 (approved):

